PR #16, opened 2019-04-03
feat: Improve logging

jan-auer:
Adds configuration for logging and allows to specify different log formats. The default is to log `INFO` in pretty (colored) output. Certain dependencies are lowered in their log level, however, to reduce noise when debugging. Finally, top-level errors are now pretty printed to the console with all their causes (and backtrace, if enabled).

Besides logging to stdout, the `sentry` logging integration is now enabled. This will send all `error!` messages to Sentry.

cc @tonyo: We should set these, just like for semaphore.


jan-auer:
The readme is out of date, I will update this separately after merging.

PR #23, opened 2019-04-10
feat: Process minidumps

untitaker:
Ability to stackwalk and symbolicate minidumps. Is not yet used in Sentry.


src/actors/cficaches.rs
```diff
@@ -0,0 +1,209 @@
...
+
+        // TODO: Backoff + retry when download is interrupted? Or should we just have retry logic
+        // in Sentry itself?
+        let result = objects
+            .send(FetchObject {
+                filetypes: FileType::from_object_type(&self.request.object_type),
+                identifier: self.request.identifier.clone(),
+                sources: self.request.sources.clone(),
+                scope: self.request.scope.clone(),
+                // TODO: Prefer files with unwind info
```
untitaker: @jan-auer right now fetching objects for cfi caches goes through the same heuristic that is used for fetching files for symcaches, i.e. it prefers objects where `has_debug_info() == true`. I assume this needs to be changed to prefer objects where `has_unwind_info() == true` instead?

src/actors/symbolication.rs
```diff
@@ -152,6 +172,203 @@ impl SymbolicationActor {
...
+            scope,
+            sources,
+        } = request;
+
+        let byteview = tryf!(ByteView::read(file).map_err(|_| SymbolicationError::Minidump));
+
+        let object_infos = self.threadpool.spawn_handle(future::lazy(move || {
+            log::info!("Minidump size: {}", byteview.len());
+            metric!(time_raw("minidump.upload.size") = byteview.len() as u64);
+            // XXX: Use ByteView::from_file
```
untitaker: TODO

src/actors/symbolication.rs
```diff
@@ -152,6 +172,203 @@ impl SymbolicationActor {
...
+            cfi_requests.and_then(clone!(threadpool, scope, sources, |(
+                byteview,
+                cfi_requests,
+            )| {
+                threadpool.spawn_handle(future::lazy(move || {
+                    let mut frame_info_map = FrameInfoMap::new();
+                    let mut modules = vec![];
+
+                    for (object_info, code_module_id, result) in &cfi_requests {
+                        // XXX: We should actually build a list of FetchedDebugFile instead of
```
untitaker: TODO

src/actors/cficaches.rs
jan-auer: Correct. See the „Symbol Lookup Strategy“ document for a detailed table.

src/actors/symbolication.rs
untitaker: I concluded we need a separate field in `ObjectInfo` for representing the status of fetching the CFI file. This might yield entirely different errors than fetching the debug file during symbolication.

untitaker: deferring

src/actors/symbolication.rs
```diff
@@ -192,19 +192,21 @@ impl SymbolicationActor {
...
-                            code_module.id()?,
-                            object_info_from_minidump_module(&os_name, code_module),
-                        ))
-                    })
-                    .collect::<Vec<_>>(),
-            ))
+            let referenced_modules = state.referenced_modules();
+
+            let object_infos: Vec<_> = state
+                .referenced_modules()
```
jan-auer: Should this call `state.modules()` instead, so that it contains all modules? 

Also, I would suggest rebuilding this list from the second process state. Due to the use of CFI, the list of referenced modules might change slightly (a module might no longer be referenced since it was a false-positive in the stack scan).

untitaker: Yes, it should be `modules()`. This list will be discarded before calling out to symbolication so the "unused" status should be correct.

I initially thought that we will later change this to use these object infos as a base for the debug images response but I am reconsidering this now.

untitaker: this should be fixed now

PR #27, opened 2019-04-17
ref: Propagate hub data across entire symbolication process

untitaker:


untitaker:
@jan-auer mentioned that he wants new breadcrumbs per actix message, that would mean we could omit the ugliest part from this PR

PR #28, opened 2019-04-17
fix: Call capture_fail in every step of the pipeline

untitaker:
We want to capture the real `Fail` without any wrapper `Fail`s attached, because:

* due to how error chains are currently grouped in Sentry (least specific error is the one shown in title)
* in async envs, breadcrumbs are probably better when capturing close to the error

### DIFF
--- a/src/actors/objects/mod.rs
+++ b/src/actors/objects/mod.rs
@@ -8,6 +8,7 @@ use actix::{Actor, Addr, Context, Handler, Message, ResponseFuture};
 use bytes::Bytes;
 use failure::{Fail, ResultExt};
 
+use ::sentry::integrations::failure::capture_fail;
 use futures::{future, Future, IntoFuture, Stream};
 use symbolic::common::ByteView;
 use symbolic::debuginfo::Object;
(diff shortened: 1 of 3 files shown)


untitaker:
Sure, we discussed that. I didn't really want to commit to that approach but given how scopes work in the SDK I think we have not much of a choice.

untitaker:
I'll look into this tomorrow

untitaker:
I'd still like to merge this independently because the "proper" refactor will touch everything everywhere.

src/actors/objects/mod.rs
```diff
@@ -8,6 +8,7 @@ use actix::{Actor, Addr, Context, Handler, Message, ResponseFuture};
 use bytes::Bytes;
 use failure::{Fail, ResultExt};
 
+use ::sentry::integrations::failure::capture_fail;
```
jan-auer: _nit_: Remove the leading `::`

untitaker: as discussed irl, not possible due to `objects::sentry` and us being in `mod.rs` rn

PR #29, opened 2019-04-17
feat: Filter sources by path prefix

untitaker:
We have scaling and ratelimiting concerns when hitting the MS symbol
server with every symbol we try to download.

Also refactor the schema to have a `filters` section in source configs,
and deduplicate some code between S3 and HTTP.


untitaker:
**do not deploy until filters are sent the same way by sentry**

src/actors/objects/paths.rs
```diff
@@ -2,7 +2,56 @@ use std::fmt::Write;
...
+            Some(DownloadPath(get_directory_path(
+                layout, filetype, &object_id,
+            )?))
+        } else {
+            None
+        }
+    })
+}
+
+fn matches_path_prefixes(object_id: &ObjectId, prefixes: &[String]) -> bool {
```
jan-auer: If we choose to keep this, let's add a utility function to `symbolic_common` that does this. It fits quite well alongside the other stuff in `symbolic/common/src/path.rs`.

src/types.rs
```diff
@@ -136,9 +136,9 @@ pub struct S3SourceConfig {
 /// Common parameters for external filesystem-like buckets configured by users.
 #[derive(Deserialize, Clone, Debug)]
 pub struct ExternalSourceConfigBase {
-    /// File types that are supported by this server.
-    #[serde(default = "FileType::all_vec")]
-    pub filetypes: Vec<FileType>,
+    /// Influence whether this source will be selected
+    #[serde(default)]
+    pub filters: SourceFilters,
```
jan-auer: 👍

src/types.rs
```diff
@@ -148,6 +148,24 @@ pub struct ExternalSourceConfigBase {
...
+
+    /// When nonempty, a list of prefixes to fuzzy-match filepaths against. The source is then only
+    /// used if one of the path prefixes matches.
+    ///
+    /// "Fuzzy" in this context means that (ascii) casing is ignored, `\` is treated as equal
+    /// to `/` and consecutive slashes are treated as single slash.
+    ///
+    /// If a debug image does not contain any path information it will be treated like an image
+    /// whose path doesn't match any prefix.
+    pub path_prefixes: Vec<String>,
```
jan-auer: Let's make these globs instead of just prefixes. Apart from greater flexibility (and efficient library support!) this solves one important issue: With exact prefixes, we cannot specify an arbitrary drive letter for windows paths (e.g. the system could be in `D:\Windows`.

src/actors/objects/paths.rs
untitaker: To me this is just an aspect of the symbolicator protocol. The heavy lifting is done by `glob` anyway.

src/types.rs
untitaker: Good point, done.

Ironically `glob` conflates `//` with `/` because it assumes a Unix FS. So specifying `C://Windows` as a glob does not seem to match anything.

PR #30, opened 2019-04-17
feat: Add unwind_status to debug images

untitaker:


src/types.rs
```diff
@@ -277,6 +277,7 @@ pub struct ObjectInfo {
 
     /// Identifier of the code file.
     #[serde(skip_serializing_if = "is_default")]
+    #[serde(default)]
```
jan-auer: _nit_: For `Option<_>` this is not required, but it's good to have this here explicitly 👍

src/actors/symbolication.rs
```diff
@@ -894,3 +906,30 @@ fn symbolic_registers_to_protocol_registers(
...
+            ObjectFileStatus::Other
+        }
+    }
+}
+
+fn symcache_error_to_status(e: &SymCacheError) -> ObjectFileStatus {
+    match e.kind() {
+        SymCacheErrorKind::Fetching => ObjectFileStatus::FetchingFailed,
+        // nb: Timeouts during download are caught by Fetching
+        SymCacheErrorKind::Timeout => ObjectFileStatus::TooLarge,
```
jan-auer: Let's add an explicit `ObjectFileStatus::Timeout` status or similar to indicate that we timed out during processing. I think this might be quite valuable for us to distinguish from files that are simply too large. 

As an example: We might be downloading _very_ large debug files but only a subset of their information is usable line info. On the other hand, a file that's perfectly in our limits might take too long to process because of a regression in `symbolic` or some other bug.

Finally, I think it would make sense to capture these timeouts to Sentry, with extra info on which file it was. This will be extremely useful for debugging.

(same for CFI above)

src/actors/symbolication.rs
```diff
@@ -239,9 +242,10 @@ impl SymbolicationActor {
             )| {
                 threadpool.spawn_handle(future::lazy(move || {
                     let mut frame_info_map = FrameInfoMap::new();
+                    let mut unwind_statuses = BTreeMap::new();
 
                     for (code_module_id, result) in &cfi_requests {
-                        // XXX: We should actually build a list of FetchedDebugFile instead of
+                        // XXX: We should actually build a list of FetchedObjectFile instead of
```
jan-auer: This is now obsolete.

src/actors/symbolication.rs
```diff
@@ -371,6 +397,14 @@ impl SymbolicationActor {
                             stacktrace.is_requesting = requesting_thread_index.map(|r| r == i);
                         }
 
+                        for (mut module, unwind_status) in response
+                            .modules
+                            .iter_mut()
+                            .zip(unwind_statuses_by_module_index)
+                        {
+                            module.unwind_status = Some(unwind_status);
```
jan-auer: Suggestion: How about we change `SymbolicateStacktraces` to contain the full object info already, and then `do_process_minidump` already writes unwind status in there. It seems a bit clumsy that this is only done after the fact.

src/types.rs
```diff
@@ -419,9 +420,12 @@ pub enum DebugFileStatus {
 
 /// Enhanced information on an in
 #[derive(Debug, Clone, Serialize)]
-pub struct FetchedDebugFile {
-    /// Status for handling this debug file.
-    pub status: DebugFileStatus,
+pub struct FetchedObjectFile {
```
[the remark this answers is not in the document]
untitaker: Done, went with first suggestion

src/types.rs
```diff
@@ -398,18 +399,18 @@ pub struct SymbolicatedStacktrace {
...
     Found,
-    /// The debug image was not referenced in the stack trace and not further handled.
+    /// The image was not referenced in the stack trace and not further handled.
     Unused,
     /// The file could not be found in any of the specified sources.
-    MissingDebugFile,
-    /// The debug file failed to process.
-    MalformedDebugFile,
-    /// The debug file could not be downloaded.
+    MissingFile,
```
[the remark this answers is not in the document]
untitaker: Done

src/actors/symbolication.rs
```diff
@@ -894,3 +906,30 @@ fn symbolic_registers_to_protocol_registers(
...
+        CfiCacheErrorKind::ObjectParsing => ObjectFileStatus::MalformedFile,
+
+        _ => {
+            capture_fail(e);
+            ObjectFileStatus::Other
+        }
+    }
+}
+
+fn symcache_error_to_status(e: &SymCacheError) -> ObjectFileStatus {
```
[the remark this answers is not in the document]
untitaker: Done

src/actors/symbolication.rs
untitaker: Done

src/actors/symbolication.rs
untitaker: Done

src/actors/symbolication.rs
untitaker: Effectively this means renaming `TooLarge` to `Timeout`, as we don't have size limits on symcaches. Is that correct?

src/actors/symbolication.rs
```diff
@@ -498,33 +532,10 @@ impl SymCacheLookup {
...
-                                                capture_fail(&*e);
-                                                DebugFileStatus::Other
-                                            }
-                                        };
-
-                                        Ok((None, status))
+                                        Some(_) => Ok((Some(symcache), ObjectFileStatus::Found)),
+                                        None => Ok((Some(symcache), ObjectFileStatus::MissingFile)),
                                     })
+                                    .or_else(|e| Ok((None, symcache_error_to_status(&e))))
```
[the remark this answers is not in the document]
untitaker: This would constrain us to only sending `Found` when a symcache is present. My assumption is that we will extend this set of states to also include `Breakpad` and other variants where we have a symcache but still want to return a variety of statuses

src/actors/symbolication.rs
jan-auer: Correct 👍

untitaker: done

PR #42, opened 2019-05-07
fix: Normalize debug ids for minidumps

untitaker:
This leaves the behavior for incoming requests untouched, where the debug ids will be echoed back unmodified.


untitaker:
@jan-auer pls check again

src/types.rs
```diff
@@ -499,7 +501,19 @@ pub struct CompleteObjectInfo {
...
-    fn from(raw: RawObjectInfo) -> Self {
+    fn from(mut raw: RawObjectInfo) -> Self {
+        if let Some(ref raw_id) = raw.debug_id {
+            if let Ok(id) = DebugId::from_breakpad(raw_id.as_str()) {
+                raw.debug_id = Some(id.to_string());
+            }
+        }
+
+        if let Some(ref raw_id) = raw.code_id {
+            if let Ok(id) = raw_id.parse() {
```
jan-auer: This parses `raw_id` into a String, which is a no-op. Run this instead:

```suggestion
            if let Ok(id) = raw_id.parse::<CodeId>() {
```

untitaker: Good catch...

untitaker:
Before we merge this we need to update snapshots in Sentry again or find an alternative solution there.

PR #45, opened 2019-05-08
feat: Add trust for minidump frames

untitaker:


untitaker:
@jan-auer I hope the stuff I pushed to symbolic is fine.. it seems uncontroversial to me :)

jan-auer:
All good. In Sentry we should set `FrameTrust::Prewalked` when we retain ingested stack traces, such as the portable call stack.

untitaker:
@jan-auer I was thinking of doing that in store normalization

PR #48, opened 2019-05-09
fix: Send more data to Sentry to debug symbolic

untitaker:


src/actors/common/cache.rs
```diff
@@ -125,6 +126,19 @@ impl<T: CacheItemRequest> Cacher<T> {
...
                 );
 
+                configure_scope(|scope| {
+                    scope.set_extra(
+                        &format!("cache.{}.cache_path", name),
+                        format!("{:?}", path).into(),
+                    );
+                });
+
+                log::debug!(
```
jan-auer: _nit_: Since this happens *very* frequently and we have a tag for this, can we make this `log::trace!`?

src/sentry.rs
```diff
@@ -37,3 +37,7 @@ pub trait SentryFutureExt: Sized {
 }
 
 impl<F> SentryFutureExt for F {}
+
+pub trait WriteSentryScope {
```
jan-auer: _nit_: Docs for this, please. I guess the trait is nice, but just a plain function would've sufficed :D

PR #50, opened 2019-05-10
fix: Bump download timeouts

untitaker:
I don't quite understand the problem but it seems Sentry is just slow to respond here. Maybe we want this for external buckets someday too, but I think we generally give the customer less connectivity guarantees to external services. For now just make sure we can roll Symbolicator out without regressing.

### DIFF
--- a/src/actors/objects/http.rs
+++ b/src/actors/objects/http.rs
@@ -1,4 +1,5 @@
 use std::sync::Arc;
+use std::time::Duration;
 
 use actix_web::http::header::HeaderName;
 use actix_web::{client, HttpMessage};
@@ -52,7 +53,7 @@ pub fn prepare_downloads(
 }
 
 pub fn download_from_source(
-    source: &HttpSourceConfig,
+    source: Arc<HttpSourceConfig>,
     download_path: &DownloadPath,
 ) -> Box<Future<Item = Option<DownloadStream>, Error = ObjectError>> {
     // XXX: Probably should send an error if the URL turns out to be invalid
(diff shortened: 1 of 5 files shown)


untitaker:
BTW, this timeout is still not the overall timeout for downloading the entire file. It seems Sentry doesn't even send headers in 5 secs.

jan-auer:
I’ll have to double check but it doesn’t make sense to me right now. Judging by the logs it seemed that headers had arrived at that point (at least the initial line containing the response code).

jan-auer:
Waiting a minute for nothing seems pretty hard, anyway. We need to try to provoke the same locally. I did some tests against production and the response always was instant. That might have been due to caches, though

untitaker:
@jan-auer this does seem to happen reasonably rarely but i think we need to just make sure to get the symbol no matter what. If Sentry experiences abnormal slowness we also can't just chug on and create new groups because we didn't manage to symbolicate.

untitaker:
I realize that this will probably happen with external buckets and also happens all the time with JS, but I think native users have different expectations towards the internal bucket.

jan-auer:
So if this is really about headers not arriving in time, then this is definitely due to delays in fechting chunks from the blob store. That we have to anticipate, but then even 60s will not be enough. 
Unfortunately, it was also not that infrequent. Don’t remember the exact numbers but it looks like it happened for about 10% of all symbolication requests.

untitaker:
>then even 60s will not be enough

I want to figure this out on the go. there are a lot of seconds between 5 and 60 already

untitaker:
When writing this I asked for clarification on Gitter. Nikolay said it's
time until a response obj is returned, which happens before streaming
starts.

jan-auer:
Aha, then we have a bug in symbolicator. Presumably, the timeout starts too early as the future is spawned before it executes.

untitaker:
Let's discuss this irl. You mentioned this a bunch of times but I don't really understand why you think that. I think it's entirely possible for single connections to become slow when we send 100s of requests in parallel.

untitaker:
After a while we figured that the timeout applies to the entire stream, probably a bug in actix.

Follow-up at #54

PR #52, opened 2019-05-13
ref: Skip empty identifiers gracefully

jan-auer:
This really shouldn't happen, but there's also no reason not to handle empty strings gracefully.


jan-auer:
Requires https://github.com/getsentry/symbolic/pull/142

PR #55, opened 2019-05-14
build(docker): Build with debug information

jan-auer:
Enables release builds with debug information. Locally, release builds still do not contain debugging information for faster builds and smaller target directory sizes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN apt-get update \
 ADD Cargo.toml Cargo.lock ./
 RUN mkdir -p src \
     && echo "fn main() {}" > src/main.rs \
-    && cargo build --release --locked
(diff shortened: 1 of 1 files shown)


jan-auer:
@tonyo let's check if this builds correctly on Google Cloud. 
I ran a local build and it seems alright (other than taking ages).

Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN mkdir -p src \
     && cargo build --release --locked
```
tonyo: Do we want to do the same thing here, for dependencies?

tonyo: Ah, probably I don't understand how this works, ignore this.

jan-auer: :/ oh yes, completely missed that.

tonyo:
The non-stripped binary is around 260MB now.

PR #56, opened 2019-05-14
ref: Show useful errors when minidump processing fails

untitaker:


jan-auer:
@tonyo @untitaker Note that this **must not** be deployed before Sentry is updated with the new status code. 

@untitaker When implementing the corresponding change in Sentry, please also make sure we're catching down unexpected status codes and coerce them into an event error.

PR #74, opened 2019-05-31
ref(cache): Regenerate outdated caches automatically

jan-auer:
This discards symcaches and cficaches with an outdated version (actually, any version that is not current) and computes them again.

### DIFF
--- a/src/actors/common/cache.rs
+++ b/src/actors/common/cache.rs
@@ -141,7 +141,8 @@ impl<T: CacheItemRequest> Cacher<T> {
                         None => continue,
                     };
 
-                    if !request.should_load(&byteview) {
+                    let status = CacheStatus::from_content(&byteview);
+                    if status == CacheStatus::Positive && !request.should_load(&byteview) {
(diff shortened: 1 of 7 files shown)


src/actors/common/cache.rs
```diff
@@ -126,6 +141,13 @@ impl<T: CacheItemRequest> Cacher<T> {
                         None => continue,
                     };
 
+                    let status = CacheStatus::from_content(&byteview);
+                    if status == CacheStatus::Positive && !request.should_load(&byteview) {
+                        log::trace!("Discarding {} at path {:?}", name, path);
+                        metric!(counter(&format!("caches.{}.file.discarded", name)) += 1);
+                        continue;
```
untitaker: Note that at this point the mtime of the file is already bumped. This is fine since the file is about to get overwritten anyway for local scopes, and for global scopes it probably just expires normally.

Maybe we will figure out one day how to clean up discarded caches from within the cleanup command.

PR #83, opened 2019-06-11
fix: Prefer symbol tables over nothing

untitaker:
Extracted from #80, preferrably deploy this with #82 such that we get new, better symcaches.


untitaker:
We can merge this before the other one, and even deploy it independently. We just want to invalidate the symcaches at some point.

PR #91, opened 2019-06-27
Fix breakpad symbol path

dschmidt:
A proper path looks like this: Qt5Core.pdb/6CA082053866490D87E82C42650B67F71/Qt5Core.sym
Before this change a path like this was generated:  Qt5Core.sym/6CA082053866490D87E82C42650B67F71/Qt5Core.pdb


dschmidt:
Well, I had it working before with `minidump_stackwalk` - so it was spurious symbolicator didn't request a working path from my http source :)

PR #93, opened 2019-07-02
feat: Block reserved IPs by default

### LINKED ISSUE #4: Block download requests to internal IPs
Outgoing requests for HTTP sources need to verify that no connection is made to an internal host based on some configurable blacklist.

This does not apply for S3 sources, since those only allow connection directly with Amazon; assuming that the local network is not compromised.

The Sentry source is somewhat special cased, as this is an internal use case. Most likely, there should be configuration whether Sentry sources are even allowed defaulting to `false`.

untitaker:
Fix #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,10 +98,10 @@ See [the documentation on caching](docs/caching.md).
 
 <a name=ref-security />
 
-In the context of Sentry, source configuration is considered untrusted user
-input. This is why, by default, (`connect_to_reserved_ips=false`) source configs
-are not allowed to connect to [reserved IP
-ranges](https://en.wikipedia.org/wiki/Reserved_IP_addresses).
+In the context of Sentry, a subset of source configuration is user input (has
+to be valid JSON, only certain source types allowed). This is why, by default,
+(`connect_to_reserved_ips=false`) source configs are not allowed to connect to
+[reserved IP ranges](https://en.wikipedia.org/wiki/Reserved_IP_addresses).
 
 An exception from this rule is the Sentry source type, which is always allowed
 to connect to any kind of IP address because the user can't enter a custom
--- a/README.md
+++ b/README.md
@@ -73,6 +73,14 @@ sentry_dsn: https://mykey@sentry.io/4711
   default but requires the `sources` key to be configured.  The symstore proxy
   lets you download raw symbols from the symbolicator as if it was a symstore
   (Microsoft Symbol Server) compatible server.
+- `connect_to_unsafe_ips`: Allow reserved IP addresses for requests to sources.
+    
+   By default (value of `false`) source configs are not allowed to connect to
+   internal IP networks or loopback, because most types of sources are
+   considered untrusted user input.
+    
+   An exception from this rule is the Sentry source type, which is always
+   allowed to connect to any kind of IP address.
 
 ### Starting
 
--- a/README.md
+++ b/README.md
@@ -73,14 +73,9 @@ sentry_dsn: https://mykey@sentry.io/4711
   default but requires the `sources` key to be configured.  The symstore proxy
   lets you download raw symbols from the symbolicator as if it was a symstore
   (Microsoft Symbol Server) compatible server.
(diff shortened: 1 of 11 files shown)


src/actors/objects/sentry.rs
```diff
@@ -170,7 +176,7 @@ pub(super) fn download_from_source(
     });
 
     let response = Retry::spawn(
-        ExponentialBackoff::from_millis(100).map(jitter).take(3),
+        ExponentialBackoff::from_millis(10).map(jitter).take(3),
```
untitaker: The retry intervals were way too large. This would retry a single http request for over a minute.

Since we do a catch-all retry around the HTTP request we will also retry blacklisted IPs. It is unwieldy to not retry on a particular error as it's wrapped in a bunch of untyped actix errors.

README.md
```diff
@@ -89,18 +92,20 @@ settings in this case.
...
+## Security and trust of source configs
 
-The cache is currently unbound, there is no expiration logic. It is recommended
-to prune files yourself, starting with files having the oldest `mtime`.
-Effectively implementing a LRU cache from the outside of the process. You
-should be able to `unlink` files while the server is running without it causing
-issues (assuming a POSIX filesystem).
+<a name=ref-security />
+
+In the context of Sentry, source configuration is considered untrusted user
```
jan-auer: This is completely independent of Sentry.

src/config.rs
```diff
@@ -150,6 +151,9 @@ pub struct Config {
 
     /// Default list of sources and the sources used for proxy mode.
     pub sources: Arc<Vec<SourceConfig>>,
+
+    /// Allow reserved IP addresses for requests to sources.
+    pub connect_to_reserved_ips: bool,
```
jan-auer: Instead of making this just a bool, let's assume that everyone will want to avoid requests to localhost by default. Also, in Sentry we will need to override this with a bunch of network-internal IP ranges. That's why I would suggest to make this a list, and assign `RESERVED_IP_BLOCKS` (from `http.rs`) as `#[serde(default = "...")]`. Then, we can _always_ use the `SafeResolver` by default.

src/http.rs
```diff
@@ -69,3 +90,107 @@ pub fn follow_redirects(
...
+            .map_err(|mailbox_e| ResolverError::Resolver(mailbox_e.to_string()))
+            .flatten()
+            .and_then(move |mut addrs| {
+                addrs.retain(|addr| {
+                    for network in &*RESERVED_IP_BLOCKS {
+                        if network.contains(addr.ip()) {
+                            metric!(counter("http.blacklisted_ip") += 1);
+                            log::error!(
+                                "Blocked attempt to connect to reserved IP address: {}",
+                                addr.ip()
```
jan-auer: Let's not log a error here, as this is very much expected to happen. The metric is A+ though.

src/http.rs
```diff
@@ -69,3 +90,107 @@ pub fn follow_redirects(
...
+                            );
+                            return false;
+                        }
+                    }
+
+                    true
+                });
+
+                if addrs.is_empty() {
+                    Either::A(Err(ResolverError::InvalidInput("FUCK")).into_future())
```
jan-auer: Nice language, please.

src/config.rs
untitaker: `RESERVED_IP_BLOCKS` is a standard, see https://en.wikipedia.org/wiki/Reserved_IP_addresses. The value in sentry also corresponds to that standard, and ops themselves have talked about making it a bool in sentry because there is no reason to modify it.

src/http.rs
untitaker: Fixed!!

src/http.rs
untitaker: I think we want to make it visible when a user tries to screw with us like that, also there are usability issues that come from having security by default that I want to work around with by using an error log (user tries symbolicator out locally and wonders why their local bucket doesn't receive any requests)

untitaker: Discussed f2f, user needs to look into debug logs anyway if files could not be fetched (because it can be a lot of things that went wrong, not just this)

PR #104, opened 2019-07-31
fix(docker): chown default data directory

tonyo:


docker-entrypoint.sh
```diff
@@ -2,7 +2,10 @@
 set -eu
 
 if [ "$(id -u)" == "0" ]; then
-  echo '127.0.0.1 api.local.sentry.io' >> /etc/hosts  # For tests
+  # Prepare default data directory
+  mkdir -p /data
+  chown symbolicator:symbolicator /data
```
tonyo: Not doing "-R" here because the directory might contain quite a lot of files. If permissions for some files inside are wrong, it's better to fix them manually and find the reason of that.

tonyo:
🤦‍♂ You're right

PR #106, opened 2019-08-04
ref: Upgrade to Actix 1

jan-auer:
This refactors the entire web layer including middlewares and restructures services. To speed up compilation, many `impl Trait` have been replaced with Boxes. Also, the library and binary crate have been merged into one. Rust tests also need less hacks now.

The inner works of services (e.g. symbol lookup) has not been touched!

Since the diff is pretty unusable, I suggest to review the outcome directly. Ideally, start from `main`, follow to `server::run` and `Service::create`, and then look at the endpoints and middlewares.

The Sentry middleware and utilities can later be moved into a new version of `sentry-actix`, and I would also propose to create a new crate, `sentry-futures`.


jan-auer:
Tests are green now 🤷‍♂ 

@untitaker See https://github.com/getsentry/symbolicator/pull/106/commits/0b0cc2afbfcc587bd15211c3b85fca78356b4436. This commit contains the fix for failing CAB extraction. The doc comments describe the reason for this fix. Summary: The file is only namable, until `NamedTemporaryFile` is dropped.

jan-auer:
FWIW, test failures were caused by file descriptor limits. With the logs at hand this was not completely obvious, but we can work on surfacing them better.

PR #110, opened 2019-08-12
feat(config): Default cache_dir to /data on Docker

BYK:
On our Docker images, we already define `/data` as a persistent volume, even more, it is quite tricky to mount a new cache directory as a volume with the right symbolicator user so make that the default on Docker. This also allows us to run symbolicator with the default config in getsentry/onpremise#220.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -29,15 +29,15 @@ settings in this case.
 Write this to a file (`config.yml`):
 
 ```yaml
-cache_dir: '/tmp/symbolicator'
-bind: '0.0.0.0:3021'
+cache_dir: "/tmp/symbolicator"
+bind: "0.0.0.0:3021"
 logging:
-  level: 'info'
-  format: 'pretty'
+  level: "info"
+  format: "pretty"
   enable_backtraces: true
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -29,34 +29,35 @@ settings in this case.
 Write this to a file (`config.yml`):
 
 ```yaml
-cache_dir: "/tmp/symbolicator"
-bind: "0.0.0.0:3021"
+cache_dir: '/tmp/symbolicator'
```
BYK: Happy to revert these style changes but we force autoFormat on our repo for VSCode and this is how Prettier wants our md files to look like :D

jan-auer: Most changes are fine, but please revert the indentation level. Unfortunately, `mkdocs` is kinda inflexible and doesn't allow to change the indentation level (afaik), and this would break rendering of the docs. I'm also fine if you prefer to swap the markdown renderer :)

BYK: Yeah, prettier is quite weird when it comes to indentation. Will revert.

PR #111, opened 2019-08-13
ref(docker): Various improvements on Docker image

BYK:
This patch does the following:

- Upgrade to debian buster images
- Move smoke tests to `cloudbuild.yaml`, reducing layers on prod image
- Remove unnecessary `/etc/symbolicator` volume
- Fix `/data` ownership change logic
- Optimize build stage caching a bit
- Fix the incorrect 'modified' suffix in version due to partial git copying


Dockerfile
```diff
@@ -31,15 +32,14 @@ ENV \
 RUN groupadd --system symbolicator --gid $SYMBOLICATOR_GID \
     && useradd --system --gid symbolicator --uid $SYMBOLICATOR_UID symbolicator
 
-RUN mkdir /etc/symbolicator /data \
-    && chown symbolicator:symbolicator /etc/symbolicator /data
-VOLUME ["/etc/symbolicator", "/data"]
+VOLUME ["/data"]
+RUN mkdir /etc/symbolicator && \
+    chown symbolicator:symbolicator /etc/symbolicator /data
+ONBUILD COPY --chown=symbolicator:symbolicator config.yml /etc/symbolicator/config.yml
```
tonyo: Could you elaborate on why we need this?

Dockerfile
```diff
@@ -31,15 +32,14 @@ ENV \
...
 
 EXPOSE 3021
 
-COPY --from=symbolicator-build /usr/local/bin/symbolicator /bin
-# Smoke test
-RUN symbolicator --version && symbolicator --help
+COPY --from=symbolicator-build --chown=symbolicator:symbolicator /usr/local/bin/symbolicator /bin
```
tonyo: Any specific reason why we need that `chown` here?

cloudbuild.yaml
```diff
@@ -37,6 +37,14 @@ steps:
         ".",
       ]
 
+  - name: "us.gcr.io/$PROJECT_ID/symbolicator:latest"
+    args: ['--version']
+    id: "smoke test 1"
```
tonyo: That's nice 👍

docker-entrypoint.sh
```diff
@@ -2,8 +2,8 @@
 set -eu
 
 if [ "$(id -u)" == "0" ]; then
-  # Prepare default data directory
-  mkdir -p /data
+  # Prepare default data and config directories
+  chown -R symbolicator:symbolicator /etc/symbolicator
```
tonyo: I don't think this is necessary. `/etc/symbolicator` isn't meant to be writable (e.g., we currently mount it in read-only mode), so it can potentially break things.

Dockerfile
BYK: This one is not really necessary. I added it for good hygiene but thinking again, I guess it is actually better to leave it as root?

Dockerfile
BYK: This just allows any downstream users of the image to bake in their own config, instead of using volumes. This is better because volumes are host-bound by definition whereas a config copied in like this is baked into the new image, making distribution easier and safer for consistency.

This follows the same patter from https://github.com/getsentry/docker-sentry/blob/master/git/onbuild/Dockerfile

docker-entrypoint.sh
BYK: WFM, as long as the files are readable by symbolicator user by default (and I think that is the case right now).

cloudbuild.yaml
tonyo: @BYK Just to confirm, are you sure that the recently built image will be used, and not the remote image with the tag `latest`?

Dockerfile
tonyo: I don't mind this, but for my understanding: if `config.yml` is absent during the downstream build, will the build fail?

cloudbuild.yaml
tonyo: ^

Dockerfile
BYK: Nope it won't (and I tested this).

>The ONBUILD instruction adds to the image a trigger instruction to be executed at a later time, when the image is used as the base for another build. The trigger will be executed in the context of the downstream build, as if it had been inserted immediately after the FROM instruction in the downstream Dockerfile.

BYK: Just checked, there are some tricks to make this optional but none are great. We deploy these images directly and I'd argue anyone building a new image off of this should bake the config in (instead of providing it via a volume). That said if this becomes an issue, we can very easily separate this and create a new tag with the `-onbuild` suffix, which seems to be the consensus.

Will leave the final decision to you as I'm okay with both, with a bias towards simplicity.

cloudbuild.yaml
BYK: Since we tag the image locally, that should be the one used. There are no clear examples but this page suggests what I do is "the right way": https://cloud.google.com/cloud-build/docs/configuring-builds/build-test-deploy-artifacts#running_unit_tests_and_integration_tests

tonyo: Cool

Dockerfile
tonyo: If we aim at simplicity, I'd suggest we omit both ONBUILD and VOLUME for `/etc/symbolicator`. This way, it's easy to use your own config for downstream images and there are no hidden requirements of having `config.yml` in the same directory.

I refreshed my understanding of VOLUME directive ([this SO thread](https://stackoverflow.com/questions/52570093/what-is-the-practical-purpose-of-volume-in-dockerfile) summarizes it nicely), and now I see no strong reason of using VOLUME basically anywhere, since it cannot be undeclared in downstream images and just makes things more complicated without any real benefit.

BYK: Alright, removing the `ONBUILD` as any downstream user can easily add their own config file via `COPY`. No need to make that compulsory.

PR #136, opened 2019-10-08
ref(threadpools): Switch to tokio runtime for thread pools

jan-auer:
**Please review carefully.**

Switches from `tokio-threadpool` to `tokio::Runtime` for thread pools. Also, wraps thread pools into our own type so that we abstract over the implementation and can more easily spawn futures.

Also, this thread pool reintroduces test mode. In this mode, all futures are spawned on the same thread. This has the advantage that `cargo test` can capture output correctly and we get more deterministic test results. In E2E tests, this is disabled of course.


jan-auer:
Note that with this implementation we used to see less CPU utilization in the past. I will be monitoring deployment of this PR.

PR #140, opened 2019-10-09
build(docker): Revert to debian stretch-slim

jan-auer:
There is a suspected regression in OpenSSL `1.1.1c` used on Debian buster-slim that might cause a memory leak. We suspect that this might have to do with HTTP2 connections, but it has not been confirmed yet.

cc @BYK

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ branches:
   only:
     - master
     - /^release\/[\d.]+$/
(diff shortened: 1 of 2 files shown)


jan-auer:
What ever it was exactly, memory consumption is stable. 
cc @mitsuhiko you've got a diff to play with :)

PR #147, opened 2019-10-28
ref: Strip debug information for server-side symbolication

jan-auer:
This moves to server-side symbolication for errors and crashes happening in symbolicator. This will allow us to use **source context** and improve runtime-performance of errors, as symbolicator does not have to load its own debug information when stack unwinding.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -794,6 +794,15 @@ version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "findshlibs"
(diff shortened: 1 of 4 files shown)


jan-auer:
Seems fine, but Microsoft Symbol Server is currently broken. This breaks one of our integration tests. Assuming that the Docker image builds fine, I think we can still merge. @tonyo wdyt?

jan-auer:
Verified source archives, merging now as tests are just broken due to MS Symbol Server failures. We'll tackle those separately, depending on how long this situation persists.

PR #157, opened 2019-11-21
ci: Split tests into separate jobs

jan-auer:
The old test job first builds the test target and then rebuilds the binary target to run integration tests. Since this is running serially, it might take longer than needed. By running two jobs, we might speed up builds slightly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,10 @@ matrix:
       script: make -e lint
 
     - name: "test"
-      script: make -e test
+      script: make -e test-rust
+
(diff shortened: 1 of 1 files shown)


tonyo:
I don't get this. It's 11 min in master vs 2 x 10 min in this PR 🤔

jan-auer:
Hm, doesn't help as much as I hoped. Do we still want to keep integration tests and unit tests separated?

tonyo:
Yes, if we commit to writing more tests 😉

PR #158, opened 2019-11-22
build(docker): Push images to Docker Hub too

BYK:
Ported logic from https://github.com/getsentry/sentry/blob/master/docker/cloudbuild.yaml

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -47,11 +47,23 @@ steps:
     id: "smoke test 2"
 
   # Only tag "latest" when on master
-  - name: "gcr.io/cloud-builders/docker"
-    entrypoint: "bash"
+  - name: 'gcr.io/cloud-builders/docker'
+    secretEnv: ['DOCKER_PASSWORD']
+    entrypoint: 'bash'
     args:
-      - "-c"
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -47,11 +47,23 @@ steps:
...
+    - '-c'
+    - |
+      # Only push to Docker Hub from master
+      [ "$BRANCH_NAME" != "master" ] && exit 0
+      docker push us.gcr.io/$PROJECT_ID/symbolicator:latest
+      echo "$$DOCKER_PASSWORD" | docker login --username=sentrybuilder --password-stdin
+      docker tag us.gcr.io/$PROJECT_ID/symbolicator:$COMMIT_SHA getsentry/symbolicator:$SHORT_SHA
+      docker push getsentry/symbolicator:$SHORT_SHA
+      docker tag us.gcr.io/$PROJECT_ID/symbolicator:$COMMIT_SHA getsentry/symbolicator:$COMMIT_SHA
+      docker push getsentry/sentsymbolicatorry:$COMMIT_SHA
```
tonyo:
```suggestion
      docker push getsentry/symbolicator:$COMMIT_SHA
```

BYK: I mean, I was _a bit_ tired when I made this commit but still, wow. Thanks for catching this embarrassment :)

PR #159, opened 2019-11-27
fix: Do not let IO errors of sentry index request bubble up

untitaker:
Also add a few log stmts to be able to debug this.

### DIFF
--- a/src/actors/objects/mod.rs
+++ b/src/actors/objects/mod.rs
@@ -9,8 +9,11 @@ use std::time::Duration;
 use bytes::Bytes;
 use failure::{Fail, ResultExt};
 
+#[rustfmt::skip]
 use ::sentry::configure_scope;
+#[rustfmt::skip]
 use ::sentry::integrations::failure::capture_fail;
+
 use futures::{future, Future, IntoFuture, Stream};
 use symbolic::common::ByteView;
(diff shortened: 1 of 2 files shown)


src/actors/objects/mod.rs
```diff
@@ -9,15 +9,19 @@ use std::time::Duration;
 use bytes::Bytes;
 use failure::{Fail, ResultExt};
 
+#[rustfmt::skip]
 use ::sentry::configure_scope;
+#[rustfmt::skip]
```
untitaker: my rustfmt tried to reformat this to

```
use sentry::...
```

which resolves to the local module instead of the sdk crate

src/actors/objects/mod.rs
```diff
@@ -555,60 +559,73 @@ impl ObjectsActor {
...
                 .into_iter()
-                .flatten()
                 .enumerate()
                 .min_by_key(|(i, response)| {
                     // Prefer files that contain an object over unparseable files
                     let object = match response {
                         Ok(object) => object,
-                        _ => return (3, *i),
+                        Err(e) => {
+                            log::trace!("Failed to download: {}", LogError(e));
```
jan-auer:
```suggestion
                            log::debug!("Failed to download: {}", LogError(e));
```

Trace won't make it into breadcrumbs iirc.

PR #160, opened 2019-11-28
ref: Add futures 0.3, use async-await in symbolication actor

untitaker:
I would like to commit this to master and figure out the story for `future_metrics!` and other things on the go. It seems like we can freely mix this and gradually migrate to async-await.

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -297,21 +297,20 @@ impl SourceLookup {
             }
         }
 
-        let mut results = Vec::new();
         let mut futures = Vec::new();
 
         for (i, (mut object_info, _)) in self.inner.into_iter().enumerate() {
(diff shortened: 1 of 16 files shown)


untitaker:
This has been tested in prod for a couple hours now.

src/actors/symbolication.rs
```diff
@@ -10,6 +10,8 @@ use bytes::{Bytes, IntoBuf};
 use failure::Fail;
 use futures::future::{self, join_all, Either, Future, IntoFuture, Shared};
 use futures::sync::oneshot;
+use futures03::compat::Future01CompatExt;
+use futures03::future::{FutureExt as FuckYou, TryFutureExt};
```
jan-auer:
```suggestion
use futures03::future::{FutureExt as _, TryFutureExt};
```

PR #179, opened 2020-01-08
ref(futures): Migrate symbolication actor to futures 0.3

jan-auer:
A first pass over the symbolication "actor" to migrate it to futures 0.3 and async / await. The implementation is still quite messy and ugly, but can be cleaned up now.


jan-auer:
This has been implemented in other PRs.

PR #182, opened 2020-01-16
feat(config): Set default metrics statsd from environment var

slve:
Pick up statsd server config from ~METRICS_STATSD~ STATSD_SERVER environment variable.
Statsd config in the config.yml will override the value picked up from env var.


untitaker:
A quick google search doesn't show much precedence for that choice of envvar name. Where did you take METRICS_STATSD from?

slve:
The environment name is just one I've came up with, based on the structure of the config.yml to keep something constistent with.
Sure, let me add it to the docs, once agreed on a better naming.
Did a bit of a lookup, but not yet closer to a great standardized name,
what do you think about naming it STATSD_SERVER?

untitaker:
I have to ask, why do you need this in the first place?

slve:
Great, thanks!
We are running it in kubernetes and the statsd server may change over time, I wanted to avoid generating config.yml from a template.

untitaker:
@slve can you rebase, a lot of stuff changed since then. I think the docs may have moved to getsentry/sentry-docs

Sorry for dropping the ball on this.

flub:
I thought that getting an env var into a kubernetes pod still requires you to add it to the yaml of the pod spec?  Does this actually save on templating?

It would be nice to have a slightly more generic way to set things in the configuration instead of the statsd-specific thing.  Especially because this doesn't seem to be a common standard.  But there is also a nice simplicity to the current approach, and this seems to me more like a problem that should be solved in the infrastructure (k8s) rather than the application.

slve:
@untitaker Thanks for getting back to me.
- I've merged in the latest master.
- Looked into getsentry docs but could only find it mentioned here https://github.com/getsentry/sentry-docs/pull/1042

@flub Yes, templating is still needed to have the appropriate kubernetes manifests.
We have a standard solution using kapitan, where all of our manifests are generated.
The static configurations live inside kubernetes configmaps,
but a server address, like statsd here, may differ on the actual node the pod is deployed to.
A deployment manifest snippet can be something like this below,
while, as far as I am aware, there is no solution to do the same with configmaps.
```
env:
  - name: STATSD_HOST
    valueFrom:
      fieldRef:
        apiVersion: v1
        fieldPath: status.hostIP
  - name: STATSD_SERVER
    value: $(STATSD_HOST):8025
```
As for the second part, I checked other configs to see whether they could be as dynamic as the statsd host,
but I think this is the only one we don't have control over in kubernetes lifecycle.
I see that there is no standard in symbolicator to use environment variables as of yet,
still it may be just fine for now if there isn't much similar requests.

untitaker (approved):
please add a changelog entry here: https://github.com/getsentry/symbolicator/blob/master/CHANGELOG.md

I think this is fine to merge. At a later point we can move to a standardized scheme for envvars. We don't really have obligations for backwards-compat here.

flub:
there's a chance that #288 fixes some of the test failures.  would you mind merging master again to see if this works?

flub:
didn't quite work yet, I'm having another go in https://github.com/getsentry/symbolicator/pull/291 but there's no need to further hold this up.  Thanks @slve !

slve:
Thank you both!

PR #210, opened 2020-04-08
feat(metric): Add metrics to track OAuth token usage

flub:
I'm a little unsure of this metrics choice.  The sum of these is probably a proxy for number of requests made?  But that's really not obvious which is why I don't like this so much.  Which makes me wonder if it should really be tracked in a metric in the caller.  The `token.requests` metric would then stand on its own.

### DIFF
--- a/src/actors/objects/gcs.rs
+++ b/src/actors/objects/gcs.rs
@@ -124,12 +124,14 @@ fn get_token(
 ) -> Box<dyn Future<Item = Arc<GcsToken>, Error = ObjectError>> {
     if let Some(token) = GCS_TOKENS.lock().get(source_key) {
         if token.expires_at >= Utc::now() {
+            metric!(counter("source.gcs.token.cached") += 1);
             return Box::new(Ok(token.clone()).into_future());
         }
     }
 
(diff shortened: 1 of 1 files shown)


jan-auer:
@flub Since it's really one metric, you could use tags instead, like we do for object requests:

https://github.com/getsentry/symbolicator/blob/dbc714db90b5ce032fe1c27aea1ce3c4af04b4da/src/actors/common/cache.rs#L205-L208

flub:
My first version did that... but I changed it because in that case a single metric tracks two widely different things: requesting a new token vs using a cached token.  It could work maybe work with a bit better naming like a `source.gcs.token_use` metric with `source:cache` & `source:request` as tags.

jan-auer:
Gotcha, in that case, maybe remove the `source.gcs.token.cached` and make it a generic one that just tracks GCS accesses in general. At your discretion, either way is fine with me.

flub:
I looked into separating the counters more logically, but the current success/error flow is too complicated to easily do that.  I think this will do.

(out of habit did an unnecessary pull-rebase-push cycle so now waiting for ci again before merging)

PR #218, opened 2020-05-04
Propagate the spawn error from the process pool up at startup

flub:
This newtypes the error in the ServiceState but doesn't bother in the
SymbolicationActor itself.  It's a bit more graceful compared to the
plain .expect() which is probably nice.

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -55,19 +55,6 @@ lazy_static::lazy_static! {
     static ref OS_MACOS_REGEX: Regex = Regex::new(r#"^Mac OS X (?P<version>\d+\.\d+\.\d+)( \((?P<build>[a-fA-F0-9]+)\))?$"#).unwrap();
 }
 
-/// Variants of [SybolicationCreationError].
-#[derive(Debug, Fail)]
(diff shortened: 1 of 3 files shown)


src/actors/symbolication.rs
```diff
@@ -137,17 +137,17 @@ impl SymbolicationActor {
         symcaches: SymCacheActor,
         cficaches: CfiCacheActor,
         threadpool: ThreadPool,
-    ) -> Self {
+    ) -> Result<Self, procspawn::SpawnError> {
```
jan-auer, line 140: Might be nice to wrap this in our own error here already so that the caller no longer has to deal with `procspawn`. It seems like an implementation detail of `SymbolicationActor`.

src/server.rs
```diff
@@ -13,6 +13,8 @@ pub enum ServerErrorKind {
     /// Failed to bind to a port.
     #[fail(display = "failed to bind to the port")]
     Bind,
+    #[fail(display = "failed to create process pool")]
+    Spawn,
```
jan-auer, line 17: Can we give this a different name and description? Later on, there might be other errors thrown in `ServiceState::Create` that are unrelated to process pools.

PR #223, opened 2020-05-12
fix(download): Do not panic on malformed URIs

flub:
Building the HTTP request could fail when the URI provided is invalid,
which apparently sometimes happens.  This caters for this and
propagates the correct error up.

### DIFF
--- a/src/utils/http.rs
+++ b/src/utils/http.rs
@@ -35,8 +35,7 @@ where
     let state = (initial_url, max_redirects, true);
     Box::new(future::loop_fn(state, move |(url, redirects, trusted)| {
         let mut request = tryf!(make_request(url.as_str()).map_err(|err| {
-            let fail =
-                failure::format_err!("Bad URI when creating request: {} (URI: {})", err, url);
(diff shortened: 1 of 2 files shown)


src/utils/http.rs
```diff
@@ -28,19 +30,24 @@ pub fn follow_redirects<F>(
...
 where
-    F: Fn(&str) -> ClientRequest + Send + 'static,
+    F: Fn(&str) -> Result<ClientRequest, actix_web::error::Error> + Send + 'static,
 {
     let state = (initial_url, max_redirects, true);
     Box::new(future::loop_fn(state, move |(url, redirects, trusted)| {
-        let mut request = make_request(url.as_str());
+        let mut request = tryf!(make_request(url.as_str()).map_err(|err| {
+            let fail =
+                failure::format_err!("Bad URI when creating request: {} (URI: {})", err, url);
```
jan-auer, line 39: This can also be an error other than URI parsing, should we write a more generic message here?

flub: good point, improved

PR #224, opened 2020-05-13
ref(symbolication): Avoid races and leaks in task handling

jan-auer:
This simplifies symbolication task creation and also prevents races or task leaks:

- Most notably, this introduces `CallOnDrop`, a type that executes a callback when it is dropped. Instead of relying on the symbolication future to call its final `then` callback, we move `CallOnDrop` into the future, which eventually evicts the task from the global computation map. This is safer when the future is dropped without polling.
- When creating a new symbolication task, we assume that there are no UUID collisions.
- The signature of `create_symbolication_request` has been simplified since it is infallible.
- All use of `impl Future` has been removed in favor of boxed future types. This should help with compilation times, and also makes the code slightly more readable in some areas.

There are two TODOs left:
 1. When creating a symbolication task, the initial factory is run synchronously. Fixing this requires a refactoring of some methods to make them `Send`.
 2. The symbolication task is spawned into the web arbiter's thread, which could lead to delays. Instead, this should run in a dedicated background thread or thread pool.

See https://github.com/getsentry/symbolicator/pull/112 for the original PR

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -9,9 +9,9 @@ use apple_crash_report_parser::AppleCrashReport;
 use bytes::{Bytes, IntoBuf};
 use failure::{Fail, ResultExt};
 use futures::{compat::Future01CompatExt, FutureExt as _, TryFutureExt};
-use futures01::future::{self, join_all, Either, Future, IntoFuture, Shared};
+use futures01::future::{self, join_all, Future, IntoFuture, Shared};
 use futures01::sync::oneshot;
-use parking_lot::RwLock;
+use parking_lot::Mutex;
 use regex::Regex;
 use sentry::integrations::failure::capture_fail;
 use serde::{Deserialize, Serialize};
@@ -24,7 +24,7 @@ use symbolic::minidump::cfi::CfiCache;
 use symbolic::minidump::processor::{
     CodeModule, CodeModuleId, FrameTrust, ProcessMinidumpError, ProcessState, RegVal,
 };
-use tokio::prelude::FutureExt;
+use tokio::timer::Delay;
 
 use crate::actors::cficaches::{
     CfiCacheActor, CfiCacheError, CfiCacheErrorKind, CfiCacheFile, FetchCfiCache,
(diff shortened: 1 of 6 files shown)


src/actors/symbolication.rs
```diff
@@ -156,83 +159,87 @@ impl SymbolicationActor {
...
+        let (sender, receiver) = oneshot::channel();
 
-        let (tx, rx) = oneshot::channel();
+        let requests = self.requests.clone();
+        let request_id = RequestId::new(uuid::Uuid::new_v4());
 
-        self.requests.write().insert(request_id, rx.shared());
+        // Assume that there are no UUID4 collisions in practice.
+        let evicted = requests.lock().insert(request_id, receiver.shared());
+        debug_assert!(evicted.is_none());
```
flub, line 206: We don't yet have any `debug_assert!()` calls.  In Python I've learned to avoid these constructs (`if __debug__`) because this almost always turns out to not work well.  Either the invariant holds and things are fine, or we silently have errors in production.  And in this case collisions are much more likely to occur in production than when testing so this will likely never do much good.

src/utils/futures.rs
```diff
@@ -78,3 +78,29 @@ impl ThreadPool {
...
+    pub fn new<F: FnOnce() + Send + 'static>(f: F) -> CallOnDrop {
+        CallOnDrop {
+            f: Some(Box::new(f)),
+        }
+    }
+}
+
+impl Drop for CallOnDrop {
+    fn drop(&mut self) {
+        debug_assert!(self.f.is_some(), "CallOnDrop cannot be empty");
```
flub, line 101: same general comment on `debug_assert`, though here it seems less of a footgun

src/actors/symbolication.rs
jan-auer: Fair point. I'm making the following distinction here:
- Be defensive in production. The case will still be handled correctly, so I would not want to error out.
- Be strict in testing, to detect regressions early on.

WDYT?

flub: Isn't swallowing this silently what we do in production now?

If I understand correctly, if you happen to have evicted a request (so unlikely that this is all highly theoretical) then when the future of that evicted request gets dropped this request will be removed from the set of requests.  This means anyone polling for it will get an error saying this request doesn't exist.  Both the original request and this request will have its receiver dropped and sends on them will fail.  So I think when a collision happens both the affected requests are lost.

I'm fine with accepting this, it is extremely unlikely.  I'm still not sure panicing in debug mode does much to improve the situation though.

flub: I'm also not objecting to this btw, I did approve the PR after all :)

flub: While on a short walk I also realised that another way of looking at this is that if it is assertable under debug, then it could be written as a unit test.

jan-auer: You convinced me, I'll remove them :)

PR #225, opened 2020-05-13
ref(cache): Avoid races and leaks in file handling

jan-auer:
Refactors the cache code and fixes race conditions in cache handling: 

- Most importantly, it keeps the lock while checking and inserting into the cache, and creating a computation future. 
- The response is sent into the channel *after* dropping it from the map. This ensures that clients retrieving an existing channel always get a response.
- Instead of panicking, emit an error if computation fails for any reason. This is usually only the case if the arbiter restarts or a panic occurs. Otherwise, we always receive an explicit Result and send that into the channel.

Follow-up to https://github.com/getsentry/symbolicator/pull/224
Adds cache changes from https://github.com/getsentry/symbolicator/pull/112

### DIFF
--- a/src/actors/common/cache.rs
+++ b/src/actors/common/cache.rs
@@ -5,22 +5,23 @@ use std::io;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
-use futures01::future::{lazy, Future, IntoFuture, Shared};
+use actix::ResponseFuture;
+use futures01::future::{self, Future, Shared};
 use futures01::sync::oneshot;
-use parking_lot::RwLock;
-use sentry::configure_scope;
+use parking_lot::Mutex;
 use symbolic::common::ByteView;
 use tempfile::NamedTempFile;
 
 use crate::cache::{get_scope_path, Cache, CacheKey, CacheStatus};
 use crate::types::Scope;
+use crate::utils::futures::CallOnDrop;
 use crate::utils::sentry::SentryFutureExt;
 
 // Inner result necessary because `futures::Shared` won't give us `Arc`s but its own custom
 // newtype around it.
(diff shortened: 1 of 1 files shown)


src/actors/common/cache.rs
```diff
@@ -167,138 +172,138 @@ impl<T: CacheItemRequest> Cacher<T> {
...
+        }));
 
-                metric!(
-                    counter(&format!("caches.{}.file.write", name)) += 1,
-                    "status" => status.as_ref(),
-                );
+        // Run the computation and wrap the result in Arcs to make them clonable.
+        let channel = future::lazy(move || slf.compute(request, key))
+            .then(move |result| {
+                drop(remove_computation_token);
```
Swatinem, line 263: might be a good idea to also attach your comment here about the order. What is the expectation when this is dropped for other reasons?

jan-auer: Added, thanks!

src/actors/common/cache.rs
```diff
@@ -167,138 +172,140 @@ impl<T: CacheItemRequest> Cacher<T> {
...
-            metric!(counter(&format!("caches.{}.file.miss", name)) += 1);
+            let byteview = ByteView::open(temp_file.path())?;
 
-            let cache_path = get_scope_path(slf.config.cache_dir(), &key.scope, &key.cache_key);
+            metric!(
+                counter(&format!("caches.{}.file.write", name)) += 1,
+                "status" => status.as_ref(),
+            );
+            metric!(
+                time_raw(&format!("caches.{}.file.size", name)) = byteview.len() as u64,
```
flub, line 232: So there's a comment in the macro saying that `time_raw` is intentional for file sizes.  But why?

jan-auer: We essentially use timer metrics across the board for measuring distributions of individual things. Potentially it would be better to use histograms.

PR #226, opened 2020-05-14
fix(download): Percent-encode HTTP paths with unicode names

jan-auer:
Some symbol server layouts, such as SSQP and Microsoft's SymStore, contain file names in the path. These paths do not necessarily have to be valid URL paths and may contain special characters. To join them, the path must be percent-encoded.

This switches from `Url::join`, which assumes a valid URL path, to `Url::path_segments_mut`, which percent-encodes the path properly.

### DIFF
--- a/src/actors/objects/http.rs
+++ b/src/actors/objects/http.rs
@@ -4,9 +4,10 @@ use std::time::Duration;
 use actix_web::http::header;
 use actix_web::{client, HttpMessage};
 use failure::Fail;
-use futures01::{Future, IntoFuture, Stream};
+use futures01::{future, Future, Stream};
 use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use tokio_retry::Retry;
+use url::Url;
 
 use crate::actors::objects::common::prepare_download_paths;
 use crate::actors::objects::{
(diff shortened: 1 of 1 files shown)


src/actors/objects/http.rs
```diff
@@ -32,17 +51,17 @@ pub(super) fn prepare_downloads(
...
     source: Arc<HttpSourceConfig>,
     download_path: &DownloadPath,
 ) -> Box<dyn Future<Item = Option<DownloadStream>, Error = ObjectError>> {
-    // XXX: Probably should send an error if the URL turns out to be invalid
-    let download_url = match source.url.join(&download_path.0) {
+    // This can effectively never error since the URL is always validated to be a base URL.
+    let download_url = match join_url_encoded(&source.url, &download_path.0) {
         Ok(x) => x,
-        Err(_) => return Box::new(Ok(None).into_future()),
+        Err(_) => return Box::new(future::ok(None)),
```
Swatinem, line 64: is this something worth logging? so we catch any invalid things that might flow through here sometime in the future?

jan-auer: As the comment above states, this is something that "cannot happen" ™️  since the base URL is validated beforehand already. I just don't like to write `.unwrap()`.

As for the additional tests: Could add, but we already test the space, which shows that percent encoding works. I'm not sure if we should be unit-testing the `url` crate.

PR #228, opened 2020-05-25
Initial sketch of download service

flub:
This is now ready for merging.  It could do with better local testing however but this is probably better in a followup PR.


src/services/download/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    ///
+    /// `dest`: Pathname of filename to save the downloaded file into.
+    pub fn download(
+        &self,
+        source: SourceId,
+        dest: PathBuf,
+    ) -> Box<dyn Future<Item = PathBuf, Error = DownloadError> + Send + 'static> {
+        let fut03 = self.worker.spawn(|| async move {
+            match source {
+                SourceId::Http(source, loc) => {
```
Swatinem, line 51: you can move the match outside. That way, you can even avoid having to spawn anything in that case.

I would also very much advocate to make any *new* code be `std::Future`.

src/services/download/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+                    http::download_source(source, loc, dest).compat().await
+                }
+                _ => Err(DownloadErrorKind::Tmp.into()),
+            }
+        });
+        let fut01 = fut03
+            .then(|spawn_ret| {
+                let result = match spawn_ret {
```
Swatinem, line 59: The reason for this is that `fut03` is essentially a `Result<Result<T, E>, RemoteCanceled>`.
I think you should be able to `.unwrap_or(Err(E))` to flatten that.

src/services/download/http.rs
```diff
@@ -0,0 +1,264 @@
...
+//                 })
+//                 .map(|_| future::ok(dest));
+//             Box::new(accumulator)
+//         });
+//     // Box::new(ret)
+//     Box::new(ret) as Box<dyn Future<Item = PathBuf, Error = DownloadError>>
+// }
+
+// The actors::objects::http::download_from_source with minimal changes.
+fn download_from_source(
```
Swatinem, line 154: since this is local to this file, maybe its a prime candidate to transform to an `async fn`

src/services/download/http.rs
```diff
@@ -0,0 +1,264 @@
...
+    source: Arc<HttpSourceConfig>,
+    download_path: &SourceLocation,
+) -> Box<dyn Future<Item = Option<DownloadStream>, Error = DownloadError>> {
```
Swatinem, line 157: just curious why we have an `Option` here, instead of another `DownloadError` variant.

src/services/download/http.rs
flub: Yep, that's the basic idea behind this that these become isolated enough to safely and easily transform to new-style code.

src/services/download/http.rs
flub: I did wonder that too... this is the bit of code i kept as unchanged as possible.  the glue that calls this transforms it into a DownloadError

flub: So this is to not make a 404 an error.  On the service-level interface i've replaced this with a more obvious enum instead.

flub:
This now includes a huge split from src/types.rs into src/sources.rs.  This because I wanted to re-use the `FileId` type from actors/objects/mod.rs in services/downloads/mod.rs.  Furthermore it contains more then just types, having a bunch of impls on those as well.  So I think this is a good thing to split up (types.rs was also huge anyway).

However, I'm up for doing that as a no-change PR separately and rebasing this on top instead of doing it here directly if that'd be preferred.

src/actors/objects/mod.rs
```diff
@@ -257,131 +257,80 @@ impl CacheItemRequest for FetchFileDataRequest {
...
+            .map_err(Into::into);
 
-                        future::Either::B(future)
-                    }
-                };
-
-                let future = download_file.and_then(clone!(threadpool, |download_file| {
+        let result = request.and_then(move |status| -> Result<CacheStatus, ObjectError> {
+            match status {
+                crate::services::download::DownloadStatus::Completed => {
```
Swatinem, line 269: maybe import this enum? ;-)

src/actors/objects/mod.rs
```diff
@@ -257,131 +257,80 @@ impl CacheItemRequest for FetchFileDataRequest {
...
-                            .map(|file| DownloadedFile::Local(path.clone(), file))
-                            .context(ObjectErrorKind::Io)
-                            .map_err(ObjectError::from)
-                            .into_future();
+        let download_dir = tryf!(path.parent().ok_or(ObjectErrorKind::NoTempDir)).to_owned();
+        let download_file =
+            tryf!(NamedTempFile::new_in(&download_dir).context(ObjectErrorKind::Io));
+        let request = download_svc
+            .download(self.0.file_id.clone(), download_file.path().to_owned())
+            .map_err(Into::into);
```
Swatinem, line 265: I’m a bit surprised you can `map_err` on the future here, but I guess thats a futures 0.1 thing.

src/app.rs
```diff
@@ -39,6 +39,8 @@ pub struct ServiceState {
     objects: ObjectsActor,
     /// The config object.
     config: Arc<Config>,
+    /// The download service.
+    download_svc: Arc<crate::services::download::Downloader>,
```
Swatinem, line 43: maybe import this type here as well

flub: You mean `Downloader` instead of `crate::services::download::Downloader`?  I kind of like the prefix, says a lot more about the thing, I considered `use crate::services` though.  But the crate prefix doesn't bother me that much and there's only one service so far.

But this does remind me that I think actual rust naming guidelines would suggest `services::download::Service` which is a rename I was still considering.

src/services/download/types.rs
```diff
@@ -15,35 +15,29 @@ use crate::types::{
...
     #[fail(display = "bad file destination")]
     BadDestination,
-    #[fail(display = "failed to create temporary download file")]
-    TempFile,
     #[fail(display = "failed writing the downloaded file")]
     Write,
-    #[fail(display = "temporary fudge error, not to be commited")]
+    #[fail(display = "download was cancelled")]
+    Canceled,
+    #[fail(display = "temporary fudge error, not to be committed")]
```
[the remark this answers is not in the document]
flub: heh, yes.  though clearly it got committed it should not e merged like this.

src/app.rs
Swatinem: Ok, this is getting quite bikesheddy now. I would actually prefer `DownloadService`, and importing that from `crate::services::DownloadService`, which itself is a re-export. Well its just personal preference ;-)

flub: Nothing wrong with some friendly bikeshedding.  I think https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#avoid-redundant-prefixes-[rfc-356] (which we seem to explicitly call out on the rust notion page) is against `services::download::DownloadService` and pro `services::download::Service`.  Or at least that's my interpretation.  That doesn't say anything about a `services::DownloadService` re-export though.

Swatinem: I’m also thinking about `type_id` and `Debug`. The first includes the complete canonical path, the latter only the struct name itself. Sure, `Debug` as the name says should not really appear in production anywhere, but its still good to disambiguate there. When you want to `use` a lot of these, or have your IDE auto-import them, it also is good to have a more unique name, or at least re-export.

src/services/download/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// HTTP 404, `None` is returned.  If there is an error during the
+    /// downloading [`DownloadError`] is returned.
+    ///
+    /// [`DownloadError`]: types/struct.DownloadError.html
+    pub fn download(
+        &self,
+        source: SourceFileId,
+        dest: PathBuf,
+    ) -> Box<dyn Future<Item = DownloadStatus, Error = DownloadError> + Send + 'static> {
+        let fut03 = self.worker.spawn(|| async move {
```
flub, line 71: This toplevel spawn should probably be recording metrics for the "futures.wait_time" metric.

src/actors/objects/mod.rs
```diff
@@ -75,13 +71,24 @@ impl From<io::Error> for ObjectError {
     }
 }
 
+impl From<crate::services::download::DownloadError> for ObjectError {
```
Swatinem, line 74: maybe `use crate::services::download::{DownloadError, DownloadErrorKind}`?

src/actors/objects/mod.rs
```diff
@@ -663,16 +586,17 @@ impl ObjectsActor {
...
-    FutureStream(Box<dyn Stream<Item = Bytes, Error = ObjectError>>),
-    File(PathBuf),
-}
-
+/// Create a list of all DIF downloads to be attempted.
+///
+/// Builds the list of download locations where the relevant Debug Information
+/// Files for a given [ObjectId] might be relative to the [SourceConfig].  This
+/// does not mean that each file exists on the source.  For some sources,
+/// e.g. [SourceConfig::Sentry], this might involve querying the source.
```
Swatinem, line 594: as we found out, stable rustdoc does not auto-resolve these links :-(

src/services/download/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+/// Download the source from a streaming future.
+///
+/// These streaming futures are currently implemented per source type.
+fn download_stream_fut(
+    source_desc: String,
+    stream: Box<dyn Future<Item = Option<DownloadStream>, Error = DownloadError>>,
+    dest: PathBuf,
+) -> Box<dyn Future<Item = DownloadStatus, Error = DownloadError>> {
```
Swatinem, line 110: maybe as an excercise for later, this function is a prime candidate to convert to async/await. You are using `.compat()` up above. And inside here, you actually box twice.

Anyway, lets do this as a followup ;-)

src/actors/objects/mod.rs
flub: How strongly do you feel about this?  It's only used in 2 places in this file so is far from a central type used in this file.  Both places have no issues with line splitting/length etc, the code layout would stay the same.  It makes it very obvious that this status lives is part of the download service.  And finally, and a rather weak consideration, while when reading code in your editor this makes only a little difference, it makes a bigger difference when reading code on github where we don't have symbol hints.

I know I'm happier to use longer paths for types then seems to be usual in this code base.  So if there is a desire to keep importing types as much as reasonable then I'm fine to adjust to this.  I don't feel overly strongly, just described my thoughts when choosing the current implementation.

src/actors/objects/mod.rs
flub: yep, futures01 are essentially a Result.

src/app.rs
flub: Renamed to `DownloadService`.  Didn't go with the re-export on services yet though, the indirection doesn't yet seem worth it with just one service.  Also did some more `use` of it now, though I'm not really convinced of that.

src/actors/objects/mod.rs
flub: same logic as before: this is used in one place, not core to this module.

src/actors/objects/mod.rs
flub: good catch, i tried to update this all but stable rustdoc doesn't even seem to warn when links are broken which makes this really hard to spot :(

src/services/download/mod.rs
flub: Yes, the types with futures01 are terrible.  This double boxing is because there a match and both arms need to return the same type.  But one returns an `AndThen` type while the other returns an `FutureResult` type.  And boxing them makes them both the "same" type.

The second box is needed because otherwise it also returns an `AndThen` type, which again is the wrong type.

src/actors/objects/mod.rs
Swatinem: I tent to like short names in general. Especially in JS, you *have* to use `import`.

IMO, `std` is the only thing where it makes sense to have 3 segments, like `std::error::Error` or `std::fmt::Display`, etc.

Having 5 segments for a custom type (enum variant in this case) is a bit much. But this is very bikesheddy, and I don’t have strong opinion on this.

src/services/download/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        let (_srv, source) = test::symbol_server();
+        let source_id = match source {
+            SourceConfig::Http(source) => {
+                SourceFileId::Http(source, SourceLocation::new("hello.txt"))
+            }
+            _ => panic!("unexpected source"),
+        };
+
+        let dl_svc = DownloadService::new(RemoteThread::new_threaded());
```
flub, line 167: We can't get rid of `::new_threaded()` here because `crate::utils::futures::IS_TEST` is global and could affect other concurrent tests.  So I think `::new_threaded()` is fine.

src/services/download/mod.rs
```diff
@@ -0,0 +1,173 @@
...
+            }
+        });
+        let fut01 = fut03
+            .map(|spawn_ret| spawn_ret.unwrap_or_else(|_| Err(DownloadErrorKind::Canceled.into())))
+            .boxed()
+            .compat();
+        Box::new(future_metrics!(
+            "service.download",
+            Some((
+                std::time::Duration::from_secs(3600),
```
flub, line 96: What is a good timeout here?

src/actors/objects/mod.rs
jan-auer: This is just some bike shedding, so please disregard.

I agree with Arpad here. There are few cases where importing the module is a better choice (I usually go with this for `fmt::Debug` to avoid also importing `fmt::Formatter` and `fmt::Result`), but generally I would only qualify if it is still a short overall name. For example, `std::mem::swap` is still cool. `crate::services::download::DownloadStatus::Completed` is a bit long here.

PR #229, opened 2020-05-29
ref: Report failed function demangling to Sentry

jan-auer:
Instead of silently ignoring when a function cannot be demangled, log it to Sentry. This only reports an error if the language can be explicitly detected to avoid false-positives in case there are unmangled names marked as some other language.

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -680,7 +680,20 @@ fn symbolicate_frame(
             }
         };
 
+        // There are a few languages that we should always be able to demangle. Only complain about
+        // those but silently ignore the rest.
         let lang = line_info.language();
+        let should_demangle = matches!(lang, Language::Cpp | Language::Rust | Language::Swift);
+        let demangled_opt = line_info.function_name().demangle(DEMANGLE_OPTIONS);
+        if should_demangle && demangled_opt.is_none() {
+            sentry::with_scope(
+                |scope| scope.set_extra("identifier", line_info.function_name().to_string().into()),
+                || {
+                    let message = format!("Failed to demangle {} identifier", lang);
+                    sentry::capture_message(&message, sentry::Level::Error);
+                },
+            );
+        }
 
         rv.push(SymbolicatedFrame {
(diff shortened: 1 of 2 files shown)


src/actors/symbolication.rs
```diff
@@ -680,7 +680,20 @@ fn symbolicate_frame(
...
+        // those but silently ignore the rest.
         let lang = line_info.language();
+        let should_demangle = matches!(lang, Language::Cpp | Language::Rust | Language::Swift);
+        let demangled_opt = line_info.function_name().demangle(DEMANGLE_OPTIONS);
+        if should_demangle && demangled_opt.is_none() {
+            sentry::with_scope(
+                |scope| scope.set_extra("identifier", line_info.function_name().to_string().into()),
+                || {
+                    let message = format!("Failed to demangle {} identifier", lang);
+                    sentry::capture_message(&message, sentry::Level::Error);
```
untitaker, line 693: Do we have no capturing of error logs set up?

jan-auer: In symbolicator it's the other way around. Rather than logging unstructured logs, we mostly capture `fail` or errors. In this case, a plain message is enough. 

Now that you're mentioning it, I'm wondering if we should rather use `log::something!` though.

untitaker: it's fine. if that's the pattern (seems to be) then so be it. it's def inconsistent with relay though

jan-auer:
This needs more work. Turns out that unmangled names can be reported with C++ as language (the prime example being `main`). Such names will not demangle, but they are perfectly valid.

We can either try to detect this case, or make `symbolic_demangle` differentiate between explicit errors and `None`.

jan-auer:
Ok, this should be better. Instead of using the reported language, we now always detect the language from the plain string. The downside of this is that we no longer report if we can't detect the language (e.g. Swift introduces yet another mangling schema or we're missing a prefix for C++, ...). I think this is a good compromise to catch the common cases:

- `cpp_demangle` or `msvc_dmeangler` error on some identifier
- Swift introduces something new and we need to update the library

PR #231, opened 2020-06-04
test: Show pytest report summary by default

flub:
This shows a summary of tests which are skipped, failed etc at the end
of the test output.


flub:
Maybe I should make a PR to pytest to make that the default...

PR #233, opened 2020-06-08
feat(minidump): Save minidumps which crash during processing

flub:
This adds a feature which saves minidumps when the process processing
them crashes.  It also improves the metrics reporting these failures
to contain more information about the failure.

The crashed minidumps are stored in the cache directory and have a
default cache timeout of 24h.

Todo:

- [ ] Testing

The way this is jammed into the symbolication "actor" is not ideal,
and as can be seen from `join_procspawn()` this is by far a nice API.
It could be worth splitting this off into a minidump service which
might be able to encapsulate this a bit better.

The other thing which isn't overly nice is how this pretends to be an
entirely normal cache directory, but it only uses one of the fields.
I could be better to create a new config struct for the minidumps
cache instead.  And perhaps it should not be an item in cache::Caches
at all and instead hooked up more manually into the cleanup.

----

#

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -70,6 +70,26 @@ metrics:
   `true`.
 - `connect_to_reserved_ips`: Allow reserved IP addresses for requests to
   sources. See [Security](#security). Defaults to `false`.
+- `caches`: Fine-tune cache expiry.
+   - `downloaded`: Fine-tune caches for downloaded files.
+      - `max_unused_for`: Maximum duration to keep a file since last
+        use of it.
+      - `retry_misses_after`: Duration to wait before re-trying to
+        download a file which was not found.
+      - `retry_malformed_after`: Duration to wait before re-trying to
+        download a file which was malformed.
+    - `derived`: Fine-tune caches for files which are derived from
+      downloaded files.  These files are usually versions of the
+      downloaded files optimised for fast lookups.
+       - `max_unused_for`: Maximum duration to keep a file since last
+         use of it.
+       - `retry_misses_after`: Duration to wait before re-trying to
+         download a file which was not found.
+       - `retry_malformed_after`: Duration to wait before re-trying to
+         download a file which was malformed.
+    - `minidump`: This configures the duration minidumps which failed
+      to be processed correctly will be stored in the cache.
+       - `max_unused_for`: Duration a minidump will be kept in this cache.
 
 ## Security
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -87,9 +87,10 @@ metrics:
          download a file which was not found.
        - `retry_malformed_after`: Duration to wait before re-trying to
          download a file which was malformed.
-    - `minidump`: This configures the duration minidumps which failed
-      to be processed correctly will be stored in the cache.
-       - `max_unused_for`: Duration a minidump will be kept in this cache.
+    - `diagnostics`: This configures the duration diagnostics data
+      will be stored in cache.  E.g. minidumps which failed to be
+      processed correctly will be stored in this cache.
+       - `max_unused_for`: Duration a file will be kept in this cache.
 
 ## Security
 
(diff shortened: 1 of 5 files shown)


docs/index.md
```diff
@@ -70,6 +70,26 @@ metrics:
...
+    - `derived`: Fine-tune caches for files which are derived from
+      downloaded files.  These files are usually versions of the
+      downloaded files optimised for fast lookups.
+       - `max_unused_for`: Maximum duration to keep a file since last
+         use of it.
+       - `retry_misses_after`: Duration to wait before re-trying to
+         download a file which was not found.
+       - `retry_malformed_after`: Duration to wait before re-trying to
+         download a file which was malformed.
+    - `minidump`: This configures the duration minidumps which failed
```
jan-auer, line 90: Should we make this a more generic category like `diagnostics` which comprises all future files that we store for debugging Symbolicator only?

src/actors/symbolication.rs
```diff
@@ -1051,17 +1063,13 @@ impl SymbolicationActor {
...
-                Ok(Err(err)) => Err(err.into()),
-                Err(perr) => Err(SymbolicationError::from(if perr.is_timeout() {
-                    metric!(counter("minidump.modules.spawn.error") += 1, "reason" => "timeout");
-                    SymbolicationErrorKind::Timeout
-                } else {
-                    metric!(counter("minidump.modules.spawn.error") += 1, "reason" => "canceled");
-                    SymbolicationErrorKind::Canceled
-                })),
-            }
+            Self::join_procspawn(
```
jan-auer, line 1066: ❤️

docs/index.md
```diff
@@ -70,6 +70,26 @@ metrics:
   `true`.
 - `connect_to_reserved_ips`: Allow reserved IP addresses for requests to
   sources. See [Security](#security). Defaults to `false`.
+- `caches`: Fine-tune cache expiry.
+   - `downloaded`: Fine-tune caches for downloaded files.
+      - `max_unused_for`: Maximum duration to keep a file since last
```
flub, line 75: It is unclear to me how we can actually set these timeouts.  In the serde structs they are `Option<std::time::Duration>`.  This may need fixing.

src/actors/symbolication.rs
```diff
@@ -1075,6 +1083,75 @@ impl SymbolicationActor {
...
+                    SymbolicationErrorKind::Canceled
+                };
+                metric!(counter(metric) += 1, "reason" => reason);
+                if let SymbolicationErrorKind::Canceled = kind {
+                    Self::save_minidump(minidump, minidump_cache)
+                        .map_err(|e| log::error!("Failed to save minidump {}", LogError(&e)))
+                        .map(|r| match r {
+                            Some(path) => sentry::configure_scope(|scope| {
+                                scope.set_tag("crashed_minidump", path.display());
+                            }),
```
flub, line 1129: I ended up setting this as a tag, but I think I wanted a context?  Only context seems much more restrictive in the things which can be represented this way.  Is there good documentation somewhere that describes what these are for and their tradeoffs?

docs/index.md
jan-auer: The way we do it in Relay is use plain numbers in the config:

https://github.com/getsentry/relay/blob/475e6232f474bfe5d18b894b9ccd87c48fd808a1/relay-config/src/config.rs#L462-L464

And then convert it in a getter function:

https://github.com/getsentry/relay/blob/475e6232f474bfe5d18b894b9ccd87c48fd808a1/relay-config/src/config.rs#L1235-L1237

src/actors/symbolication.rs
jan-auer: Sorry, I missed that. `set_extra` is the way to go.

docs/index.md
flub: As that affects existing code I'll do this in a separate pull request and just leave this as-is here as this is no regression.

PR #234, opened 2020-06-10
ref: Split sources.rs from types.rs

flub:
This creates a new sources.rs file to hold types and impls of them
related to download sources.  It also moves the FileId in the objects
actor to sources::SourceFileId so this is all together.

This is a boring and no-change refactor from #228 split off as a separate PR.


src/actors/objects/mod.rs
```diff
@@ -96,51 +94,6 @@ struct FetchFileMetaRequest {
 #[derive(Clone, Debug)]
 struct FetchFileDataRequest(FetchFileMetaRequest);
 
-#[derive(Debug, Clone)]
-enum FileId {
```
flub, line 100: Move things type (and renaming it) was the main driver for this thing, because it needs to be accessible from the download service without being unreasonable.

PR #235, opened 2020-06-10
maint(ci): Start using github actions for CI

flub:
This is only the basic rust actions, not all things done by travis
currently are covered yet.


flub:
The current travis config only tests on linux, so I did the same here.

flub:
By using the actions-rs/cargo "action" thingy we get the integration which comments in the PR on the right locations for lints etc.  If we want that functionality for the make target we'd have to write our own action I think.  I'd rather be able to use the community maintained actions I think.

I imagined that we'd add another workflow with (some of) the makefile things in it before actually ditching travis.  But my intention here wasn't actually to try and ditch travis, just try out the actions features for a while in parallel to see if it's worth investing in migrating and how we would like migration to look.

.github/workflows/rust.yml
```diff
@@ -41,6 +50,14 @@ jobs:
           toolchain: stable
           override: true
 
+      - name: Cache Rust dependencies download
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-rust-deps
+        with:
+          path: ~/.cargo/registry
+          key: ${{ runner.os }}-download-${{ env.cache-name }}-${{ hashFiles('**/Cargo.lock' }}
```
Swatinem: why is this different from above?

Also, we had mixed success with caching in general. I think having it keyed by `Cargo.lock` is a great idea! 

Not sure about caching `target` though. That was usually a performance regression, especially since it just grows without bounds.

.github/workflows/rust.yml
```diff
@@ -1,4 +1,8 @@
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
```
Swatinem, line 4: TBH, I personally prefer to have all pushes tested, even if I’m not ready yet to create a PR for that. Also, it does provide real value IMO to see if a commit on its own is failing, or the *merge with master* is. The redundant work is unfortunate though.

.github/workflows/rust.yml
jan-auer: Probably having it keyed by `Cargo.lock` is a good compromise. Ideally, I'd prefer the caches to be persistent even across changes, since the bump of a single dependency would otherwise trigger a full rebuild. That said, we can look into [`cargo-sweep`](https://github.com/holmgr/cargo-sweep) to keep the caches small.

jan-auer: That also goes on the assumption that caches are significantly faster than rebuilding. I remember that this is generally the case in symbolicator as long as:
- we make sure that they do not keep growing
- GHA has faster caches than Travis

jan-auer: Even better, we can also disable incremental compilation. See https://github.com/rust-lang/cargo/pull/4817

.github/workflows/rust.yml
flub: Draft PRs do also get tested with this config i think.  Is that not sufficient?

I'm not a huge fan of having all checks run twice and double reporting errors isn't very user friendly.

.github/workflows/rust.yml
flub: Keying on cargo.lock was indeed a compromise between caches growing large without doing too much or relying on smart stuff that could go wrong.  It's also why I decided to key on rustc.  Currently a full build is about 6mins on CI with the cached version 3mins (very rough approximations).

`cargo-sweep` looks interesting, but for the compilation times we have currently I'm tempted to say the complexity isn't worth it.  We could always add it later.

flub: > Even better, we can also disable incremental compilation. See [rust-lang/cargo#4817](https://github.com/rust-lang/cargo/pull/4817)

I guess this speeds up compilation?  Could try it out.

flub:
Deciding how to integrate the makefile functionality is more interesting.  I'm thinking adding another workflow which invokes the makefile e2e tests is straight forward, though that already duplicates the rust tests.  The overlapping linting etc is more interesting.

flub:
It would be nice if there was a filter you could apply to rust's and pytest's output to turn it into action-compatible errors and lints so we could run everything from the makefile.  Maybe that's too crazy.

.github/workflows/rust.yml
jan-auer: It speeds up repeated compilation when you only change single files in a crate. For CI, I think it matters less, and at least it would get rid of most of the bloat in `target`.

.github/workflows/rust.yml
Swatinem: hm, I think I can agree with you. Rather to just open a draft PR than to have "hidden" branches only for testing stuff. I think I can change this for native as well, or also just migrate over to github actions while at it.

flub:
This now does everything that the travis run does, but faster.  The clippy and cargofmt runs are not using the makefile so we get annotations from actions-rs.  The "Rust Check" stop is entirely new and gives faster build feedback with annotations.

Bot unittest and integration tests have to do the same cargo build which should take about the same time.  However because both test suites take >30s it's worth to duplicate this.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,172 @@
...
+      - name: Install stable toolchain
+        id: toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - name: Cache downloads and build
+        uses: actions/cache@v2
```
Swatinem, line 26: lets just keep an eye on this. we had some issues with unbounded cache growth, and the caches are super slow. current full-build jobs take in the order of >40s, not sure how long the clean build times are in comparison.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,172 @@
...
+
+      - name: Install stable toolchain
+        id: toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - name: Cache downloads and build
```
Swatinem, line 55: you might try using yaml aliases for these repeated blocks…

flub: sadly the cache keys differ each time, which afaik means you can't use aliases?

Swatinem: oh right, I missed that

.github/workflows/ci.yml
flub: Caches are deleted when not used for a week.  Or if more than 5GiB of caches exist the oldest get deleted.

Cache times will always depend on the individual project and this will change as the project evolves.  In this case the caches help a lot with build time.  They're pretty conservative wrt to reuse how i've done them here.  This could possibly be done better to even re-use caches when deps change but then also clear out the unused items in the cache.  The current version is simple and won't grow unnecessarily large, which seems like a good starting point.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,192 @@
...
+          yarn global add @zeus-ci/cli
+          echo "::add-path::$(yarn global bin)"
+
+      - name: Upload docs
+        env:
+          ZEUS_API_TOKEN: ${{ secrets.ZEUS_API_TOKEN }}
+          ZEUS_HOOK_BASE: ${{ secrets.ZEUS_HOOK_BASE }}
+        run: |
+          cd site && zip -r gh-pages .
+          zeus upload -t "application/zip+docs" site/gh-pages.zip
```
flub, line 192: I have no idea on this section, nor a clue what the zeus tokens etc are supposed to be.  This is vaguely based on what is done in the makefile.  relay doesn't seem to have a similar section.

@jan-auer what do you know what we want here?  and how to test it without creating releases?

.github/workflows/ci.yml
```diff
@@ -28,14 +30,16 @@ jobs:
...
+      - name: Install rust stable toolchain
+        uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
           profile: minimal
           components: rustfmt, clippy
           override: true
 
-      - uses: Swatinem/rust-cache@v1
+      - name: Chache rust cargo artifacts
```
jan-auer, line 41:
```suggestion
      - name: Cache rust cargo artifacts
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,192 @@
...
+
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ${{ github.workspace }}/node_modules
+            ${{ github.workspace }}/packages/**/node_modules
+            ${{ github.workspace }}/packages/**/build
+            ${{ github.workspace }}/packages/**/dist
+            ${{ github.workspace }}/packages/**/esm
+          key: ${{ runner.os }}-${{ github.sha }}
```
jan-auer, line 179: Please remove caches for published artifacts to ensure we always have a clean build. Since the key changes with every sha, there's also little benefit in caching this, since building the docs is not the bottleneck here.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,192 @@
...
+      - name: Setup Virtualenv
+        run: pip install virtualenv
+
+      - name: Integration tests
+        env:
+          SENTRY_SYMBOLICATOR_GCS_CLIENT_EMAIL: ${{ secrets.SENTRY_SYMBOLICATOR_GCS_CLIENT_EMAIL }}
+          SENTRY_SYMBOLICATOR_GCS_PRIVATE_KEY: ${{ secrets.SENTRY_SYMBOLICATOR_GCS_PRIVATE_KEY }}
+          SENTRY_SYMBOLICATOR_TEST_AWS_ACCESS_KEY_ID: ${{ secrets.SENTRY_SYMBOLICATOR_TEST_AWS_ACCESS_KEY_ID }}
+          SENTRY_SYMBOLICATOR_TEST_AWS_SECRET_ACCESS_KEY: ${{ secrets.SENTRY_SYMBOLICATOR_TEST_AWS_SECRET_ACCESS_KEY }}
+        run: make test-integration CI_ARGS="--ci"
```
jan-auer, line 141: In Relay, we've opted to put the actual pytest invocation here. Can we employ the same pattern here?

jan-auer (changes requested):
Looks like release builds are still missing. Can we create a separate `build` workflow like in Relay where all the release builds are located -- essentially moving over the docs build?

.github/workflows/build.yml
```diff
@@ -0,0 +1,41 @@
...
+  push:
+    branches:
+      - release/**
+
+  # DO NOT MERGE WITH THIS IN IT
+  pull_request:
+
+jobs:
+  linux:
+    name: Linux
```
Swatinem, line 10: maybe name this `Docs` or something appropriate

.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
...
+    name: Lints
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
```
Swatinem, line 22: do we want this explicitly? python 3.8 is the default

.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
...
+
+      - name: Install python dependencies
+        run: pip install --upgrade black flake8
+
+      - name: Run Black
+        run: black --check tests
```
Swatinem, line 28: Q: should we use this directly instead of makefile for symbolic as well?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
...
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Setup Virtualenv
+        run: pip install virtualenv
+
+      - name: Build Docs
+        run: make docs
```
Swatinem, line 137: this was moved to the other job, right?

.github/workflows/ci.yml
flub: Yes, we want to explicitly control which version of python we use IMO, even if it happens to match the system default one currently

.github/workflows/ci.yml
flub: We only build the docs here, to make sure they still build without any error on each PR.  The other job builds and uploads the docs.

Swatinem: Ah sure that makes sense

tests/integration/conftest.py
```diff
@@ -271,9 +281,13 @@ def fail_on_errors():
 
 
 @pytest.fixture
-def s3():
+def s3(pytestconfig):
     if not AWS_ACCESS_KEY_ID or not AWS_SECRET_ACCESS_KEY:
-        pytest.skip("No AWS credentials")
+        msg = "No AWS credentials"
+        if pytestconfig.getoption("ci"):
+            pytest.fail(msg)
```
untitaker, line 288: In theory this kind of stuff could be monitored by measuring coverage on the testsuite itself. Then you can find tests that are improperly skipped or not discovered at all for whatever reason.

In practice it's much harder to set up.

Swatinem (approved):

tests/integration/conftest.py
flub: I think that could work if you have a 100% test coverage enforced.  But anything less and it probably doesn't.  I like one global that you can use to potentially skip things on local workstations.

untitaker: I mean testing coverage of the code in `tests/`, not the code that is being tested. Surely all tests are running in CI?

.github/workflows/ci.yml
jan-auer: Since we're going with explicit commands now, I would say yes. The makefile becomes an interface for the developer, and CI has it's own (equivalent or equal) commands.

jan-auer (changes requested):
`craft.yml` still requires the following artifacts:

  - `symbolicator-Linux-x86_64`
  - `symbolicator-Linux-x86_64-debug.zip`

Unfortunately, merging like this will break releases. The artifacts were built by the `release: linux x86_64` Travis job. You can either remove the craft requirement, or add the release build back.

flub:
> `craft.yml` still requires the following artifacts:
> 
> * `symbolicator-Linux-x86_64`
> * `symbolicator-Linux-x86_64-debug.zip`
> 
> Unfortunately, merging like this will break releases. The artifacts were built by the `release: linux x86_64` Travis job. You can either remove the craft requirement, or add the release build back.

Deleted these two from `.craft.yml` as I think things were broken since they were added.

Producing binaries opens up so many questions for me that I really would rather not mix that up into this PR.  Apologies.

jan-auer (approved):
> Deleted these two from .craft.yml as I think things were broken since they were added.

I'm not sure what makes you say that. Both the build and the craft configuration look fine to me. I remember testing this when I added it. But it's up to you, we can bring this back later.

Swatinem:
Note to self:
```
    Updating git repository `https://github.com/Keats/jsonwebtoken`
    Updating git repository `https://github.com/getsentry/symbolic`
    Updating git submodule `https://github.com/getsentry/breakpad`
    Updating git submodule `https://github.com/getsentry/linux-syscall-support`
    Updating git repository `https://github.com/Swatinem/cpp_demangle`
    Updating git repository `https://github.com/Swatinem/msvc-demangler-rust`
   Compiling symbolicator v0.2.0 (/home/runner/work/symbolicator/symbolicator)
   Compiling cpp_demangle v0.3.0 (https://github.com/Swatinem/cpp_demangle?branch=sentry-patches#2d274c48)
   Compiling symbolic-demangle v7.5.0 (https://github.com/getsentry/symbolic?branch=fix/demangle-fixes#737969f7)
   Compiling symbolic v7.5.0 (https://github.com/getsentry/symbolic?branch=fix/demangle-fixes#737969f7)
```
My cache does not support git dependencies yet, but it looks like it would be a small win here

flub:
> > Deleted these two from .craft.yml as I think things were broken since they were added.
> 
> I'm not sure what makes you say that. Both the build and the craft configuration look fine to me. I remember testing this when I added it. But it's up to you, we can bring this back later.

Ok, I read the existing travis.ymy and makefile badly again.  Re-added this now.  Ready to merge again I think.

I assume once this is merged with admin rights the requirement for the travis checks will disappear again?

jan-auer:
Updated branch protection rules on GitHub.

PR #236, opened 2020-06-10
feat(config): Allow configuring cache durations

flub:
This adds humantime-serde to allow deserialising std::time::Duration
from reasonable strings.

It also does some horrible stuff to merge the defaults into these
fields.  I'm not sure how to make this more reasonable.

The "correct" thing to do would be to turn `CacheConfig` into a trait and have 3 specific types for each of the `downloaded`, `derived` and `diagnostics`.  Please let me know if you have opinions on whether that would be any less horrible then this.


src/app.rs
```diff
@@ -44,7 +44,7 @@ pub struct ServiceState {
 impl ServiceState {
     pub fn create(config: Config) -> Result<Self, ServiceStateError> {
         let config = Arc::new(config);
-
+        println!("{:#?}", config);
```
Swatinem, line 47:
```suggestion
```

;-)

flub:
If the general changes to the file are fine could be worth considering whether a trait for `ConfigCache` would be better than the current enum.

flub:
I'd like to do one more change: rename `caches.diagnostics.duration` name into `caches.diagnostics.retention` in the config file if there are no objections?

jan-auer:
@flub Do it, that name makes more sense.

PR #237, opened 2020-06-17
ref: Make some fields in sources private

flub:
This reduces some knowledge the callers have on this as well.


flub:
> Have you considered also implementing `AsRef<str>`, add a `as_str()`, and potentially `Deref`? 

I thought about `Deref`, but I rather try and see if it works fine as an opaque type.  And so far it seems like it does.  Same as for `AsRef<str>` I guess, it gives an escape hatch to use things as a string instead of the newtype.

The deref docs even say in bold **`Deref` should only be implemented for smart pointers**, and newtypes are not smart pointers.  I tend to agree with that so far I think.

PR #238, opened 2020-06-18
fix: Create cache directories at startup

flub:
The diagnostics cache directory was not created.  It seems all the
cache directories are created on the fly at runtime.  This changes
this to create all the cache directories at startup, ensuring the
application fails to start when it can not create one of the caches.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -107,17 +107,20 @@ pub struct Cache {
 }
 
 impl Cache {
-    pub fn new<P: AsRef<Path>>(
+    pub fn new(
         name: &'static str,
-        cache_dir: Option<P>,
+        cache_dir: Option<PathBuf>,
         cache_config: CacheConfig,
-    ) -> Self {
-        Cache {
+    ) -> io::Result<Self> {
+        if let Some(ref dir) = cache_dir {
+            std::fs::create_dir_all(dir)?;
+        }
+        Ok(Cache {
             name,
-            cache_dir: cache_dir.map(|x| x.as_ref().to_owned()),
+            cache_dir,
             start_time: SystemTime::now(),
             cache_config,
-        }
+        })
     }
 
     pub fn name(&self) -> &'static str {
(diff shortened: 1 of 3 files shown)


flub:
Good point, I think doing both up-front and at access time is best.

src/cache.rs
```diff
@@ -107,17 +107,20 @@ pub struct Cache {
...
 impl Cache {
-    pub fn new<P: AsRef<Path>>(
+    pub fn new(
         name: &'static str,
-        cache_dir: Option<P>,
+        cache_dir: Option<PathBuf>,
         cache_config: CacheConfig,
-    ) -> Self {
-        Cache {
+    ) -> io::Result<Self> {
```
Swatinem, line 114: Just wondering what people think of this. I know there is even a clippy lint enforcing that `new` only returns `Self`.

jan-auer: Right, this should be renamed to `create` or something similar. The lint was backed out at some point, but it's still a good habit.

flub: Oh, interesting. I've never spotted any consistency in this.  Though glancing through the stdlib the only example I can find of `::new()` returning `Result` is `CString` right now.  (I know that stdlibs are rarely an example of how to do things the right way after a few years...).

> `create` or something similar.

"or something similar"?  Would be good to have a reasonable convention here if we don't like new returning result.

> The lint was backed out at some point, but it's still a good habit.

Let's re-add it if we choose to prefer this.

flub: https://rust-lang.github.io/rust-clippy/master/index.html#new_ret_no_self

flub: That does seem to allow `Result<Self>`

untitaker: I would not expect `new` to return result. `CString` is an outlier to me.

flub: Now with `s/new/from_config/`

src/cache.rs
```diff
@@ -318,29 +321,29 @@ pub struct Caches {
 }
 
 impl Caches {
-    pub fn new(config: &Config) -> Self {
-        Self {
+    pub fn new(config: &Config) -> io::Result<Self> {
```
Swatinem, line 324: same applies here as well

flub: good spot, apologies!  `s/new/from_config/`

PR #239, opened 2020-06-22
ref: Remote thread timeout

flub:
The `future_metrics!` macro is `futures v0.1` only and hinders adoption of `futures v0.3`. What it does should be part of the services interface so integrate it into the `RemoteThread::spawn` call.

### DIFF
--- a/src/services/download/mod.rs
+++ b/src/services/download/mod.rs
@@ -85,35 +85,40 @@ impl DownloadService {
         source: SourceFileId,
         destination: PathBuf,
     ) -> Box<dyn Future<Item = DownloadStatus, Error = DownloadError> + Send + 'static> {
-        let fut03 = self.worker.spawn(|| async move {
-            match source {
-                SourceFileId::Sentry(source, loc) => {
-                    sentry::download_source(source, loc, destination)
-                        .compat()
-                        .await
+        let fut03 = self.worker.spawn(
+            "service.download",
+            std::time::Duration::from_secs(3600),
+            || async move {
+                match source {
+                    SourceFileId::Sentry(source, loc) => {
+                        sentry::download_source(source, loc, destination)
+                            .compat()
+                            .await
+                    }
+                    SourceFileId::Http(source, loc) => {
+                        http::download_source(source, loc, destination)
+                            .compat()
+                            .await
+                    }
+                    SourceFileId::S3(source, loc) => {
+                        s3::download_source(source, loc, destination).compat().await
+                    }
+                    SourceFileId::Gcs(source, loc) => {
+                        gcs::download_source(source, loc, destination)
+                            .compat()
+                            .await
+                    }
+                    SourceFileId::Filesystem(source, loc) => {
+                        filesystem::download_source(source, loc, destination)
+                    }
                 }
-                SourceFileId::Http(source, loc) => {
-                    http::download_source(source, loc, destination)
-                        .compat()
-                        .await
-                }
-                SourceFileId::S3(source, loc) => {
-                    s3::download_source(source, loc, destination).compat().await
-                }
-                SourceFileId::Gcs(source, loc) => {
-                    gcs::download_source(source, loc, destination)
-                        .compat()
-                        .await
-                }
-                SourceFileId::Filesystem(source, loc) => {
-                    filesystem::download_source(source, loc, destination)
-                }
-            }
-        });
+            },
+        );
         let fut01 = fut03
             .map(|spawn_ret| spawn_ret.unwrap_or_else(|_| Err(DownloadErrorKind::Canceled.into())))
             .boxed()
             .compat();
+        // TODO: remote future_metrics! macro
         Box::new(future_metrics!(
             "service.download",
             Some((
--- a/src/services/download/mod.rs
+++ b/src/services/download/mod.rs
@@ -114,19 +114,13 @@ impl DownloadService {
                 }
             },
         );
+
+        // Map all SpawnError variants into DownloadErrorKind::Canceled.
         let fut01 = fut03
             .map(|spawn_ret| spawn_ret.unwrap_or_else(|_| Err(DownloadErrorKind::Canceled.into())))
             .boxed()
             .compat();
-        // TODO: remote future_metrics! macro
-        Box::new(future_metrics!(
-            "service.download",
-            Some((
-                std::time::Duration::from_secs(3600),
-                DownloadErrorKind::Canceled.into()
-            )),
-            fut01
-        ))
+        Box::new(fut01)
     }
 }
 
--- a/src/services/download/mod.rs
+++ b/src/services/download/mod.rs
@@ -188,7 +188,7 @@ mod tests {
         };
 
         let svc = DownloadService::new(RemoteThread::new_threaded());
-        let ret = test::block_fn(|| svc.download(source_id, dest.clone()));
+        let ret = test::block_fn01(|| svc.download(source_id, dest.clone()));
         assert_eq!(ret.unwrap(), DownloadStatus::Completed);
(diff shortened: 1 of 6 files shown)


src/utils/futures.rs
```diff
@@ -119,21 +145,68 @@ impl RemoteThread {
...
-    /// execution.  The output type of the spawned future is wrapped in a
-    /// `Result`, normally the output is returned in an `Ok` but when the remote
-    /// future is dropped because the thread is shut down or restarted for any
-    /// reason it returns `Err(oneshot::Canceled)`.
-    pub fn spawn<F, R, T>(&self, factory: F) -> impl Future<Output = Result<T, RemoteCanceled>>
+    /// The returned future resolves when the spawned future has completed execution.  The
+    /// output type of the spawned future is wrapped in a `Result`, normally the output is
+    /// returned in an `Ok`.
+    ///
+    /// Then the future takes too long `SpawnError::Timeout` is returned, ifthe
```
Swatinem, line 152:
```suggestion
    /// When the future takes too long `SpawnError::Timeout` is returned, if the
```

src/utils/futures.rs
```diff
@@ -119,21 +145,68 @@ impl RemoteThread {
...
+                    .compat()
+                    .timeout(timeout)
+                    .then(move |r: Result<T, tokio::timer::timeout::Error<()>>| {
+                        match r {
+                            // TODO: Should send failures be logged?  Panic?
+                            Ok(o) => {
+                                metric!(
+                                    timer("futures.done") = start_time.elapsed(),
+                                    "task_name" => task_name,
+                                    "status" => "done",
```
Swatinem, line 190: maybe you can move the duplicate code above, and just use `"status": if r.is_ok() { "done" } else { "timeout" }`

src/utils/futures.rs
```diff
@@ -172,3 +249,79 @@ impl Drop for CallOnDrop {
...
+    use crate::test;
+
+    #[test]
+    fn test_remote_thread() {
+        // Ensure actix is setup.
+        test::setup();
+
+        // test::setup() enables logging, but this test spawns a thread where
+        // logging is not captured.  For normal test runs we don't want to
+        // pollute the stdout so silence logs here.  When debugging this test
```
Swatinem, line 266: tests suppress stdout by default, so I‘m not sure you need to silence them manually?

src/utils/futures.rs
```diff
@@ -172,3 +249,79 @@ impl Drop for CallOnDrop {
...
+            let (tx, rx) = oneshot::channel();
+            std::thread::spawn(move || {
+                std::thread::sleep(Duration::from_secs(10));
+                tx.send(42).ok();
+            });
+            rx
+        });
+        std::mem::drop(remote);
+        let ret = test::block_on(fut);
+        println!("{:?}", ret);
```
Swatinem, line 324:
```suggestion
```

although this is not captured by default…

src/utils/futures.rs
flub: Because we spawn a new thread and in new threads this is not done, and the functionality to do this isn't even exposed as a public API.  So we end up with this sad thing.

src/utils/futures.rs
flub: good call, done

src/utils/futures.rs
```diff
@@ -119,21 +145,62 @@ impl RemoteThread {
...
+                            Ok(o) => ChannelMsg::Output(o),
+                            Err(_) => {
+                                // Because we wrapped our T into future::ok() above to make a
+                                // TryFuture, we know the <Timeout as Future>::Error will only
+                                // occur if the error is actually because of a timeout, so we do
+                                // not need to check with .is_timer() or .is_inner().
+                                ChannelMsg::Timeout
+                            }
+                        };
+                        // TODO: Should send failures be logged?  Panic?
```
flub, line 199: So this is still an interesting question.  This would happen if the future which called `.spawn()` here was dropped itself and thus the receiver is no longer there.  This could realistically happen when e.g. that future itself times out and is cancelled.  So I think panic is likely the wrong thing to do.

I guess that leaves logging this?  Probably just as info, I doubt this warrants and error since it is not the cause but rather an effect.

flub: I went with info log

src/utils/futures.rs
```diff
@@ -96,6 +99,29 @@ pub struct RemoteThread {
...
+impl Drop for RemoteThread {
+    fn drop(&mut self) {
+        // Without sending the StopArbiter message the arbiter thread keeps running even if
+        // you drop the arbiter.
+        if let Some(ref addr) = self.arbiter {
+            addr.do_send(actix::msgs::StopArbiter(0));
+        }
+    }
+}
+
```
flub, line 124: This is wrong, because the `RemoteThread` can be cloned which only clones the address of the arbiter dropping a clone would terminate the arbiter while there was still another one.  Need to wrap the arbiter into an inner `Arc` to make this work.

flub:
@Swatinem PTAL, one new commit

src/services/download/mod.rs
```diff
@@ -85,43 +85,42 @@ impl DownloadService {
...
-                SourceFileId::Gcs(source, loc) => {
-                    gcs::download_source(source, loc, destination)
-                        .compat()
-                        .await
-                }
-                SourceFileId::Filesystem(source, loc) => {
-                    filesystem::download_source(source, loc, destination)
+        let fut03 = self.worker.spawn(
+            "service.download",
+            std::time::Duration::from_secs(3600),
```
Swatinem, line 90: maybe in the future we want to make this configurable

flub: Yeah, I flagged this on the PR that introduced this as fairly arbitrary and probably poorly chosen.

I agree that it should be more controlled by the caller possibly, the parent-most request, the http endpoint invoking this all, should have an overall deadline and all other things that end up being called internally do not need to do any work past this deadline.  In theory cancelling the future is fine, but this breaks across spawn boundaries, and thus service boundaries.  This does indeed mean that the caller needs to supply the deadline at spawn time.

I think it's for another refactor though, the current codebase has various hardcoded timeouts that mostly try to align somewhat with each other.

flub:
This is essentially a re-implementation of the same as the previous commit, only with a bit more async-await.

Major addition is a counter for timeouts.

@jan-auer I think this is worse and we should stick with the previous commit (or do a few small improvements on top of it like counting cancellations).

* The `futures.wait_time` and `futures.done` metrics are intrinsically bound to task spawning because we are bound to spawning functions/closures.  This would be a different story if we were spawning only futures.

* We are still dealing with an old runtime which makes us having to jump through all these compat layers every time we compose.  (Demonstrated here with the `.compat().timeout().compat()` joy)

So my proposal is to stay with the previous commit, maybe use a bit more `.unit_error()` and add the cancelation counter.  Then later introduce a handle that when dropped, drops the executing task which allows us to remove the timeout and move that functionality to the caller instead.

src/services/download/mod.rs
```diff
@@ -85,43 +85,42 @@ impl DownloadService {
...
             .compat();
-        Box::new(future_metrics!(
-            "service.download",
-            Some((
-                std::time::Duration::from_secs(3600),
-                DownloadErrorKind::Canceled.into()
-            )),
-            fut01
-        ))
+        Box::new(fut01)
```
Swatinem, line 123: you have `Box::new(.boxed().compat())`, can you reorder this in order to avoid the double boxing?

flub: IIRC no.  The `TryFuture` trait (which is where this `.compat()` comes from) is implemented for `std::future::Future` but the `.map()` returns a `futures::future::Map` and we need to box to make it a `TryFuture`.

Note that this goes away in #241

PR #240, opened 2020-06-22
style: Deny clippy by default using code attributes

flub:
This replaces the -D option on the command line, so running lints are
more consistent.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ lint-python: .venv/bin/python
 
 lint-rust:
 	@rustup component add clippy --toolchain stable 2> /dev/null
-	cargo +stable clippy --all-features --all --tests --examples -- -D clippy::all
+	cargo +stable clippy --all-features --all --tests --examples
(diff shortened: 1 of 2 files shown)


flub:
You run clippy before getting the rest of the PR ready?  my workflow is build->test->clippy->submit pr
I copied this from another project and it's never got in my way.  The difference isn't big though, I don't mind dropping this PR, just curious why you think this is such a blocker.

flub:
Ping, what shall we do with this?  Merge or close?

flub:
> Ping, what shall we do with this? Merge or close?

whoops, github was not auto-updating again... just notice there are new comments :)

flub:
> Would adding `RUSTFLAGS="-D warnings"` to the CI builds work, instead? You would still see warnings when you compile locally, but at least it's impossible to miss them when they occur on CI.

That's already the current behaviour, locally warnings, CI errors.  Let's leave it as is.

PR #241, opened 2020-06-23
ref: Make the service level API use new futures

flub:
Switch to new-style futures for the service API boundary.  This seems surprisingly not invasive for the callers.

Depends on #239


src/services/download/mod.rs
```diff
@@ -84,43 +85,46 @@ impl DownloadService {
...
-        let fut03 = self.worker.spawn(
-            "service.download",
-            std::time::Duration::from_secs(3600),
-            || async move {
-                match source {
-                    SourceFileId::Sentry(source, loc) => {
-                        sentry::download_source(source, loc, destination)
-                            .compat()
-                            .await
+    ) -> Pin<
```
Swatinem, line 88: do you really need the explicit `Pin` here?

Also, since this type spans 7 lines, how about creating a type alias for it?

flub: Honestly, I don't know this yet as I'm not overly familiar/experienced yet with the `async fn` and boxed futures interaction.  I used it because this is what `.boxed()` returns, so I assume it makes turning things into `async fn` easier later on.  Does it do any harm?

I'm kind of -1 for a type alias here.  It's still pretty easy to read as it still clearly shows one future being returned and it directly shows what's up.  Type aliases are not show in compiler errors anyway afaik so it doesn't help there.  And lastly, this is just one location - having to look up a what a new name means for just one location is more overhead then I like.

Swatinem: Can you correctly `await` the boxed future even if its not explicitly wrapped in `Pin`? I think you shouldn’t manually return pin here

flub: Because pins are scary magic (aka I have no clue what they do beyond the simplistic theoretical explanation) I tried this out.  But `TryFutureExt::compat()` (which is used above in objects/mod.rs:201 to turn the future03 returned from the service into a future01) requires that self is Unpin.  Which leads us to having to wrap the future here into pin.  Which is presumably why `FutureExt::boxed()` does this pinning in the first place.

`std::future::Future` has a blanket implementation for `Pin`, which is I think why you can await on it.  I've updated the test to use an `async move` block which `.await`s the service to prove this works (and keeps working).

Swatinem: Does `.compat()` work with standard `async fn`?

Instead of declaring the future explicitly, maybe just make this an `async fn(…) -> Result<DownloadStatus, DownloadError>`? And do a `.await` instead of the `.boxed()` inside? Would that work?

I’m under the impression that things *have to be* simpler than this. And maybe we are doing something wrong?

flub: `async fn` doesn't give you `Pin` either so `TryFuture.compat()` is broken again.  In general we have to start doing async fn from the outside in, rather than inside out.  There are some boundaries in which we can isolate things and start using `async fn`, this is one of those boundaries we're trying to build.

flub: Mind you, I think we could get this to work with async fn, but the caller would have to do something like `async move { xxx.download(...).await }.boxed().compat()` but I didn't even get this entirely to work.  I think it's possible to make it work (something with our tryf! macro implementation) but I don't think this is worth the time trying to figure this out because the boundary is now horrible.  The existing API allows us interop.

We "just" need to get on with it and keep migrating and this will go away ;)

Swatinem: > doing async fn from the outside in, rather than inside out.

Too bad :-( Is this documented somewhere? I can’t imagine we are the first ones to hit these problems. Otherwise this might be worth a short blog post, maybe if we are a bit further along with this.

jan-auer: Updated to `impl Future`. The solution would have been to return `Box<impl Future + Unpin>`.

> Is this documented somewhere? 

It is a known limitation that `async fn` captures the lifetimes of all its parameters. That said, this is actually something we want. For this reason, we also have to do this:

```
test::block_fn(move || async move { service.download(source_id, dest2).await });
```

If we didn't do this, the service would drop before the future has completed, which drops the remote thread prematurely. Long story short, we just have to swallow the pill until the entire stack is migrated.

PR #243, opened 2020-06-25
fix(download): Pass a sentry hub to the download service

flub:
This allows passing a sentry hub to the download service to use.  It experiments with the API for this:

* Passing the hub to the service is not download specific but rather depends on the service.  This attempts to not mix the arguments, instead use something which resembles the builder pattern, but isn't quite that.
* ~I can imagine the timeout could become used like this as well.~
* The downside is that adding the hub (or future timeout) is optional, nothing forces you to do this.  It is kind of nice to be able to force callers to do this.
* Making a proper builder pattern would avoid this, but every service being a builder is a bit wieldy to use: `svc.bind_hub(Hub::current()).with_timeout(Duration::from_secs(10)).build().download(source, dest)` reads that bit weirder.
* Possibly should be called `.with_hub()` instead of `.bind_hub()`

This relies on #241


src/utils/sentry.rs
```diff
@@ -22,21 +23,68 @@ where
...
+    where
+        H: Into<Arc<Hub>>,
+    {
+        SentryFuture {
+            inner: self,
+            hub: hub.into(),
+        }
+    }
+
+    fn sentry_hub_current(self) -> SentryFuture<Self> {
```
Swatinem, line 54: do you think these make sense to have in sentry itself?

src/utils/sentry.rs
```diff
@@ -5,7 +5,7 @@ use futures01::Poll;
 
 use sentry::{Hub, Scope};
 
-// TODO: where the fuck is the warning about the missing Debug impl???
```
Swatinem: 🤣

src/utils/sentry.rs
flub: These were the ones that @jan-auer wanted to drop, and I think I agree.  A single `.bind_hub()` with explicitly passing in the hub you want is probably more readable.  But good point, there's no point in providing these methods on the new-and-temporary `SentryStdFutureExt`.

src/utils/sentry.rs
flub: i didn't mean for this to get checked in but... genuine question though.  Isn't the compiler supposed to warn about this?

Swatinem: Maybe check the rust repo if there is a report about this? I noticed that `missing_doc_code_examples` is basically broken and I couldn’t even get it to work a single time.

src/utils/sentry.rs
jan-auer: Yeah, let's back those out in a future PR. We might want to add a convenience method for `Hub::new_from_top(Hub::current())` though.

PR #244, opened 2020-06-30
Implement list_files on download service

flub:
This implements a `list_files()` method on the download service and migrates over the calling code.

This does not yet upgrade any code, most is a plain copy with just the error type being adjusted.


src/services/download/mod.rs
```diff
@@ -120,12 +147,29 @@ impl DownloadService {
...
+        filetypes: &'static [FileType],
+        object_id: ObjectId,
+    ) -> impl std::future::Future<Output = Result<Vec<SourceFileId>, DownloadError>> {
+        let hub = Hub::current();
+        self.worker
+            .spawn(
+                "service.download.list_files",
+                std::time::Duration::from_secs(1800),
+                move || dispatch_list_files(source, filetypes, object_id).bind_hub(hub),
+            )
```
flub, line 169: It might be worth to avoid spawning for the ones which are implemented synchronously

src/actors/objects/mod.rs
```diff
@@ -473,18 +461,24 @@ impl ObjectsActor {
...
-                            // search by debug/code id. We do not surface those errors to the user
-                            // (instead we default to an empty search result) and only report them
-                            // internally.
-                            log::error!("Failed to download from {}: {}", type_name, LogError(&e));
-                            Ok(Vec::new())
-                        })
-                        // create a new hub for each prepare_downloads call
-                        .sentry_hub_new_from_current()
+                    let hub = Hub::new_from_top(Hub::current());
+                    Hub::run(Arc::new(hub), || {
```
flub, line 465: Not that I want to change anything here now.  But this still makes me think it might be better to integrate the hub binding into the services API because it is rather deceptive that you can't just bind a hub to the future.  But perhaps there's something even better we can do with asyncfn.

src/services/download/mod.rs
```diff
@@ -84,6 +91,26 @@ async fn dispatch_download(
...
+    source: SourceConfig,
+    filetypes: &'static [FileType],
+    object_id: ObjectId,
+) -> Result<Vec<SourceFileId>, DownloadError> {
+    match source {
+        SourceConfig::Sentry(cfg) => {
+            let vec = sentry::list_files(cfg, filetypes, object_id)
+                .compat()
+                .await?;
+            Ok(vec)
```
jan-auer, line 105: Can you just directly return the expression here?

```rust
        SourceConfig::Sentry(cfg) => {
            sentry::list_files(cfg, filetypes, object_id)
                .compat()
                .await
        }
```

src/services/download/mod.rs
jan-auer: Agreed. Considering that we only have a single variant that needs a spawn, how would you feel about moving the `match` in here and then just spawn in the `SourceConfig::Sentry` arm? We can refactor this again if we choose to add more logic in the future.

src/actors/objects/mod.rs
jan-auer: Agreed that we can just wait for `async fn`. That said, maybe add a comment to clarify why we choose this approach over `.bind_hub`.

src/services/download/mod.rs
flub: I tend to avoid `Ok()`-wrapping around multi-line expressions as I don't find it very clear.  But sure, maybe the ok-wrapping can be on the returned thing instead.

jan-auer: Unless I'm missing something, there's no need for `Ok()` wrapping. See my code example. You already get a `Result<Vec<_>, DownloadError>` from `.await`, which you can directly return.

flub: oh, oops.  yes, you're right

PR #247, opened 2020-07-10
fix: Do not skip modules without debug id

flub:
Some minidumps are missing debug IDs and skipping them results in not
even all modules showing up which results in even more confusing crash
reporting.

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -1055,7 +1055,7 @@ impl SymbolicationActor {
                             (
                                 code_module
                                     .id()
-                                    .unwrap_or_else(|| CodeModuleId::from(DebugId::nil())),
+                                    .unwrap_or_default(),
                                 object_info_from_minidump_module(object_type, code_module),
                             )
                         })
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -1304,7 +1304,7 @@ impl SymbolicationActor {
                                     object_info_from_minidump_module(object_type, code_module)
                                         .into();
 
-                                let module_id: Option<CodeModuleId> = code_module.id();
+                                let module_id = code_module.id();
 
                                 let status = match module_id {
                                     Some(id) => unwind_statuses
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -1299,7 +1299,7 @@ impl SymbolicationActor {
                         let modules = process_state
                             .modules()
                             .into_iter()
-                            .filter_map(|code_module| {
+                            .map(|code_module| {
                                 let mut info: CompleteObjectInfo =
                                     object_info_from_minidump_module(object_type, code_module)
                                         .into();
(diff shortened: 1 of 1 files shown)


flub:
It would be nice to be able to test this.  @jan-auer said he had ideas on how to create a minidump that shows this behaviour.

src/actors/symbolication.rs
```diff
@@ -1051,11 +1051,13 @@ impl SymbolicationActor {
...
                         .referenced_modules()
                         .into_iter()
-                        .filter_map(|code_module| {
-                            Some((
-                                code_module.id()?,
+                        .map(|code_module| {
+                            (
+                                code_module
+                                    .id()
+                                    .unwrap_or_else(|| CodeModuleId::from(DebugId::nil())),
```
jan-auer, line 1058: _nit_: `CodeModuleId` implements `Default`, which could come in handy here:

```suggestion
                                    .unwrap_or_default(),
```

flub:
Now actually behaves like we want it to, confirmed with local tests.  Sadly still not with automated tests.

src/actors/symbolication.rs
```diff
@@ -1304,21 +1304,28 @@ impl SymbolicationActor {
                                     object_info_from_minidump_module(object_type, code_module)
                                         .into();
 
-                                let status = unwind_statuses
-                                    .get(&code_module.id()?)
-                                    .cloned()
-                                    .unwrap_or(ObjectFileStatus::Unused);
+                                let module_id: Option<CodeModuleId> = code_module.id();
```
jan-auer, line 1307: This type annotation is redundant, as this can be inferred statically.

src/actors/symbolication.rs
```diff
@@ -1304,21 +1304,28 @@ impl SymbolicationActor {
...
-                                    .copied()
-                                    .unwrap_or_default();
+                                let features = match module_id {
+                                    Some(id) => {
+                                        object_features.get(&id).copied().unwrap_or_default()
+                                    }
+                                    None => Default::default(),
+                                };
 
                                 info.features.merge(features);
```
jan-auer, line 1331: Just below here, the linter is complaining that this closure always returns `Some`, so it can be converted from `filter_map` to `map`.

flub:
> Thanks, this looks just about right. As discussed offline, in a follow-up we can either introduce a new error status, or have the UI detect the missing ID and highlight it accordingly.

The UI already shows that the debug id's are missing.  But it isn't highlighted

src/actors/symbolication.rs
flub: Yes indeed.  I added it because I find it improves readability.  I like being able to read code without needing rust-analyzer...  But the code matches on it right away so maybe it's obvious enough if you don't care about what type the id itself actually is.

src/actors/symbolication.rs
flub: yep, forgot to run clippy locally...

PR #248, opened 2020-07-10
ref: Convert the Gcs downloader to async/await

Swatinem:
There is a lot of duplicated code with the Http Downloader, I’m tempted to extract that further.


src/services/download/gcs.rs
```diff
@@ -10,16 +10,18 @@ use std::sync::Arc;
 use actix_web::{client, HttpMessage};
 use chrono::{DateTime, Duration, Utc};
 use failure::{Fail, ResultExt};
-use futures01::prelude::*;
+use futures::compat::Future01CompatExt;
+use futures::compat::Stream01CompatExt;
```
flub, line 14: l: do we prefer to keep these two imports on one line?

src/services/download/gcs.rs
```diff
@@ -109,117 +105,106 @@ fn request_new_token(
...
+                Ok(DownloadStatus::NotFound)
+            }
+        }
+        Err(e) => {
+            log::trace!(
+                "Skipping response from GCS {} (from {}): {} ({:?})",
+                &key,
+                source.bucket,
+                &e,
+                &e
```
jan-auer, line 203: Is there a reason we format the same error twice? Looks like once should be enough. Also, no need to `&` all these, as the macro will take references only.

PR #249, opened 2020-07-10
ref: Deduplicate Http download logic

Swatinem:
This deduplicates shared code between Http and Gcs downloaders into
a common utility that can download any Url.

Some thoughts:
- Instead of `Url`, maybe refactor to use `&str` instead, avoiding some clones, etc
- I would have liked to use `impl Iterator<(&str, &str)>` for the headers, but they need to be `Send` (and probably `Clone` as well?), so creating an intermediate `BTreeMap` right now. I think its possible to avoid that as well maybe.

### DIFF
--- a/src/services/download/gcs.rs
+++ b/src/services/download/gcs.rs
@@ -4,6 +4,7 @@
 //!
 //! [`GcsSourceConfig`]: ../../../sources/struct.GcsSourceConfig.html
 
+use std::collections::BTreeMap;
 use std::path::PathBuf;
 use std::sync::Arc;
 
@@ -11,17 +12,14 @@ use actix_web::{client, HttpMessage};
 use chrono::{DateTime, Duration, Utc};
 use failure::{Fail, ResultExt};
 use futures::compat::Future01CompatExt;
-use futures::compat::Stream01CompatExt;
-use futures::prelude::*;
 use parking_lot::Mutex;
 use serde::{Deserialize, Serialize};
-use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use url::percent_encoding::{percent_encode, PATH_SEGMENT_ENCODE_SET};
+use url::Url;
 
 use super::{DownloadError, DownloadErrorKind, DownloadStatus};
 use crate::sources::{FileType, GcsSourceConfig, GcsSourceKey, SourceFileId, SourceLocation};
 use crate::types::ObjectId;
-use crate::utils::futures::delay;
 
 lazy_static::lazy_static! {
     static ref GCS_TOKENS: Mutex<lru::LruCache<Arc<GcsSourceKey>, Arc<GcsToken>>> =
@@ -137,15 +135,6 @@ async fn get_token(source_key: &Arc<GcsSourceKey>) -> Result<Arc<GcsToken>, Down
     Ok(token)
 }
 
-async fn start_request(
-    url: &str,
-    token: &GcsToken,
-) -> Result<client::ClientResponse, client::SendRequestError> {
-    let mut builder = client::get(&url);
-    builder.header("authorization", format!("Bearer {}", token.access_token));
-    builder.finish().unwrap().send().compat().await
-}
-
 pub async fn download_source(
     source: Arc<GcsSourceConfig>,
     download_path: SourceLocation,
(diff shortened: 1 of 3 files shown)


src/services/download/mod.rs
```diff
@@ -202,7 +202,7 @@ fn download_future_stream(
 ///
 /// This is common functionality used by all many downloaders.
 async fn download_stream(
-    source: SourceFileId,
+    source: &str,
```
flub, line 205: Are we passing this in purely for the trace log?  Do we even have trace log enabled?

But if we're keeping this, why can't we pass `&SourceFileId`?

src/services/download/http.rs
```diff
@@ -70,22 +71,20 @@ async fn start_request(
...
-pub async fn download_source(
-    source: Arc<HttpSourceConfig>,
-    download_path: SourceLocation,
+/// Downloads the given `url` to `destination`.
+///
+/// This will internally follow HTTP redirects, adding the given `headers` to each intermediate
+/// request. It will further apply an exponential backoff in case of download errors.
+pub async fn download_url(
+    url: Url,
+    headers: &BTreeMap<String, String>,
```
flub, line 80: Why `BTreeMap` instead of `HashMap`, do you care about order?

src/services/download/gcs.rs
```diff
@@ -156,56 +145,20 @@ pub async fn download_source(
...
-                "Skipping response from GCS {} (from {}): {} ({:?})",
-                &key,
-                source.bucket,
-                &e,
-                &e
-            );
-            Ok(DownloadStatus::NotFound)
-        }
-    }
+    super::http::download_url(url, &headers, destination).await
```
flub, line 161: This didn't used to follow redirects, now it does.  I'm not sure how much we care, but it is a change.

flub:
> * Instead of `Url`, maybe refactor to use `&str` instead, avoiding some clones, etc

I like specific types, `&str` is like `void*`...  But the whole reason for this cloning is inside `follow_redirects()`. We could push the cloning down into there which is already a little bit better.  Before I did an attempt at re-writing follow_redirects to only do the clone when a redirect needed to be followed, but ended up with something less readable.  Though maybe we could have another attempt at this.

The other thing is that `download_url()` could take a boolean (or custom enum) to disallow redirects (this also solves a behaviour change I flagged in another comment) and when disallowed no clone of the Url is needed either.

> * I would have liked to use `impl Iterator<(&str, &str)>` for the headers, but they need to be `Send` (and probably `Clone` as well?), so creating an intermediate `BTreeMap` right now. I think its possible to avoid that as well maybe.

Can an iterator not be `Send`?

src/services/download/http.rs
jan-auer: `BTreeMap` is generally the choice unless you're dealing with incredibly large maps. In small cases it is not different to a sorted vector, and then it fans out very nicely without the overhead of hashing the values. Depending on the hashing function, that's usually slower than inserting into moderately sized BTreeMaps.

flub: TIL, thanks!

jan-auer:
> Can an iterator not be `Send`?

An implementation of `Iterator` can be `Send`, but `impl Iterator` is not. And unfortunately, you cannot specify `impl Iterator + Send`. However, I'd discourage doing that in any case over writing a generic method signature `<I: Iterator + Send>`, which suddently would allow you this. 

That said, can we figure out if we can get rid of the `Send` requirement?

Swatinem:
> can we figure out if we can get rid of the `Send` requirement?

I think that comes from here: https://github.com/getsentry/symbolicator/blob/eac35a6058c7749bdf20ed219a377e49e02d0b76/src/utils/http.rs#L34

BTW, I mentioned somewhere else that the retry/redirect logic is a bit broken. Right now it does `A -> B -> C (timeout); A -> B -> C (timeout)`, It would make a lot more sense if it would do `A -> B -> C (timeout); C (timeout)`. So it may be tempting to completely inline all of this into one function?

jan-auer:
> I think that comes from here:

Actually, that requirement is no longer needed. Let's remove that, then.

> So it may be tempting to completely inline all of this into one function?

That's a good observation! You can probably keep `retry` its own function (as it is much more generic and has nothing to do with HTTP requests), and then use it within `follow_redirects`.

src/services/download/mod.rs
```diff
@@ -206,7 +202,7 @@ fn download_future_stream(
...
     stream: impl Stream<Item = Result<Bytes, DownloadError>>,
```
jan-auer, line 206: _nit_: Just noticing this. Rather than making this `impl Stream`, can we make this a type parameter on the function? That should allow for proper propagation of `Send` and the likes in all cases.

Swatinem:
Closing this, as @jan-auer took care of this code

PR #250, opened 2020-07-10
Convert the sentry downloader to async fn

flub:
Upgrades the sentry downloader to use async fn internally.

This also adds a generic timed_task utilitiy to the futures module and
an extenstion trait to use it more ergonomically.

### DIFF
--- a/src/services/download/sentry.rs
+++ b/src/services/download/sentry.rs
@@ -16,7 +16,6 @@ use failure::Fail;
 use futures::compat::{Future01CompatExt, Stream01CompatExt};
 use futures::prelude::*;
 use parking_lot::Mutex;
-use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use url::Url;
 
 use super::{DownloadError, DownloadErrorKind, DownloadStatus, USER_AGENT};
@@ -57,14 +56,7 @@ pub async fn download_source(
 ) -> Result<DownloadStatus, DownloadError> {
     let download_url = source.download_url(&file_id);
     log::debug!("Fetching debug file from {}", download_url);
-    let mut backoff = ExponentialBackoff::from_millis(10).map(jitter).take(3);
-    let response = loop {
-        let result = start_request(&source, &download_url).await;
-        match backoff.next() {
-            Some(duration) if result.is_err() => future_utils::delay(duration).await,
-            _ => break result,
-        }
-    };
+    let response = future_utils::retry(|| start_request(&source, &download_url)).await;
 
     match response {
         Ok(response) => {
@@ -200,7 +192,7 @@ pub async fn list_files(
     let search = cached_sentry_search(query);
     let entries = future_utils::time_task("downloads.sentry.index", search)
         .await?
-        .iter()
+        .into_iter()
         .map(|search_result| {
             SourceFileId::Sentry(source.clone(), SentryFileId::new(search_result.id.clone()))
         })
--- a/src/services/download/sentry.rs
+++ b/src/services/download/sentry.rs
@@ -10,19 +10,19 @@ use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use actix::{Actor, Addr};
-use actix_web::client::ClientConnector;
+use actix_web::client::{ClientConnector, ClientResponse, SendRequestError};
 use actix_web::{client, HttpMessage};
 use failure::Fail;
-use futures01::future::Either;
-use futures01::prelude::*;
+use futures::compat::{Future01CompatExt, Stream01CompatExt};
+use futures::prelude::*;
 use parking_lot::Mutex;
 use tokio_retry::strategy::{jitter, ExponentialBackoff};
-use tokio_retry::Retry;
 use url::Url;
 
 use super::{DownloadError, DownloadErrorKind, DownloadStatus, USER_AGENT};
 use crate::sources::{FileType, SentryFileId, SentrySourceConfig, SourceFileId};
 use crate::types::ObjectId;
+use crate::utils::futures::{delay, TimedTaskExt};
 
 lazy_static::lazy_static! {
     static ref CLIENT_CONNECTOR: Addr<ClientConnector> = ClientConnector::default().start();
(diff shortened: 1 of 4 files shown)


src/utils/futures.rs
```diff
@@ -280,6 +280,39 @@ pub async fn delay(duration: Duration) {
...
+    let ret = future.await;
+    metric!(
+        timer("futures.done") = start_time.elapsed(),
+        "task_name" => task_name,
+        "status" => if ret.is_ok() { "ok" } else {"err"},
+    );
+    ret
+}
+
+pub trait TimedTaskExt<T, E>: Future<Output = Result<T, E>> + Sized {
```
flub, line 300: I understand this might be controversial.  I think it reads nicer when using it, but don't feel overly strong.

src/services/download/sentry.rs
```diff
@@ -10,91 +10,86 @@ use std::sync::Arc;
...
+async fn start_request(
+    source: &SentrySourceConfig,
+    url: &Url,
+) -> Result<ClientResponse, SendRequestError> {
+    // The timeout is for the entire HTTP download *including* the response stream itself,
+    // in contrast to what the Actix-Web docs say. We have tested this manually.
+    //
+    // The intent is to disable the timeout entirely, but there is no API for that.
+    client::get(url)
+        .with_connector((*CLIENT_CONNECTOR).clone())
```
Swatinem, line 42: Interesting. We don’t use this in any of the other http-based downloaders. Why is that?

src/utils/futures.rs
Swatinem: This sure is controversial. I do think that chaining combinators is a lot nicer to read than wrapping in functions.
The problem right now is that it is both super ugly to implement, and less performant due to boxing. I also predict that stable async-fn-in-traits will probably take >1 year :-(

Swatinem:
Oh, and apart from the `with_connector` thingy which I don’t understand fully yet, this might also be a good candidate to share with #249.

flub:
> Oh, and apart from the `with_connector` thingy which I don’t understand fully yet, this might also be a good candidate to share with #249.

I was hoping this would be merged before #249 as that's still in draft, then it's the problem of that PR to integrate it here too ;)

src/utils/futures.rs
flub: Is boxing really that terrible?  We do it all over the place now with the futures01/03 migration.  Doesn't seem to cause that much issues.

src/services/download/sentry.rs
flub: Maybe this is only historical to tweak SSL certificates to Sentry or something?

jan-auer: This allows us to bypass the restriction for internal IPs that is imposed on the globally registered connector. The rationale is that symbol servers are usually supposed to be external, but Sentry is never. So the default is to restrict all HTTP requests to the internal network, except for the Sentry source. See:
https://github.com/getsentry/symbolicator/blob/56e16ce86170d6d2ba91d530043b7764b7cb9560/src/utils/http.rs#L194

And particularly:
https://github.com/getsentry/symbolicator/blob/56e16ce86170d6d2ba91d530043b7764b7cb9560/src/utils/http.rs#L20

src/services/download/sentry.rs
```diff
@@ -10,91 +10,86 @@ use std::sync::Arc;
...
-        e => panic!("{}", e),
-    });
+    let mut backoff = ExponentialBackoff::from_millis(10).map(jitter).take(3);
+    let response = loop {
+        let result = start_request(&source, &download_url).await;
+        match backoff.next() {
+            Some(duration) if result.is_err() => delay(duration).await,
+            _ => break result,
+        }
+    };
```
jan-auer, line 67: As often as we have this now, how about we create a `retry` utility for the downloaders? Doesn't have to be very configurable, I'm thinking:

```rust
async fn retry<T, E, F, C>(f: C) -> Result<T, E>
where
  C: FnMut() -> F,
  F: Future<Item = Result<T, E>>;
```

I believe that it should work just fine with `FnMut`, but not 100% sure.

src/utils/futures.rs
jan-auer: In the long run, it would be nicer to get rid of Boxes around futures. We mostly add them where they help with keeping types small or when you have to join types from two paths. Also adding a comment on this specific signature.

src/utils/futures.rs
```diff
@@ -280,6 +280,39 @@ pub async fn delay(duration: Duration) {
...
+    fn task_metrics<'s>(
+        self,
+        task_name: &'s str,
+    ) -> std::pin::Pin<Box<dyn Future<Output = Self::Output> + 's>>
```
jan-auer, line 304: On first glance, this is rather strange to look at. A few thoughts and ideas:

 - Can't you just return a plain `Box` here? When `.await`'ing, it will be pinned for you, usually. I might be missing something, though.
 - Usually, we would return a named type here. This becomes more complicated when calling an `async` function, of course. But if you prefer this pattern, I would model it more after tokio's [`Stream::timeout`](https://docs.rs/tokio/0.2.21/tokio/stream/trait.StreamExt.html#method.timeout).
- Since you have an async function already, why not just expose it directly? Note that neither `async-std` nor `tokio` do extension traits anymore and prefer plain functions, which support a more imperative coding style. It's not that unergonomic:

```rust
let search = cached_sentry_search(query);
let result = measure("downloads.sentry.index", search).await?;
```

- _nit_: You'll notice that I called the function in imperative language.

src/services/download/sentry.rs
flub: Oh, that's important.  I added it to notion page about the client upgrade

jan-auer: Thanks, I should've mentioned that much earlier!

Swatinem: But where exactly is this hooked up?

https://github.com/getsentry/symbolicator/blob/56e16ce86170d6d2ba91d530043b7764b7cb9560/src/services/download/sentry.rs#L28

This is the only place this is initialized, with the default options? Is this injected somehow via DI? Or overwritten globally?

jan-auer: The sentry client uses this exact instance via `with_connector`. All other downloaders use the global connector started via `start_safe_connector`. You can see how this is added to the system registry:

https://github.com/getsentry/symbolicator/blob/56e16ce86170d6d2ba91d530043b7764b7cb9560/src/utils/http.rs#L199

By default, every client request uses the system connector unless explicitly overridden, like above.

src/services/download/sentry.rs
```diff
@@ -204,20 +208,18 @@ pub fn list_files(
...
-                SourceFileId::Sentry(source.clone(), SentryFileId::new(api_response.id))
-            })
-            .collect()
-    }));
-
-    Box::new(entries)
+    let search = cached_sentry_search(query);
+    let entries = future_utils::time_task("downloads.sentry.index", search)
+        .await?
+        .iter()
```
jan-auer, line 219: Can you use `into_iter` here? That should avoid the clone of `search_result.id`.
_nit_: You can also break this up in two statements. One that awaits, and then the next one that maps.

flub:
Now without an extension crate and with a basic retry function.

src/utils/futures.rs
```diff
@@ -9,6 +9,7 @@ use futures::{future, FutureExt, TryFutureExt};
 use futures01::future::Future as Future01;
 use tokio::prelude::FutureExt as TokioFutureExt;
 use tokio::runtime::Runtime as TokioRuntime;
+use tokio_retry::strategy::{jitter, ExponentialBackoff};
```
jan-auer, line 12: _nit_: In a follow-up, we should migrate this over to `backoff`.

src/utils/futures.rs
```diff
@@ -280,6 +281,39 @@ pub async fn delay(duration: Duration) {
...
+    metric!(
+        timer("futures.done") = start_time.elapsed(),
+        "task_name" => task_name,
+        "status" => if ret.is_ok() { "ok" } else {"err"},
+    );
+    ret
+}
+
+/// Retry a future 3 times with exponential backoff.
+pub async fn retry<G, F, T, E>(task_gen: G) -> Result<T, E>
```
jan-auer, line 302: _nit_: Can you add unit tests for this function? There are great facilities that allow to mock the timer.

src/services/download/sentry.rs
flub: I wonder why clippy can't warn for `.into_iter()`

src/utils/futures.rs
flub: not going to happen for now

jan-auer: 👍  Agreed, this is easier with newer tokio.

flub:
> G2G after fixing the [linter errors](https://travis-ci.com/github/getsentry/symbolicator/jobs/360444309#L888)

Yep, they're because all the futures01 stuff can be cleaned up now that the other downloaders work is merged!

PR #251, opened 2020-07-10
ref: Update S3 downloader to async/await

Swatinem:
This is just a first step, as a newer rusoto has native support
for async/await, which we do not yet use.

A next step in this refactoring would probably be to directly stream
from an `AsyncRead` into an `AsyncWrite`, without manually dealing
with streams.


Swatinem:
> I presume this is once we're on tokio 0.2?

yes, that would be the first step I guess. tokio 0.2 has its own `File` which supports `AsyncWrite`.

PR #254, opened 2020-07-22
Refactor hitcounter fixture, disable pytest-rerunfailure

flub:
The hitcounter is refactored to be more modular.  It is now more
encapsulated in an object rather than have a free closure variable
bound to itself.  This work also adds some more docstrings so that
pytest --fixtures shows more useful output.

The hitcounter now also avoids raising failures in the teardown,
instead using a mechanism to raise them during the call step of the
testrun protocol.

test_timeouts gets an explicit sleep step.  Its 10 iterations need to
take longer than its 1s timeout.  This should reduce flakyness and
hopefully remove the need for pytest-rerunfailures.

test_reserved_ip_addresses gets a more readable variable name and also
an additional assert to ensure the local proxy is (not) hit.

The Service fixute helper just does linear backoff instead of
exponential.  There is no point is sparing the server, we just want to
test to start being useful as quickly as possible and exponential
backoff adds needless delays.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,9 +22,9 @@ test-rust:
 .PHONY: test-rust
 
 test-integration: .venv/bin/python
-	.venv/bin/pip install -U pytest pytest-rerunfailures pytest-localserver requests pytest-xdist pytest-icdiff boto3
+	.venv/bin/pip install -U pytest pytest-localserver requests pytest-xdist pytest-icdiff boto3
 	cargo build --locked
-	@.venv/bin/pytest tests --reruns 5 -n12 -vv
+	@.venv/bin/pytest tests -n12 -vv
 .PHONY: test-integration
 
 # Documentation
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,6 @@
 [pytest]
 addopts = -ra --tb=native --durations 5 -p no:warnings
+testpaths = tests
 markers =
     extra_failure_checks: Marker which can add additional failure checks to
        a test.  It accepts the keyword argument `checks` which should be a
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,7 @@
 [pytest]
 addopts = -ra --tb=native --durations 5 -p no:warnings
+markers =
+    extra_failure_checks: Marker which can add additional failure checks to
+       a test.  It accepts the keyword argument `checks` which should be a
+       list of functions to call after the test finished executing.  These
+       functions can fail the test by calling `pytest.fail()`.
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -1,16 +1,17 @@
 import collections
+import json
+import os
+import resource
+import socket
 import subprocess
+import threading
+import time
 import traceback
 import uuid
-import time
-import socket
-import os
-import json
+
+import boto3
 import pytest
 import requests
-import threading
-import boto3
-
 from pytest_localserver.http import WSGIServer
 
 SYMBOLICATOR_BIN = [os.environ.get("SYMBOLICATOR_BIN") or "target/debug/symbolicator"]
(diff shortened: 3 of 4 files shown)


pytest.ini
```diff
@@ -1,2 +1,8 @@
 [pytest]
 addopts = -ra --tb=native --durations 5 -p no:warnings
+testpaths = tests
```
flub, line 3: Potentially this should be even `tests/integration`

tests/integration/test_basic.py
```diff
@@ -322,6 +322,7 @@ def test_timeouts(symbolicator, hitcounter):
             request_id = response["request_id"]
         else:
             assert False
+        time.sleep(0.3)  # 0.3 * 10 iterations = 3s => expect timeout on 4th iteration
```
flub, line 325: Not 100% sure I understood this test correctly.

tests/integration/conftest.py
```diff
@@ -1,14 +1,17 @@
...
 import time
-import socket
-import os
-import json
+import traceback
+import uuid
+
+import boto3
 import pytest
 import requests
```
flub, line 14: I ran isort on this

tests/integration/conftest.py
```diff
@@ -22,6 +25,35 @@
...
 
+@pytest.hookimpl
+def pytest_sessionstart(session):
+    no_fds_soft, no_fds_hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+    if no_fds_soft < 2028:
+        session.warn(
+            pytest.PytestWarning(
+                f"Too low open filedescriptor limit: {no_fds_soft} (try `ulimit -n 4096`)"
+            )
+        )
```
flub, line 36: This warning is a bit pants and very easily missed, I looks like this:
```
> pytest tests/integration/test_basic.py -k malformed_objects --collect-only                                                                                                           1s 400ms
/Users/flub/code/symbolicator:0: PytestWarning: Too low open filedescriptor limit: 400 (try `ulimit -n 4096`)
===================================================================================== test session starts =====================================================================================
platform darwin -- Python 3.8.2, pytest-5.4.3, py-1.8.1, pluggy-0.13.1
rootdir: /Users/flub/code/symbolicator, inifile: pytest.ini
plugins: rerunfailures-9.0, localserver-0.5.0, xdist-1.33.0, forked-1.1.3, icdiff-0.5
collected 30 items / 29 deselected / 1 selected
<Module tests/integration/test_basic.py>
  <Function test_malformed_objects>
```

Perhaps calling pytest.exit() is better, but it's very rude:
```
> pytest tests/integration/test_basic.py -k malformed_objects                                                                                                                             581ms
Exit: Too low open filedescriptor limit: 400 (try `ulimit -n 4096`)
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! _pytest.outcomes.Exit: Too low open filedescriptor limit: 400 (try `ulimit -n 4096`) !!!!!!!!!!!!
```

If we prefer this perhaps we need an command line option to ignore the check

Makefile
```diff
@@ -22,9 +22,9 @@ test-rust:
 .PHONY: test-rust
 
 test-integration: .venv/bin/python
-	.venv/bin/pip install -U pytest pytest-rerunfailures pytest-localserver requests pytest-xdist pytest-icdiff boto3
+	.venv/bin/pip install -U pytest pytest-localserver requests pytest-xdist pytest-icdiff boto3
```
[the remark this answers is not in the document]
untitaker: not really, I introduced pytest-localserver to relay because I've been using it for ages. But I think it's unmaintained.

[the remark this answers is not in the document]
flub: s/unmaintained/no one is going to break things for you/

PR #260, opened 2020-08-31
fix: Do not cache sentry index if caches are disabled

untitaker:


src/services/download/sentry.rs
```diff
@@ -144,9 +148,12 @@ async fn fetch_sentry_json(query: &SearchQuery) -> Result<Vec<SearchResult>, Sen
 ///
 /// If there are cached search results this skips the actual search.
 async fn cached_sentry_search(query: SearchQuery) -> Result<Vec<SearchResult>, DownloadError> {
-    if let Some((created, entries)) = SENTRY_SEARCH_RESULTS.lock().get(&query) {
-        if created.elapsed() < Duration::from_secs(3600) {
-            return Ok(entries.clone());
+    if let Some(cache_duration) = query.index_cache_ttl {
+        unreachable!();
```
flub, line 152: this is possibly a debugging leftover?

flub (commented):
Ouch, apologies for not recalling that happened there.

I think I would rather remove the caching from the downloader entirely and move it to `ObjectsActor::find()` though.  I never wanted the download service to have any knowledge or configuration about any of this.

untitaker:
ok, but the same in-memory LRU cache, just moved to objectsactor, right?

flub:
> ok, but the same in-memory LRU cache, just moved to objectsactor, right?

Yes, I don't think there's a reason to change this.

> would the objectsactor now cache all source lookups, even the ones that don't have IO?

That's a good point.  The key would be `FindObject` I guess which is rather different but not bad I guess.

Caching all the responses rather than just those to sentry based on this key would probably result in a much larger cache though, for no actual gain however.  So that's perhaps more questionable.  You could only cache those with a `SentrySourceConfig` but now these two depend on implementation details of each other.  So both of these aren't very nice options.

Maybe your solution here is the least bad of all after all.  Though theoretically I prefer the objectsactor to do the caching and take the memory hit.  I just don't know how large that memory hit is.

untitaker:
I'm going to keep it as-is as the cache is too specific to Sentry search results. One could theoretically do the same patch for connection pools and GCP token caches, but if we want to cache those the behavior would be different from how we cache Sentry search results and a better fit for the download actor.

jan-auer:
> Caching all the responses rather than just those to sentry based on this key would probably result in a much larger cache though, for no actual gain however.

Yeah, I think this is the key point. Beyond caching actual debug files, I think we should be extremely conservative in caching further responses. The only reason this cache was introduced in the first place is because we need to issue a request for **every** request in order to even obtain a list of candidate images. 

I believe that this can be changed in the Sentry source, but it would require us to rework the caching system. Particularly, rather than caching by the source-specific path, we should just cache all files we can get by their debug identifier (and source id). Then, we can skip lookups as soon as we find a cached file. I do remember that this was the original design, but I cannot recall reasons why it wasn't implemented like this.

> Though theoretically I prefer the objectsactor to do the caching and take the memory hit.

I agree with this. The object service (! 😄 ) should be responsible of all caching and deduplication concerns, ideally keeping files open in memory in addition to that. There can be a lower-level service actually handling file writes and such. That said, sources *can* have caches, such as credentials and session tokens. In that sense, I view the current Sentry source's cache as an implementation detail that we can hopefully get rid of.

src/services/download/mod.rs
```diff
@@ -128,14 +130,30 @@ impl DownloadService {
         hub: Arc<Hub>,
     ) -> impl Future<Output = Result<Vec<SourceFileId>, DownloadError>> {
         let worker = self.worker.clone();
+
+        // The Sentry cache index should expire as soon as we attempt to retry negative caches.
+        let sentry_index_cache_ttl = if self.config.cache_dir.is_some() {
+            self.config.caches.downloaded.retry_misses_after
+        } else {
+            None
+        };
```
jan-auer, line 139: This is an implementation detail of the sentry source. Can you move this into the `sentry` source and rather pass the entire config in?

@flub this is probably another good reason to make sources `structs`. That way, we get rid of `lazy_static`s and we can hold on to a config reference so we don't have to pipe arbitrary attributes through multiple methods (e.g. having the TTL on `SearchQuery` is odd).

src/services/download/sentry.rs
```diff
@@ -144,9 +148,11 @@ async fn fetch_sentry_json(query: &SearchQuery) -> Result<Vec<SearchResult>, Sen
 ///
 /// If there are cached search results this skips the actual search.
 async fn cached_sentry_search(query: SearchQuery) -> Result<Vec<SearchResult>, DownloadError> {
-    if let Some((created, entries)) = SENTRY_SEARCH_RESULTS.lock().get(&query) {
-        if created.elapsed() < Duration::from_secs(3600) {
-            return Ok(entries.clone());
+    if let Some(cache_duration) = query.index_cache_ttl {
```
jan-auer, line 151: The `index_cache_ttl` should just be `Duration` as `Duration::default()` is 0. Particularly, this gets rid of a branch here, and the comparison below still works.

src/services/download/mod.rs
untitaker: ok done. this clones the arc unfortunately. we use async and std futures, but unfortunately we need to guarantee that those are `'static` because we convert them into old futures somewhere, I think. I'll attempt this separately

src/services/download/sentry.rs
untitaker: I did this, the entire config reading is now in this function, so the amount of conditionals _in tihs function_ did not decrease as it gained the conditionals that were previously in the downloads actor

untitaker: yeah no dice, even our own Worker requires static even though it has new-style futures

src/services/download/mod.rs
```diff
@@ -128,15 +130,16 @@ impl DownloadService {
         hub: Arc<Hub>,
     ) -> impl Future<Output = Result<Vec<SourceFileId>, DownloadError>> {
         let worker = self.worker.clone();
+        let config = self.config.clone();
+
         async move {
             match source {
                 SourceConfig::Sentry(cfg) => {
+                    let job =
+                        move || sentry::list_files(cfg, filetypes, object_id, config).bind_hub(hub);
```
flub, line 139: Does this always clone or is the compiler smart enough to only clone for the sentry source?

flub (approved):

src/services/download/mod.rs
flub: @jan-auer not sure i follow, you mean `SourceConfig` should be a struct and you want to attach the caches to that?

src/services/download/mod.rs
untitaker: This likely always clones, I think I would have to write two match stmts to effectively clone only once.

untitaker: I don't thikn I can do this, it makes my future non-static either way

PR #265, opened 2020-09-04
feat(cache): Write temporary files in a dedicated directory

flub:
This is not as clean as it should be because all locations do a lot of
path manipulation manually.


src/actors/common/cache.rs
```diff
@@ -42,6 +41,12 @@ pub struct Cacher<T: CacheItemRequest> {
     current_computations: ComputationMap<T::Item, T::Error>,
 }
 
+impl<T: CacheItemRequest> Cacher<T> {
```
untitaker, line 44: There is already such an impl block

src/cache.rs
```diff
@@ -323,26 +347,58 @@ pub struct Caches {
 
 impl Caches {
     pub fn from_config(config: &Config) -> io::Result<Self> {
+        let tmp_dir = config.cache_dir("tmp");
+        if let Some(ref tmp) = tmp_dir {
+            if tmp.exists() {
```
untitaker, line 352: If the cleanup job starts, won't this get executed and the tmpdir wiped even though symbolicator is still working?

src/cache.rs
```diff
@@ -96,8 +96,19 @@ pub struct Cache {
...
     cache_dir: Option<PathBuf>,
 
+    /// Directory to use for temporary files.
+    ///
+    /// When writing a new file into the cache it is best to write it to a temporary file in
+    /// a sibling directory, once fully written it can then be atomically moved to the
+    /// actual location withing the `cache_dir`.
+    ///
+    /// Just like for `cache_dir` when this cache is disabled this will be `None`.
+    tmp_dir: Option<PathBuf>,
```
untitaker, line 110: Is there a situation where one wants to make this configurable? afaict it is not exposed as config param but I think we should avoid exposing it entirely.. I think `NamedTempFile::new()` is never what you want

untitaker (approved):

src/cache.rs
flub: ugh, good catch!

src/cache.rs
flub: My intention was also not to expose this.  You only get to chose a cache directory (or not) and we decide what to do inside of there.  It makes me a little sad that you can create a `Cache` with `cache_dir=Some(...)` and `tmp_dir=None` or the other way around.  But only code in this module should be able to do that.

src/cache.rs
```diff
@@ -403,6 +397,20 @@ impl Caches {
         })
     }
 
+    /// Clear the temporary files.
```
untitaker, line 400: I think we can make a case for emitting an error to Sentry if the tmpdir is not empty on restart

flub: This should only happen after a crash and there probably have been alerts for this already in that case?  I fear this might just lead to duplicate reporting.  It's not really a bad failure mode, just a backstop to some files never being cleaned up

untitaker (approved):

Swatinem (approved):

PR #267, opened 2020-09-11
ref: Replace failure with anyhow for CLI errors

Swatinem:
#skip-changelog

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -35,7 +35,11 @@ fn main() {
     match cli::execute() {
         Ok(()) => std::process::exit(0),
         Err(error) => {
-            logging::ensure_log_error(&error);
+            if log::log_enabled!(log::Level::Error) {
+                log::error!("{:?}", error);
+            } else {
+                eprintln!("{:?}", error);
+            }
             std::process::exit(1);
         }
     }
--- a/src/main.rs
+++ b/src/main.rs
@@ -35,11 +35,7 @@ fn main() {
     match cli::execute() {
         Ok(()) => std::process::exit(0),
         Err(error) => {
-            if log::log_enabled!(log::Level::Error) {
-                log::error!("{:?}", error);
-            } else {
-                eprintln!("{:?}", error);
-            }
+            logging::ensure_log_error(&error);
             std::process::exit(1);
         }
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -163,6 +163,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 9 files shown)


flub:
This is effectively replacing specific error types with generic ones.  I'm not sure I'm a fan.

To me the more direct replacement of these is via thiserror, something like
```
#[derive(Debug, thiserror::Error)
pub enum ServiceStateError {
    #[error("failed to create process pool")]
    Spawn(#[from] SpawnError),
    ...
}
```

The main downside of this is that the chained errors are no longer boxed, which is what the failure context did.

Swatinem:
> This is effectively replacing specific error types with generic ones.

That’s why you need to evaluate this on a case-by-case basis.
For this specific usecase, we never ever match on the specific error types, that’s why I think `anyhow` is the way to go here.

Looking at the downloader errors, etc, I think that is where we would rather want to use `thiserror`. Also note that those currently throw away all the child errors.

flub:
> Also note that those currently throw away all the child errors.

Not sure this is true.  Failure's Context kept the original errors available and you can show the entire chain

Swatinem:
Oh, what I mean is that you can’t `match` on the specific errors:

https://github.com/getsentry/symbolicator/blob/d95667c8505b0a02ca194759a6f59d9a5b1ab152/src/services/download/mod.rs#L37-L49

Its just an opaque `DownloadErrorKind::Sentry`. Or maybe I misunderstand something.

src/main.rs
```diff
@@ -35,7 +35,11 @@ fn main() {
     match cli::execute() {
         Ok(()) => std::process::exit(0),
         Err(error) => {
-            logging::ensure_log_error(&error);
+            if log::log_enabled!(log::Level::Error) {
+                log::error!("{:?}", error);
+            } else {
+                eprintln!("{:?}", error);
+            }
```
jan-auer, line 42: Any particular reason this was inlined? The doc string on the method was quite useful. Also, please keep the `LogError` wrapper to print causes in the format `LogError` does.

jan-auer (changes requested):
LGTM. One requested change regarding `LogError` and `ensure_log_error`.

src/main.rs
Swatinem: Its a bit weird and one thing I kind of dislike about anyhow:

`{:?}` prints the error with all its causes and a backtrace if available, as explained here:

https://docs.rs/anyhow/1.0.32/anyhow/struct.Error.html#display-representations

And the function was inlined because it was the only usage. I can separate it out again though.

jan-auer: I see. I'm curious how this will work when we have both `std::Error` and `anyhow::Error`, although at this level it's only going to be `anyhow`.

jan-auer (approved):

PR #268, opened 2020-09-11
ref: Replace failure with anyhow in symsorter

Swatinem:
Also updated symbolic dependency version because of the lifetime inference problem.

#skip-changelog

### DIFF
--- a/symsorter/src/app.rs
+++ b/symsorter/src/app.rs
@@ -332,9 +332,9 @@ pub fn main() -> ! {
         Ok(()) => std::process::exit(0),
         Err(error) => {
             eprintln!("{}: {}", style("error").red().bold(), error);
-            error.chain().skip(1).for_each(|cause| {
+            for cause in error.chain().skip(1) {
                 eprintln!("{}", style(format!("  caused by {}", cause)).dim());
-            });
+            }
 
             eprintln!();
             eprintln!("To skip this error, use --ignore-errors.");
--- a/symsorter/src/app.rs
+++ b/symsorter/src/app.rs
@@ -4,9 +4,9 @@ use std::io;
 use std::path::{Path, PathBuf};
 use std::time::Duration;
 
+use anyhow::{anyhow, Context, Result};
 use chrono::{DateTime, Utc};
 use console::style;
-use failure::{err_msg, Error, Fail};
 use serde::Serialize;
 use serde_json;
 use structopt::StructOpt;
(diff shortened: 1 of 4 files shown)


jan-auer:
You can shim `symbolic`'s errors similar to:

https://github.com/getsentry/relay/blob/bbb063c512ee113986d12117c49577e4635c75bc/tools/scrub-minidump/src/main.rs#L80-L83

symsorter/src/app.rs
```diff
@@ -124,13 +121,15 @@ fn process_file(
...
     };
-    let archive = maybe_ignore_error!(Archive::parse(&bv));
+    let archive = maybe_ignore_error!(
+        Archive::parse(&bv).map_err(|e| anyhow!("failed to parse archive {}", e))
+    );
     let root = &RunConfig::get().output;
 
     for obj in archive.objects() {
-        let obj = maybe_ignore_error!(obj);
+        let obj = maybe_ignore_error!(obj.map_err(|e| anyhow!(e)));
```
jan-auer, line 130: NB: We need to remember to come back to these and replace them with `.context("xyz")` once we get `symbolic` fixed.

symsorter/src/app.rs
```diff
@@ -335,9 +332,9 @@ pub fn main() -> ! {
         Ok(()) => std::process::exit(0),
         Err(error) => {
             eprintln!("{}: {}", style("error").red().bold(), error);
-            for cause in error.iter_causes() {
+            error.chain().skip(1).for_each(|cause| {
```
jan-auer, line 335: Let's stick to the regular `for` syntax:

```suggestion
            for cause in error.chain().skip(1) {
```

symsorter/src/utils.rs
```diff
@@ -54,21 +54,22 @@ pub fn get_target_filename(obj: &Object) -> Option<PathBuf> {
...
-            if writer.write_object(
-                &obj,
-                &path.file_name().unwrap().to_string_lossy().to_string(),
-            )? {
+            let writer =
+                SourceBundleWriter::start(Cursor::new(&mut out)).map_err(|e| anyhow!(e))?;
+            if writer
+                .write_object(
+                    &obj,
+                    &path.file_name().unwrap().to_string_lossy().to_string(),
```
jan-auer, line 69: _nit_: While we're at it, this could be moved into a temporary variable for readability.

jan-auer (approved):

symsorter/src/utils.rs
Swatinem: also the `to_string` is useless there ;-)

PR #269, opened 2020-09-11
ref: Replace failure with thiserror in the download service

Swatinem:
I think @jan-auer had a few ideas about making these errors better, especially around authentication errors on upstream services (gcs, s3), so this might be a good idea to figure out what we want to do here.

#skip-changelog

### DIFF
--- a/src/services/download/mod.rs
+++ b/src/services/download/mod.rs
@@ -38,19 +38,19 @@ const USER_AGENT: &str = concat!("symbolicator/", env!("CARGO_PKG_VERSION"));
 /// Errors happening while downloading from sources.
 #[derive(Debug, Error)]
 pub enum DownloadError {
-    #[error("Failed to download")]
+    #[error("failed to download")]
     Io(#[source] std::io::Error),
-    #[error("Failed to download stream")]
+    #[error("failed to download stream")]
     Stream(#[source] failure::Compat<PayloadError>),
-    #[error("Bad file destination")]
+    #[error("bad file destination")]
     BadDestination(#[source] std::io::Error),
-    #[error("Failed writing the downloaded file")]
+    #[error("failed writing the downloaded file")]
     Write(#[source] std::io::Error),
-    #[error("Download was cancelled")]
+    #[error("download was cancelled")]
     Canceled,
-    #[error("Failed to fetch data from GCS")]
+    #[error("failed to fetch data from GCS")]
     Gcs(#[from] gcs::GcsError),
-    #[error("Failed to fetch data from Sentry")]
+    #[error("failed to fetch data from Sentry")]
     Sentry(#[from] sentry::SentryError),
 }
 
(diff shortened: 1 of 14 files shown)


src/services/download/mod.rs
```diff
@@ -34,26 +34,21 @@ pub use crate::types::ObjectId;
...
-    #[fail(display = "bad file destination")]
-    BadDestination,
-    #[fail(display = "failed writing the downloaded file")]
-    Write,
-    #[fail(display = "download was cancelled")]
+#[derive(Debug, Error)]
+pub enum DownloadError {
+    // TODO: this is currently an opaque `failure::Error`, as we pack all kinds of errors into this
+    #[error("failed to download")]
+    Io(failure::Error),
```
jan-auer, line 41: Would it work if you make this `Box<dyn Error + 'static>` or even `Box<dyn Error + Send + Sync + 'static>`? 

Also, consider implementing a utility function for `map_err`:

```rust
impl DownloadError {
    fn io<E>(error: E) -> Self 
    where
        E: Error + 'static
    {
        Self::Io(Box::new(error))
    }
}

some_action().map_err(DownloadError::io)?
```

src/services/download/mod.rs
```diff
@@ -34,26 +34,21 @@ pub use crate::types::ObjectId;
...
+    #[error("bad file destination")]
+    BadDestination(std::io::Error),
```
flub, line 43: You should probably annotate all these variants like `BadDestination(#[source] std::io::Error)` should you not?

src/services/download/sentry.rs
```diff
@@ -109,11 +99,17 @@ struct SearchQuery {
...
 
-symbolic::common::derive_failure!(
-    SentryError,
-    SentryErrorKind,
-    doc = "Errors happening while fetching data from Sentry"
-);
+#[derive(Debug, Error)]
+pub enum SentryError {
+    #[error("failed parsing JSON response from Sentry")]
+    Parsing(JsonPayloadError),
```
flub, line 105: here you can even use `#[from]` instead of `#[source]` and you won't need the `.map_err()` calls in the code.

src/services/download/mod.rs
flub: Alternatively we're piling too many things into one variant and should use more variants?

src/services/download/sentry.rs
Swatinem: No I can’t do neither, since `failure::Error` does not `impl std::error::Error` :-(

src/services/download/mod.rs
Swatinem: I introduced a separate `GcsError` and split up the `DownloadError::Io` into better variants.

The problem is that quite some foreign error types, most notably those from actix_web don’t `impl std::error::Error` and other traits such as `Clone`, etc :-(

src/services/download/sentry.rs
Swatinem: I now use the `failure::Fail::compat` function to convert it to a `std::error::Error` in order to use `#[source]`.

flub (approved):

PR #270, opened 2020-09-15
ref: Convert symbolicator actors to thiserror

Swatinem:
the next iteration on top of #269 

#skip-changelog

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -12,6 +12,14 @@
       "args": ["run", "-c", "local.yml"],
       "cwd": "${workspaceFolder}",
       "preLaunchTask": "cargo build"
+    },
+    {
+      "type": "lldb",
+      "request": "attach",
+      "name": "Attach to running process",
+      "program": "${workspaceFolder}/target/debug/symbolicator",
+      "stopOnEntry": true,
+      "pid": "${command:pickMyProcess}"
     }
   ]
 }
--- a/src/actors/cficaches.rs
+++ b/src/actors/cficaches.rs
@@ -4,52 +4,46 @@ use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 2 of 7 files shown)


.vscode/launch.json
```diff
@@ -12,6 +12,14 @@
...
       "cwd": "${workspaceFolder}",
       "preLaunchTask": "cargo build"
+    },
+    {
+      "type": "lldb",
+      "request": "attach",
+      "name": "Attach to running process",
+      "program": "${workspaceFolder}/target/debug/symbolicator",
+      "stopOnEntry": true,
+      "pid": "${command:pickMyProcess}"
```
jan-auer, line 22: Do we need both `program` and `pid`?

src/services/download/mod.rs
```diff
@@ -34,25 +36,30 @@ pub use crate::types::ObjectId;
...
+    Sentry(#[from] sentry::SentryError),
 }
 
-symbolic::common::derive_failure!(
-    DownloadError,
-    DownloadErrorKind,
-    doc = "Errors happening while downloading from sources."
-);
+impl DownloadError {
+    pub fn stream(err: PayloadError) -> Self {
```
flub, line 59: Not sure I like this naming much, reads more as a method than a constructor.

flub (approved):
I like how this seems to get easier to read everywhere

.vscode/launch.json
Swatinem: otherwise it will throw an error if there are multiple instances running, like they are for symbolicator…

src/services/download/mod.rs
jan-auer: probably `from_stream` would be better. The intention here was to make those method names as short as possible since we use them in a lot of places and I wanted to avoid ugly line breaking.

PR #271, opened 2020-09-15
ci(craft): Add docker target to craft

BYK:
`symbolicator:0.2.0` was missing from Docker Hub as it was not defined in Craft as a publish target. This PR adds it. Whoever releases Symbolicator will now need write access to https://hub.docker.com/r/getsentry/symbolicator


BYK:
I think now is a good time to add a release action via GitHub Actions.

jan-auer (approved):

PR #273, opened 2020-09-29
feat: Add an option to configure subprocess pool size

tonyo:
This can be useful when Symbolicator is not the only service running on the machine (e.g. in Kubernetes environment).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fix a bug where Sentry requests were cached even though caches were disabled. ([#260](https://github.com/getsentry/symbolicator/pull/260))
 - Publish Docker containers to DockerHub at `getsentry/symbolicator`. ([#271](https://github.com/getsentry/symbolicator/pull/271))
+- Add `processing_pool_size` configuration option that allows to set the size of the processing pool ()
 
 ## 0.2.0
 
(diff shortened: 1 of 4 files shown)


Swatinem (approved):

jan-auer (commented):
You will also need to update this here, which creates our standard thread pools:

https://github.com/getsentry/symbolicator/blob/2cad3b1924a0b8fcd7a55687ce50a8d104fcf3a9/src/utils/futures.rs#L53

jan-auer (changes requested):
Sorry, meant to request changes.

tonyo:
@jan-auer you sure we want to have one setting for everything? Technically, there's also a worker pool in actix HttpServer where we can also adjust worker count
https://github.com/getsentry/symbolicator/blob/master/src/server.rs#L32

jan-auer:
That's a good question. We may want to split this, indeed. 

Some of these thread pools used to perform CPU-intensive work, but do no longer due to subprocessing. So in that case, we may consider to dial them down independently, so I think your patch is fine.

Should we maybe clarify in docs then that this is the number of sub processes?

flub (approved):

tonyo:
@jan-auer updated the docstring, but in the we already have this:

image

jan-auer (approved):

PR #274, opened 2020-10-01
ref: Use failure-less symbolic

Swatinem:
#skip-changelog


jan-auer (approved):
Note that we're losing stack traces in some of the more sensitive parts. @flub are you ok with this?

PR #277, opened 2020-10-27
ref(objects): Simplify decompression calls

flub:
Doesn't change anything but simplify the caller.  As there's only one
might as well depend on the NamedTempFile struct.

----

#skip-changelog

### DIFF
--- a/src/actors/objects/mod.rs
+++ b/src/actors/objects/mod.rs
@@ -581,7 +581,7 @@ impl ObjectsActor {
     }
 }
 
-/// Decompresses and object file.
+/// Decompresses an object file.
 ///
 /// Some compression methods are implemented by spawning an external tool and can only
 /// process from a named pathname, hence we need a [NamedTemporaryFile] as source.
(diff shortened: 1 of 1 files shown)


src/actors/objects/mod.rs
```diff
@@ -585,23 +581,22 @@ impl ObjectsActor {
...
-) -> io::Result<fs::File> {
+/// Decompresses an object file.
+///
+/// Some compression methods are implemented by spawning an external tool and can only
+/// process from a named pathname, hence we need a [NamedTemporaryFile] as source.
+fn decompress_object_file(src: &NamedTempFile, mut dst: fs::File) -> io::Result<fs::File> {
     // Ensure that both meta data and file contents are available to the
     // subsequent reads of the file metadata and reads from other threads.
-    download_file.sync_all()?;
+    src.as_file().sync_all()?;
```
Swatinem, line 591: maybe introduce a `src_file` local instead of doing a call all the time?

Swatinem (approved):

src/actors/objects/mod.rs
flub: heh, i did that first but ended up not liking it.  It read a bit weird with two same-named variables but one of them being in hungarian notation.  This is more explicit and unambigiuous.

PR #279, opened 2020-10-29
fix(objects): Make sure returned files can be used for purpose

flub:
When the objects actor downloads files it ranks them and chooses the
highest ranking file.  However before returning we must make sure the
file can actually be used for the requested purpose.

In particular this avoids asking for CFI but only getting a (64-bit)
PDB but no PE file.  This file does not actually provide the requested
info, which then confuses the CfiActor because it tries to use it as a
CFI file, even though the file does not have CFI.

### DIFF
--- a/src/actors/common/cache.rs
+++ b/src/actors/common/cache.rs
@@ -195,6 +195,7 @@ impl<T: CacheItemRequest> Cacher<T> {
     /// This method does not take care of ensuring the computation only happens once even
     /// for concurrent requests, see the public [Cacher::compute_memoized] for this.
     fn compute(&self, request: T, key: CacheKey) -> ResponseFuture<T::Item, T::Error> {
+        // cache_path is None when caching is disabled.
         let cache_path = get_scope_path(self.config.cache_dir(), &key.scope, &key.cache_key);
         if let Some(ref path) = cache_path {
             if let Some(item) = tryf!(self.lookup_cache(&request, &key, &path)) {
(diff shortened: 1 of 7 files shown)


src/actors/common/cache.rs
```diff
@@ -207,39 +212,41 @@ impl<T: CacheItemRequest> Cacher<T> {
...
-                    status.persist_item(cache_path, temp_file)?;
-                    CachePath::Cached(cache_path.to_path_buf())
-                }
-                None => CachePath::Temp(temp_file.into_temp_path()),
-            };
-
-            Ok(request.load(key.scope.clone(), status, byteview, path))
-        });
+                Ok(request.load(key.scope.clone(), status, byteview, path))
+            });
```
flub, line 249: This code didn't change, I only added the type signature to the `|status: CacheStatus|` closure and the rest is rust-fmt's doing.

src/actors/objects/mod.rs
```diff
@@ -552,29 +596,44 @@ impl ObjectsActor {
...
+                                    object_meta.features.has_debug_info
+                                        || object_meta.features.has_symbols
+                                }
+                                ObjectPurpose::Source => object_meta.features.has_sources,
+                            }
+                        } else {
+                            true
+                        }
+                    }
+                    Err(_) => true,
```
flub, line 635: This `.filter()` is the actual bugfix.  The the minidump processor assumes that when it finds a file in the CFI cache for the module it has a file with valid cfi info.  In turn the file gets into the cfi cache because the cfi actor assumes that the object actor returns a file fit for purpose, and it asked for a CFI file.  However the object actor's ranking mechanism didn't respect that.  Adding this makes it respect this.

src/actors/objects/mod.rs
```diff
@@ -265,13 +293,15 @@ impl CacheItemRequest for FetchFileDataRequest {
...
                         let object = match object_opt {
                             Some(object) => object,
-                            None => return Ok(CacheStatus::Negative),
+                            None => {
+                                if archive.objects().all(|r| r.is_err()) {
+                                    return Ok(CacheStatus::Malformed);
+                                } else {
+                                    return Ok(CacheStatus::Negative);
+                                }
+                            }
```
flub, line 304: This fixes a bug where if an archive contained only malformed objects it would show up as missing rather than malformed.  Arguably this isn't aggressive enough and should return malformed if the object was not found and **any** member of the archive failed to be parsed.

src/actors/objects/mod.rs
Swatinem, line 635: just out of curiosity: the new filter now throws away things we can’t use, but why does it pass through errors and negative cache statuses?

Swatinem (approved):

src/actors/common/cache.rs
```diff
@@ -207,39 +212,41 @@ impl<T: CacheItemRequest> Cacher<T> {
...
-            if let Some(ref cache_path) = cache_path {
-                sentry::configure_scope(|scope| {
-                    scope.set_extra(
-                        &format!("cache.{}.cache_path", name),
-                        cache_path.to_string_lossy().into(),
-                    );
-                });
+        let future = request
+            .compute(temp_file.path())
+            .and_then(move |status: CacheStatus| {
```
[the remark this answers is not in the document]
flub: This is entirely a subjective readability choice.  I'd say this is no different than spelling out the types in functions, and this is a large enough function...

src/actors/objects/mod.rs
flub: because all those could have been object files which might have contained the info we needed.  but we don't know until we can actually fetch it.  and passing them through allows us to report on their status later (download error, malformed, negative cache).  this whole behaviour of only returning one result here a bit questionable and may have to change to improve reporting

src/actors/objects/mod.rs
```diff
@@ -552,29 +596,44 @@ impl ObjectsActor {
...
-                .min_by_key(|(i, response)| {
+                .min_by_key(|response| {
                     // Prefer files that contain an object over unparseable files
-                    let object = match response {
-                        Ok(object) => object,
+                    let object_meta = match response {
+                        Ok(object_meta) => object_meta,
                         Err(e) => {
                             log::debug!("Failed to download: {:#?}", e);
-                            return (3, *i);
```
untitaker, line 562: iirc we did this for stable `min` such that ties are resolved by order of sources.

src/actors/objects/mod.rs
```diff
@@ -552,29 +596,44 @@ impl ObjectsActor {
...
+                        ObjectPurpose::Debug if object_meta.features.has_symbols => 1,
+                        ObjectPurpose::Source if object_meta.features.has_sources => 0,
                         _ => 2,
-                    };
-
-                    (score, *i)
+                    }
+                })
+                .filter(|response| match response {
+                    // Make sure we only return objects which provide the requested info
```
untitaker, line 619: Can we dedupe this with the content in `min_by_key`? iiuc all scores of 2 should basically be filtered out.

Perhaps the closure of `min_by_key` should go into a `filter_map` before `min_by` happens. Or we refactor this entire thing into a regular for-loop.

untitaker (approved):

src/actors/objects/mod.rs
flub: The `i` here entirely depends on the order of the futures being executed above above calling `download_svc.list_files()`.  Practically this means there is no order, it's a race on who gets executed by the executor first (for non-sentry sources) or how fast the network is (for the sentry source).

At least that's my understanding and the reason I took it out.  I don't think it enforces anything and only makes the code look more complicated.

untitaker: `join_all` preserves order of futures. `list_files` can return any order it wants, but joining all `list_files` together resembles the source order.

In theory this gives you the possibility to upload your own Windows kernel symbols and let them have precedence over our own built-in symbol source. I am not sure if anybody relies on this.

jan-auer: This is definitely behavior we should retain.

src/actors/objects/mod.rs
flub: Thanks for complaining about this.  I've looked at all the consumers again and concluded the filtering needs to be moved to before the `min_by_key` in order to still return a possible `Err` if we filtered out an object file.

I'm not sure I would like to merge the two, I think having the two as separate operations is easier to understand as each have a defined scope.

untitaker: Yeah no worries was just an idea.

PR #280, opened 2020-10-30
fix(cfi): Treat no CFI as missing CFI

flub:
If the objects actor gives us an object that has no CFI, because it
could not find one and gave use the next best thing, treat it as if we
are missing the CFI.


flub:
FWIW #279 is an alternative PR for the same issue which is a better fix I think.  Leaving this here for now as reference while the other isn't merged yet.

PR #282, opened 2020-11-03
ci: Check rust doc comments on CI

flub:
This adds a check to ensure the rust doc comments are correct and can
be used to internally document the application.

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install rust stable toolchain
+        uses: actions-rs/toolchain@v1
         with:
           toolchain: nightly
           profile: minimal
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -118,6 +121,33 @@ jobs:
...
+        with:
+          toolchain: nightly
+          profile: minimal
+          components: rust-docs
+          override: true
+
+      - name: Cache rust cargo artifacts
+        uses: swatinem/rust-cache@v1
+        with:
+          key: ${{ github.job }}
```
Swatinem, line 144: you don’t need this anymore, and might as well remove it from the other jobs as well ;-)

.github/workflows/ci.yml
```diff
@@ -118,6 +121,33 @@ jobs:
           SENTRY_SYMBOLICATOR_TEST_AWS_SECRET_ACCESS_KEY: ${{ secrets.SENTRY_SYMBOLICATOR_TEST_AWS_SECRET_ACCESS_KEY }}
         run: pytest --ci -n12 -vv tests/integration
 
+  doc-comments:
+    name: Rust doc comments
+    runs-on: ubuntu-latest
+    env:
+      RUSTDOCFLAGS: -Dwarnings
```
Swatinem, line 128: not sure we should be doing it this way, according to the comment on the relay PR from a rust dev.

flub: Because this environment variable is set for every invocation of this job and does not change I don't think there's anything wrong with using it this way.  Cargo sees exactly the same flags when compiling all artefacts and re-uses the cache.  I just pushed a noop change and the entire cache got reused.

Swatinem (approved):

PR #283, opened 2020-11-06
ref: Use a more strongly-typed timestamp

flub:
This avoids duck-typing an integer as a timestamp.

Additionally this upgrades chrono and sorts dependencies in cargo.toml
alphabetically.

----

#skip-changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,56 +8,56 @@ include = ["src/**/*", "Cargo.toml", "README.md"]
 readme = "README.md"
 
 [dependencies]
-actix-web = { version = "0.7.19", features = ["tls"], default-features = false }
 actix = "0.7.9"
+actix-web = { version = "0.7.19", features = ["tls"], default-features = false }
 anyhow = "1.0.32"
 apple-crash-report-parser = "0.4.2"
 backtrace = "0.3.53"
+base64 = "0.11.0"
+bytes = { version = "0.4.12", features = ["serde"] }
+cadence = "0.18.0"
+chrono = { version = "0.4.19", features = ["serde"] }
+console = "0.10.0"
+crc32fast = "1.2.0"
+derive_more = "0.15.0"
+env_logger = "0.7.1"
 failure = "0.1.6"
-thiserror = "1.0.20"
-serde = { version = "1.0.101", features = ["derive", "rc"] }
+flate2 = "1.0.14"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -8,56 +8,56 @@ include = ["src/**/*", "Cargo.toml", "README.md"]
...
-thiserror = "1.0.20"
-serde = { version = "1.0.101", features = ["derive", "rc"] }
-futures = { version = "0.3", features = ["compat"] }
-futures01 = { version = "0.1.29", package = "futures" }
+base64 = "0.11.0"
 bytes = { version = "0.4.12", features = ["serde"] }
-procspawn = { version = "0.9.0", features = ["backtrace", "json"] }
-url = "1.7.2"
+cadence = "0.18.0"
+chrono = { version = "0.4.19", features = ["serde"] }
```
flub, line 19: Only line which has changed here, the others are just moved

src/actors/symbolication.rs
```diff
@@ -1372,7 +1373,8 @@ impl SymbolicationActor {
                             .collect::<CodeModulesBuilder>();
 
                         let minidump_state = MinidumpState {
-                            timestamp: process_state.timestamp(),
+                            timestamp: Utc
+                                .timestamp(process_state.timestamp().try_into().unwrap(), 0),
```
Swatinem, line 1377: is this really never gonna fail?

Swatinem (approved):

Swatinem:
> sorts dependencies in cargo.toml alphabetically.

do you have a script for that? I would like to do that in general, lol.

src/actors/symbolication.rs
flub: well, if we get a minidump that was created at 2**63 seconds since the epoch it'd fail.  Or if a clock was that far off.  The former won't be for our lifetime (something like 292 trillion years in the future if i got this right?).  The bad clock thing is more interesting I guess, I chose panic based on that this is too far in the future to care, not on the bad clock.  And I reckoned panic would be able to be debugged if it does happen.

However this happens in the procspawned part of the code... so maybe we wouldn't notice this :(

flub:
> > sorts dependencies in cargo.toml alphabetically.
> 
> do you have a script for that? I would like to do that in general, lol.

I select the lines and say `M-x sort-lines` in emacs.... does it need to be fancier than that?

PR #287, opened 2020-11-10
ref(stackwalk): simplify MinidumpState creation

flub:
This moves the creation of MinidumpState into a std::convert::From
impl which simplifies the stackwalking code itself by moving boring
stuff out of the way.

It also removes get_object_type_from_minidump() completely forcing the
get_referenced_modules() code to build a full MinidumpState to get
this information.  That's a few wasteful allocations but avoids more
special purpose code matching strings.

----

#skip-changelog

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -98,22 +98,22 @@ type ComputationChannel = Shared<oneshot::Receiver<(Instant, SymbolicationRespon
 
 type ComputationMap = Arc<Mutex<BTreeMap<RequestId, ComputationChannel>>>;
 
-/// A builder for a clean set of code modules.
+/// A builder for the modules list of the symbolication response.
 ///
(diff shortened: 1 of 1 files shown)


Swatinem (approved):

src/actors/symbolication.rs
```diff
@@ -1043,7 +1042,44 @@ struct MinidumpState {
     assertion: String,
 }
 
+impl From<&ProcessState<'_>> for MinidumpState {
```
jan-auer, line 1045: _nit_: This should probably be an explicit `new` function rather than an implicit converter; especially, since it operates on a reference.

flub: Could you elaborate a bit more?

jan-auer: Constructor functions are usually named `new` or `build` rather than via conversion traits. We're computing a state from a process state, not "casting" it.

PR #288, opened 2020-11-11
test: Be more liberal when skipping credentials

flub:
When an external person makes a PR they do not get access to the
credentials in the environment variables and the tests will fail.  So
we now check we're actually running as getsentry to still allow
skipping for external contributors.

This also removes the --ci flag, as we are now depending on Github
Actions environment variables we might as well depend on them to
detect running in CI in general.

#skip-changelog

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -242,9 +242,16 @@ def hitcounter(request):
 
 @pytest.fixture
 def s3(pytestconfig):
+    """AWS S3 credentials for testing S3 buckets.
+
+    This will skip if the secrets are not in the environment, unless we are running on CI in
+    the getsentry team, in which case it will fail.  When running CI not as part of the
+    getsentry team you do not automatically get access to the configured secrets so skipping
+    is allowed.
+    """
     if not AWS_ACCESS_KEY_ID or not AWS_SECRET_ACCESS_KEY:
         msg = "No AWS credentials"
-        if pytestconfig.getoption("ci"):
+        if os.getenv("CI") and os.getenv("GITHUB_REPOSITORY").startswith("getsentry/"):
             pytest.fail(msg)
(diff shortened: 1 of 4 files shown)


flub:
This addresses a problem seen in #182

tests/integration/conftest.py
```diff
@@ -242,9 +242,16 @@ def fail_on_errors():
...
+
+    This will skip if the secrets are not in the environment, unless we are running on CI in
+    the getsentry team, in which case it will fail.  When running CI not as part of the
+    getsentry team you do not automatically get access to the configured secrets so skipping
+    is allowed.
+    """
     if not AWS_ACCESS_KEY_ID or not AWS_SECRET_ACCESS_KEY:
         msg = "No AWS credentials"
-        if pytestconfig.getoption("ci"):
+        if os.getenv("CI") and os.getenv("GITHUB_REPOSITORY").startswith("getsentry/"):
```
Swatinem, line 254: I’m not sure this is correct.
Maybe you were inspired by https://github.com/getsentry/sentry-native/blob/dfc96e3fcbe162b84fd1dd71ca65b6387eac0ca2/.github/workflows/ci.yml#L152 ?

However that doesn’t really work as seen here: https://github.com/getsentry/sentry-native/pull/413/checks?check_run_id=1358094950 :-(

I’m not sure there is a good solution to "is this PR from a fork or internal"?

jan-auer (approved):

tests/integration/conftest.py
flub: arguably it should check for the full `getsentry/symbolicator` in the `GITHUB_REPOSITORY` env var.  But otherwise I think this should be fine?  Ok, some tests are not required for external contributors, but at least our master will break if that was bad.

PR #291, opened 2020-11-12
test: Try skipping on missing credentials for forks again

flub:
The intention is that we allow to skip these tests when they are being
executed from a fork in CI but they still fail when on the credentials
are missing on the main repo.

#skip-changelog

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -240,6 +240,23 @@ def hitcounter(request):
         yield app
 
 
+def fail_missing_secrets(msg):
+    """Fail or skip the test because of missing secrets.
+
+    Secrets are only guaranteed to be available on CI when the branch being tested is
+    committed to the main getsentry repository.  Especially for forks they are hidden to
+    stop external contributors from stealing the secrets.
+
+    This will call `pytest.skip` of the secrets are missing but this is acceptable because
+    e.g. we're on a developer's machine or on a forked repository.  If running on CI in the
+    main repository this will call `pytest.skip`.
+    """
+    if os.getenv("CI") and os.getenv("GITHUB_BASE_REF"):
+        pytest.fail(msg)
+    else:
+        pytest.skip(msg)
+
+
 @pytest.fixture
 def s3(pytestconfig):
     """AWS S3 credentials for testing S3 buckets.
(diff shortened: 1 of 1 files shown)


untitaker (approved):

tests/integration/conftest.py
```diff
@@ -240,6 +240,23 @@ def fail_on_errors():
...
+def fail_missing_secrets(msg):
+    """Fail or skip the test because of missing secrets.
+
+    Secrets are only guaranteed to be available on CI when the branch being tested is
+    committed to the main getsentry repository.  Especially for forks they are hidden to
+    stop external contributors from stealing the secrets.
+
+    This will call `pytest.skip` of the secrets are missing but this is acceptable because
+    e.g. we're on a developer's machine or on a forked repository.  If running on CI in the
+    main repository this will call `pytest.skip`.
```
Swatinem, line 252:
```suggestion
    This will call `pytest.skip` if the secrets are missing but this is acceptable because
    e.g. we're on a developer's machine or on a forked repository.  If running on CI in the
    main repository this will call `pytest.fail`.
```

tests/integration/conftest.py
```diff
@@ -240,6 +240,23 @@ def fail_on_errors():
...
+    """
+    if os.getenv("CI") and os.getenv("GITHUB_BASE_REF"):
```
Swatinem, line 254: according to the docs:

> The base_ref or target branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is a pull_request.

> The head_ref or source branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is a pull_request.

hm, maybe you can check if base_ref and head_ref are from the same repository? Otherwise I think `base_ref` will always be set also for PRs from forks…

Can you maybe just try this by printing out these vars and see what they are?

Swatinem (changes requested):

tests/integration/conftest.py
flub: The docs say (https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables):
> GITHUB_HEAD_REF | Only set for forked repositories. The branch of the head repository.
> GITHUB_BASE_REF | Only set for forked repositories. The branch of the base repository.

So I think just presence is enough.  But ugh, the logic is wrong, if it's present it should skip, if not fail.

If you are not forked I think you only get `GITHUB_REF`.

Swatinem (approved):

PR #292, opened 2020-11-12
ref: Make some of the symbolication fns async

Swatinem:
This does not go as far as #179 but cleans up some things nevertheless

#skip-changelog

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -882,6 +882,11 @@ pub struct SymbolicateStacktraces {
     pub modules: Vec<CompleteObjectInfo>,
 }
 
+/// Run future with a timeout, instrumented with metrics.
+///
+/// This runs the future `future` and exports the duration as the `futures.done` metric
+/// using `task_name` as metric field.  The future will be cancelled if it runs for longer
+/// than `duration` with this status also being represented in the metric.
 async fn measure_task_timeout<T, F>(task_name: &str, future: F, duration: Duration) -> F::Output
 where
     F: Future<Output = Result<T, SymbolicationError>>,
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -15,6 +15,7 @@ use futures01::future::{self, join_all, Future, IntoFuture, Shared};
 use futures01::sync::oneshot;
 use parking_lot::Mutex;
 use regex::Regex;
+use sentry::SentryFutureExt;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use symbolic::common::{
@@ -42,7 +43,7 @@ use crate::types::{
 };
 use crate::utils::futures::{CallOnDrop, ThreadPool};
 use crate::utils::hex::HexValue;
-use crate::utils::sentry::SentryFutureExt;
+use crate::utils::sentry::SentryFutureExt as SentryLegacyFutureExt;
 
 /// Options for demangling all symbols.
 const DEMANGLE_OPTIONS: DemangleOptions = DemangleOptions {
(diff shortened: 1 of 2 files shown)


src/actors/symbolication.rs
```diff
@@ -42,7 +43,7 @@ use crate::types::{
 };
 use crate::utils::futures::{CallOnDrop, ThreadPool};
 use crate::utils::hex::HexValue;
-use crate::utils::sentry::SentryFutureExt;
+use crate::utils::sentry::SentryFutureExt as SentryLegacyFutureExt;
```
jan-auer, line 46: You can just use `as _`.

src/actors/symbolication.rs
```diff
@@ -245,10 +246,10 @@ impl SymbolicationActor {
         }
     }
 
-    fn create_symbolication_request<F, R>(&self, f: F) -> RequestId
+    fn create_symbolication_request<F>(&self, f: F) -> RequestId
     where
-        F: FnOnce() -> R,
-        R: Future<Item = CompletedSymbolicationResponse, Error = SymbolicationError> + 'static,
+        F: std::future::Future<Output = Result<CompletedSymbolicationResponse, SymbolicationError>>
```
jan-auer, line 251: Can you import this as `Future` and import the legacy future as `Future01` or `Future1`?

src/actors/symbolication.rs
```diff
@@ -882,21 +883,19 @@ pub struct SymbolicateStacktraces {
...
-        let result = self
-            .clone()
-            .do_symbolicate_impl(request)
-            .boxed_local()
-            .compat();
+    ) -> Result<CompletedSymbolicationResponse, SymbolicationError> {
+        let result = self.do_symbolicate_impl(request).boxed_local().compat();
 
-        Box::new(future_metrics!(
+        future_metrics!(
```
jan-auer, line 892: Before we continue with this PR, can we first create a version of `future_metrics!`, or ideally a combinator, that works with std future? In the current way, the code is less readable than it was before.

src/actors/symbolication.rs
```diff
@@ -881,22 +882,41 @@ pub struct SymbolicateStacktraces {
     pub modules: Vec<CompleteObjectInfo>,
 }
 
+async fn measure_task_timeout<T, F>(task_name: &str, future: F, duration: Duration) -> F::Output
```
jan-auer, line 885: _nit_: Can be moved into utils if we figure out how to capture timeouts generically. Fine for now.

src/actors/symbolication.rs
```diff
@@ -1495,56 +1508,42 @@ impl SymbolicationActor {
...
+    ) -> Result<(SymbolicateStacktraces, MinidumpState), SymbolicationError> {
+        let future = async move {
+            let sources = Arc::new(sources);
 
-                let cfi_caches = self
-                    .load_cfi_caches(scope.clone(), referenced_modules, sources.clone())
-                    .compat()
-                    .await?;
+            let referenced_modules = self
+                .get_referenced_modules_from_minidump(minidump.clone())
```
jan-auer, line 1535: The only reason we still need these clones is because of the Threadpool requirements internally. I would love to explore if we can get rid of this requirement in a follow-up.

Probably, the proper solution is to get rid of threadpools completely and embrace full asynchronous execution.

jan-auer (approved):
I think this is "good enough". It is a clear improvement over the previous state, but I would like to attempt another cleanup afterwards to see if we further simplify the control flow now, particularly regarding:

- All use of `async` blocks
- Timeouts
- `bind_hub`

src/actors/symbolication.rs
Swatinem: I would like to hear @flub opinion on this. I’m +1 on just using async directly and have the runtime manage things instead of moving things manually to a threadpool. But I’m not deep enough to understand all the tradeoffs here.

flub: huh, while looking at this i (re)realised that we write the entire minidump back to disk just for fun when spawning to procpools.  anyway.

There's is not much gained from `get_referenced_modules_from_minidump` spawning its work onto the symbolication actor's threadpool.  It does have to do a lot of disk IO thanks to the previous paragraph though.  

However in principle we didn't decide to abandon multiple threadpools yet, services will also each control what executor/threadpool they want to run their work on.  Potentially crossing an executor boundary for each service called.  That's perhaps not the best design long-term.  Maybe once we're on a modern tokio and can correctly do disk I/O async the need for this will go away further.

flub (approved):
nice!

src/actors/symbolication.rs
Swatinem: Agreed, yes. I wanted to avoid double-wrapping this for now, but will do that as a followup

PR #293, opened 2020-11-17
fix: Report unused missing unwind info as unused

jan-auer:
We report unwind information as missing, even if it is not required for stack walking on platforms or builds where a frame pointer is available. Since this is not an error and the user should not be alerted to this, we should report as unused, instead.

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -1428,6 +1428,7 @@ impl SymbolicationActor {
                                         // The code module was not in the original set of referenced
                                         // modules, but was now required for stack walking.
                                         None if scanned_modules.contains(&id) => {
+                                            sentry::capture_message("Referenced module not found during initial stack scan", sentry::Level::Error);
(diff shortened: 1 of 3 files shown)


src/actors/symbolication.rs
```diff
@@ -1405,19 +1423,34 @@ impl SymbolicationActor {
...
-                                    Some(id) => {
-                                        unwind_statuses.get(&id).copied().unwrap_or_default()
-                                    }
+                                let unwind_status = match module_id {
+                                    Some(id) => match unwind_statuses.get(&id) {
+                                        // The code module was not in the original set of referenced
+                                        // modules, but was now required for stack walking.
+                                        None if scanned_modules.contains(&id) => {
+                                            ObjectFileStatus::Missing
+                                        }
```
flub, line 1432: I think this would be a bug, could we get a sentry error event for this?  Probably just `log::error!(...)`?

flub (approved):

PR #297, opened 2020-11-20
ref: Updates for Rust 1.48

flub:
This requires a new insta because one of its dependencies needed to be
fixed.  In turn this changes the names of the snapshots being
generated.  The snapshots themselves do not change.

Intra-doc linking is now stable, no need to run rustdoc on nightly
anymore.

#skip-changelog


jan-auer (approved):

flub:
Still doesn't work, but the `Arc<[T]>` are worth reviewing already

src/actors/symbolication.rs
```diff
@@ -1570,7 +1570,7 @@ impl SymbolicationActor {
         sources: Vec<SourceConfig>,
     ) -> Result<(SymbolicateStacktraces, MinidumpState), SymbolicationError> {
         let future = async move {
-            let sources = Arc::new(sources);
+            let sources: Arc<[SourceConfig]> = Arc::from(sources);
```
Swatinem, line 1573: do you really need the type annotation here?

src/config.rs
```diff
@@ -302,7 +302,7 @@ impl Default for Config {
             sentry_dsn: None,
             caches: CacheConfigs::default(),
             symstore_proxy: true,
-            sources: Arc::new(vec![]),
+            sources: Arc::from(vec![]),
```
Swatinem, line 305: even though this does not allocate, I wonder if there is a simpler way?

src/endpoints/applecrashreport.rs
```diff
@@ -117,7 +117,7 @@ fn handle_apple_crash_report_request(
         let response_future = request_future
             .and_then(clone!(symbolication, |mut request| {
                 if request.sources.is_none() {
-                    request.sources = Some((*default_sources).clone());
+                    request.sources = Some(default_sources.to_vec());
```
Swatinem, line 120: before we had a clone here, but I think the `to_vec` re-allocates as well?
Is it possible to use a borrow for `request.sources`? no idea what type that has.

src/main.rs
```diff
@@ -7,6 +7,7 @@
 
 #![warn(missing_docs)]
 #![warn(missing_debug_implementations)]
+#![warn(clippy::rc_buffer)]
```
Swatinem, line 10: why do you need this? since its included in `all`

src/actors/symbolication.rs
flub: No, rust is smart enough. but otherwise the reader has to figure this out somehow by looking up what the type of the parameters are.  I don't mind too much though, I'm pretty sure @jan-auer will be of your opinion as well.

src/endpoints/applecrashreport.rs
flub: Before this was cloning the underlying vector, so actually copying the data.  Now it's also copying this into a new vector.  I actually think this became much more readable.

src/main.rs
flub: maybe we don't.  i'll remove it again

src/actors/symbolication.rs
jan-auer: > I'm pretty sure @jan-auer will be of your opinion as well.

Correct. 

Aside from that, please update tests, as now the types no longer match.

src/endpoints/applecrashreport.rs
jan-auer: It is exactly the same. Cloning a vector or calling `to_vec` on a slice that is backed by a vector is identical.

src/actors/symbolication.rs
```diff
@@ -47,10 +47,7 @@ use crate::utils::hex::HexValue;
 use crate::utils::sentry::SentryFutureExt as _;
 
 /// Options for demangling all symbols.
-const DEMANGLE_OPTIONS: DemangleOptions = DemangleOptions {
-    with_arguments: true,
-    format: DemangleFormat::Short,
-};
+const DEMANGLE_OPTIONS: DemangleOptions = DemangleOptions::complete();
```
Swatinem, line 50: basically short + arguments would become name_only + parameters, but I think @jan-auer has a better idea what we should do here?

flub: Hmm, I went with `DemangleOptions::complete().return_type(false)` which seems to give the desired output because arguments are included.

src/actors/symbolication.rs
flub: lol, it **does** actually need this type.  sorry folks

flub: > Aside from that, please update tests, as now the types no longer match.

Not sure what you refer to with this comment.

Swatinem (approved):

src/actors/symbolication.rs
jan-auer: You resolved this in the meanwhile.

src/config.rs
jan-auer: Probably `Arc::from(&[])`

jan-auer (approved):

src/config.rs
flub: Nope, you need to get a slice but getting a slice from an array is unstable.

PR #298, opened 2020-11-20
ci: Run tests with rust nightly and beta channels on a schedule

flub:
This runs only the unittests once a day using the nightly and the beta
channel.  This should not block PRs but will give us earlier warnings
when we need to adjust things.

No caching since the compiler will be different every day anyway and
it is not blocking PRs anyway.

#skip-changelog


Swatinem (approved):
can we run clippy as well? since they introduce new breaking lints quite frequently

flub:
Yeah, i wondered if this would be too noisy.  But i'm also just fixing clippy stuff in symbolicator... and since we run normal ci on latest stable this blocks new PRs as soon as a new rust is released. so let's add it and try out i guess

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,25 @@
+name: Rust nightly CI
```
jan-auer, line 1: Aligning with other task names. This is evidently Rust.

```suggestion
name: Nightly CI
```

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,25 @@
...
+    - cron: '11 7 * * *'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust: [nightly, beta]
+    env:
+      RUSTFLAGS: -Dwarnings
```
jan-auer, line 14: Please put this on top-level, like in `ci.yml`.

jan-auer (changes requested):
I would be ok with running clippy since usually they only make lints required for features that have been released for 1-2 versions. I might be mistaken though, in which case clippy would just be throwing us off.

flub:
CI failures here are fixed by #297 btw, not because this PR is bad.

jan-auer (approved):
Should be good after merging master.

PR #299, opened 2020-11-23
ref(minidump): Consolidate building of module info

flub:
The minidump stack unwiding starts the building of the module info
which is eventually sent back in the symbolication response.  This
consolidates this code by introducing a few more helpers making the
main stackwalking code smaller and separating concerns a bit better.

This makes one change:

- The unwind_status field is now marked as "unused" even when CFI for
  the module was found, before it was only marked as unsued if the
  CFI was missing.  This behaviour is more correct, the features map
  still indicates the presence of the CFI for the module.

#skip-changelog

No snapshot updates required in sentry

### DIFF
--- a/src/services/snapshots/symbolicator__services__symbolication__tests__minidump_windows.snap
+++ b/src/services/snapshots/symbolicator__services__symbolication__tests__minidump_windows.snap
@@ -154,7 +154,7 @@ stacktraces:
         trust: context
 modules:
   - debug_status: found
-    unwind_status: unused
+    unwind_status: found
     features:
       has_debug_info: true
       has_unwind_info: true
--- a/src/services/snapshots/symbolicator__services__symbolication__tests__minidump_windows.snap
+++ b/src/services/snapshots/symbolicator__services__symbolication__tests__minidump_windows.snap
@@ -154,7 +154,7 @@ stacktraces:
         trust: context
 modules:
   - debug_status: found
-    unwind_status: found
+    unwind_status: unused
     features:
       has_debug_info: true
       has_unwind_info: true
--- a/src/services/symbolication.rs
+++ b/src/services/symbolication.rs
@@ -145,7 +145,7 @@ struct CfiCacheModules {
 }
 
 impl CfiCacheModules {
-    /// Creates a new
+    /// Creates a new CFI cache entries for code modules.
     fn new<'a>(cfi_caches: impl Iterator<Item = &'a CfiCacheResult>) -> Self {
         let inner = cfi_caches
             .map(|(code_id, cache_result)| {
@@ -169,17 +169,14 @@ impl CfiCacheModules {
                             cfi_status,
                             cfi_path,
                             cfi_candidates: cfi_cache.candidates().clone(), // TODO(flub): fix clone
-                            scanned: false,
+                            ..Default::default()
                         }
                     }
                     Err(err) => {
                         log::debug!("Error while fetching cficache: {}", LogError(err.as_ref()));
                         CfiModule {
-                            features: Default::default(),
                             cfi_status: ObjectFileStatus::from(err.as_ref()),
-                            cfi_path: None,
-                            cfi_candidates: AllObjectCandidates::default(),
-                            scanned: false,
+                            ..Default::default()
                         }
                     }
                 };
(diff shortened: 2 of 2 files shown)


src/actors/symbolication.rs
```diff
@@ -97,37 +98,217 @@ type ComputationChannel = Shared<oneshot::Receiver<(Instant, SymbolicationRespon
...
+    }
+
+    /// Marks a module as scanned for CFI.
+    ///
+    /// If during stack unwinding the module was scanned that means we needed its CFI.  We
+    /// need to keep track of this because it may indicate we didn't fetch a CFI file we
+    /// needed, or we thought a CFI module was missing but it wasn't needed.  So this also
+    /// updates the [`CfiModule::cfi_status`] field.
+    fn mark_scanned(&mut self, code_id: CodeModuleId) {
+        match self.inner.get_mut(&code_id) {
```
Swatinem, line 196: does `BTreeMap` have the same `entry` api as hashmaps? Maybe `entry().or_insert_with()` may be simpler?

src/actors/symbolication.rs
```diff
@@ -97,37 +98,217 @@ type ComputationChannel = Shared<oneshot::Receiver<(Instant, SymbolicationRespon
...
+/// A module which was referenced in a minidump and processing information for it.
+#[derive(Debug, Serialize, Deserialize, Default)]
+struct CfiModule {
+    /// Combined features provided by all the DIFs we found for this module.
+    features: ObjectFeatures,
+    /// Status of the CFI or unwind information for this module.
+    cfi_status: ObjectFileStatus,
+    /// Path to the CFI file in our cache, if there was a cache.
+    cfi_path: Option<PathBuf>,
+    ///
```
Swatinem, line 241:
```suggestion
    /// Indicates if the module was user for stack scanning.
```

src/actors/symbolication.rs
```diff
@@ -97,37 +98,217 @@ type ComputationChannel = Shared<oneshot::Receiver<(Instant, SymbolicationRespon
...
+    scanned: bool,
+}
+
+/// Builder object to collect modules from the minidump.
+///
+/// This collects the modules found by the minidump processor and constructs the
+/// [`CompleteObjectInfo`] objects from them, which are used to eventually return the
+/// module list in the [`SymbolicateStacktraces`] object of the final JSON API response.
+///
+/// The builder requires marking modules are referenced by stack frames.  This allows it to
```
Swatinem, line 251:
```suggestion
/// The builder requires marking modules that are referenced by stack frames.  This allows it to
```

src/actors/symbolication.rs
flub: Yes it does.  I thought I didn't like it but tried that again and it is indeed nicer.

Swatinem (approved):

src/actors/symbolication.rs
```diff
@@ -97,37 +98,211 @@ type ComputationChannel = Shared<oneshot::Receiver<(Instant, SymbolicationRespon
...
+        minidump_process_state: &ProcessState,
+        executable_type: ObjectType,
+    ) -> Self {
+        // Firstly mark modules for which we needed CFI information during unwinding.  This
+        // will insert entries into cfi_cache for modules which were missing.
+        for thread in minidump_process_state.threads() {
+            for frame_pair in thread.frames().windows(2) {
+                if let [prev_frame, next_frame] = frame_pair {
+                    if let Some(code_id) = prev_frame.module().and_then(|m| m.id()) {
+                        if matches!(next_frame.trust(), FrameTrust::Scan | FrameTrust::CFIScan) {
```
flub, line 265: @jan-auer I think your original code forgot to add `FrameTrust::CFI` to this match but this wasn't visible because you only used this information for missing frames instead of for all.  But we want to mark all modules which either used or should have used CFI.  (perhaps the method should be renamed to `mark_cfi_required()` instead of the current `mark_scanned`)

flub: https://github.com/getsentry/breakpad/blob/2aaba88045393a024b0e8d285fd2a416c4e05ff4/src/processor/stackwalker_x86.cc#L138 is where these values get assigned btw.

jan-auer: No, the intention is to look for the inverse: missing CFI. We detect this by searching for scan trusts.

src/actors/snapshots/symbolicator__actors__symbolication__tests__minidump_windows.snap
```diff
@@ -154,7 +154,7 @@ stacktraces:
         trust: context
 modules:
   - debug_status: found
-    unwind_status: found
+    unwind_status: unused
```
flub, line 157: So perhaps this is wrong.  Line 42-52 of this snapshot shows a frame with `trust: cfi` so this should have been `found`.  @jan-auer could you confirm this?

jan-auer: Confirmed.

flub:
re-based on master, please re-review :)

Swatinem (approved):
the problem with rebase workflows is that they are crappy in github, or in git in general.

Anyhow, I just superficially scanned through this, because I have no idea what actually changed to the last time I reviewed this :-D

PR #309, opened 2020-11-27
feat(difs): Add dif candidates info to module list

flub:
This adds new information about which DIF candidates were considered
to be used for modules.  So far only part of all the information we
want to expose is present.

sentry snapshot updates: https://github.com/getsentry/sentry/pull/22395


src/actors/cficaches.rs
```diff
@@ -220,6 +220,7 @@ impl CfiCacheActor {
 
         object.and_then(move |object| {
             object
+                .meta
```
jan-auer, line 223: _nit_: Also rename the variable in the closure below to `object_meta`.

src/actors/objects/mod.rs
```diff
@@ -168,7 +168,7 @@ impl CacheItemRequest for FetchFileMetaRequest {
             .data_cache
             .compute_memoized(FetchFileDataRequest(self.clone()))
             .map_err(ObjectError::Caching)
-            .and_then(move |data| {
+            .and_then(move |data: Arc<ObjectFile>| {
```
jan-auer, line 171: `ObjectFile` is probably a bad name, since it's rather a `CachedObject` or `ObjectCacheEntry`, but let's at least rename the variable to make it more clear. Then, you can also remove the type annotation again.

src/actors/objects/mod.rs
```diff
@@ -550,7 +576,11 @@ impl ObjectsActor {
                             // the search by debug/code id. We do not surface those errors to the
                             // user (instead we default to an empty search result) and only report
                             // them internally.
-                            log::error!("Failed to download from {}: {}", type_name, LogError(&e));
+                            log::error!(
+                                "Source list-files failed for {}: {}",
```
jan-auer, line 580: _nit_: To be more consistent with other error messages, how about:

```suggestion
                                "Failed to fetch file list from {}: {}",
```

src/actors/objects/mod.rs
```diff
@@ -593,57 +623,86 @@ impl ObjectsActor {
...
-                    Ok(object_meta) => {
-                        if object_meta.status == CacheStatus::Positive {
-                            // object_meta.features is meaningless when CacheStatus != Positive
-                            match purpose {
-                                ObjectPurpose::Unwind => object_meta.features.has_unwind_info,
-                                ObjectPurpose::Debug => {
-                                    object_meta.features.has_debug_info
-                                        || object_meta.features.has_symbols
+        file_metas.and_then(
+            move |responses: Vec<Result<Arc<ObjectFileMeta>, _>>| -> Result<FoundObject, _> {
```
jan-auer, line 627: Is there a way we can get rid of those type annotations? Again, here we can also work with better variable naming, like `object_metas`.

src/actors/objects/mod.rs
```diff
@@ -593,57 +623,86 @@ impl ObjectsActor {
...
+                                DifStatus::Ok
+                            } else {
+                                DifStatus::NotFound
+                            };
+                            candidates.push(DifInfo {
+                                source_id: obj_meta.request.file_id.source_id().clone(),
+                                source_location: obj_meta.request.file_id.location(),
+                                status,
+                                features: obj_meta.features(),
+                            })
```
jan-auer, line 654:
```suggestion
                            });
```

src/types.rs
```diff
@@ -434,9 +435,56 @@ impl ObjectFeatures {
...
+/// Information about a Debug Information File in the [`CompleteObjectInfo`].
+///
+/// All DIFs are backed by an [`ObjectFile`](crate::actors::objects::ObjectFile).  But we
+/// may not have been able to get hold of this object file.  We still want to describe the
+/// relevant DIF however.
+///
+/// Currently has no [`ObjectId`] attached and the parent container is expected to know
+/// which ID this DIF info was for.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct DifInfo {
```
jan-auer, line 447: Can we name this `*Candidate`, and also rename the field to `CompleteObjectInfo::candidates`? 

Also, the other types are usually named around "objects" and not "DIFs". Should we consolidate this one, too?

src/types.rs
```diff
@@ -434,9 +435,56 @@ impl ObjectFeatures {
...
+    /// The location of this DIF on the object source.
+    ///
+    /// This will generally be some sort of path name or key, if you know the type of object
+    /// source this may give you an idea of where this DIF was expected to be found.
+    pub source_location: SourceLocation,
+    /// The status of this DIF.
+    pub status: DifStatus,
+    /// The features provided by this DIF.
+    // TODO(flub): document better + hide when we don't know.
+    pub features: ObjectFeatures,
```
jan-auer, line 464: Let's give this a `#[serde(skip_serializing_if = "...")]` to ensure we write smaller responses.

src/types.rs
```diff
@@ -434,9 +435,56 @@ impl ObjectFeatures {
...
+    ///
+    /// Generally this is a short readable string.
+    pub source_id: SourceId,
+    /// The location of this DIF on the object source.
+    ///
+    /// This will generally be some sort of path name or key, if you know the type of object
+    /// source this may give you an idea of where this DIF was expected to be found.
+    pub source_location: SourceLocation,
+    /// The status of this DIF.
+    pub status: DifStatus,
```
jan-auer, line 461: _For later_: I'm starting to question this status, but I think this should clear up in a follow-up. At the moment, it seems that it would contain a mushed together version of resolution, downloading, CFI conversion and symcache conversion.

jan-auer (commented):
Let's please put all the candidate information behind a config flag. That is, we can compute all the information internally, but only serialize it when the flag is given. This allows us to iterate on this feature and make releases without impacting any users -- including Sentry.

src/actors/objects/mod.rs
flub: +1 on renaming `ObjectFile`, maybe to `ObjectHandle`.  -0 on removing the type annotation.  It's super handy to be able to easily jump to the type declaration of this variable.  In functions it's normal to have explicit type declarations, a closure of more than a few lines is basically the same as a function.  Anyway, that's all my reasoning for this.

src/actors/objects/mod.rs
flub: `object_meta` is already used as a variable in this code where applicable.  I think `responses` is probably fine-ish.

src/types.rs
```diff
@@ -455,6 +579,17 @@ pub struct CompleteObjectInfo {
...
+    /// More information about the DIF files which were consulted for this object file.
+    ///
+    /// For stackwalking and symbolication we need various Debug Information Files about
+    /// this module.  We look for these DIF files in various locations, this describes all
+    /// the DIF files we looked up and what we know about them, how we used them.  It can be
+    /// helpful to understand what information was available or missing and for which
+    /// reasons.
+    ///
+    /// This list is not serialised if it is empty.
+    pub candidates: Vec<ObjectCandidate>,
```
flub, line 592: I tried adding `#[serde(skip_serializing_if = "Vec::is_empty")]` here but that failed in very strange ways where it would never terminate, I stopped trying to debug it but it might still be worth to re-visit this.

flub:
This PR only covers the download info, the debug and unwind info will follow in followup PRs.

src/actors/symbolication.rs
```diff
@@ -671,6 +672,10 @@ impl SymCacheLookup {
                 if let Some(ref symcache) = symcache {
                     entry.object_info.arch = symcache.arch();
                     entry.object_info.features.merge(symcache.features());
+
+                    let mut difs = Vec::new();
+                    difs.extend_from_slice(&symcache.candidates());
+                    entry.object_info.candidates = difs; // TODO(flub): merge!
```
Swatinem, line 678: Not sure what this comment is about, but can’t you just `extend_from_slice` on the `object_info.candidates`?

src/actors/symcaches.rs
```diff
@@ -222,35 +230,38 @@ impl SymCacheActor {
...
-                .map(move |object| {
-                    Either::A(symcaches.compute_memoized(FetchSymCacheInternal {
-                        request,
-                        objects_actor: objects,
-                        object_meta: object,
-                        threadpool,
+        find_result_future.and_then(move |find_result: FoundObject| {
+            let FoundObject { meta, candidates } = find_result;
+            let candidates: Arc<[ObjectCandidate]> = Arc::from(candidates);
+            meta.map(clone!(candidates, |object_meta| {
```
Swatinem, line 236: hm, is the whole reason `candidates` is wrapped in an `Arc` this `clone!` here? I wonder if its worth converting to `async fn` to get rid of it?

src/types.rs
```diff
@@ -434,9 +435,129 @@ impl ObjectFeatures {
...
+    /// The DIF object was successfully used to provide the required information.
+    ///
+    /// This means the object was used for CFI when used for [`ObjectCandidate::unwind`]
+    Ok,
+    /// The DIF object contained malformed data which could not be used.
+    Malformed,
+    /// An error occurred when attempting to use this DIF object.
+    ///
+    /// This is mostly an internal error from symbolicator which is considered transient.
+    /// The next attempt to access this DIF object will retry the using this DIF object.
```
Swatinem, line 532:
```suggestion
    /// The next attempt to access this DIF object will retry using this DIF object.
```

src/actors/symbolication.rs
flub: That comment is for future work, when the info is also added for CFI objects as well we have to be more careful rather than just extend a vec.  This already improves in the followup PR (#316)

src/actors/symcaches.rs
flub: Kind off, but it doesn't really matter much.  This changes again in (#316) where we get a newtype for this is fine for now.

Swatinem (approved):
lets do it. Since you will solve some of the things I mentioned in the followup

jan-auer (approved):

flub:
@Swatinem @jan-auer PTAL, I realised it didn't include a config setting to disable it yet.  Last commit adds this.  No other changes.

src/actors/symbolication.rs
```diff
@@ -1029,12 +1032,20 @@ impl SymbolicationActor {
...
     ) -> Result<CompletedSymbolicationResponse, SymbolicationError> {
+        let serialize_dif_candidates = self.serialize_dif_candidates;
+
         measure_task_timeout(
             "symbolicate",
             self.do_symbolicate_impl(request),
             Duration::from_secs(3600),
         )
         .await
+        .map(|mut response| {
```
flub, line 1043: nitpicking question, is `move |mut response|` any better for something?  I'm guessing the compiler sees through this completely and would do exactly the same for this trivial case.

jan-auer (approved):

src/actors/symbolication.rs
jan-auer: In this case it would not, since you use a copy type by value, so I think it applies the move implicitly.

src/config.rs
```diff
@@ -242,6 +242,14 @@ pub struct Config {
...
     /// Number of subprocesses in the internal processing pool.
     pub processing_pool_size: usize,
+
+    /// Disables DIF object candidates being returned for symbolication responses.
+    ///
+    /// By default symbolication responses will include diagnostics information about which
+    /// DIF object files were considered for which purposes for each module in the modules
+    /// list.  As this can get rather verbose this option acts as a feature flag to be able
+    /// to disable this.
+    pub serialize_dif_candidates: bool,
```
jan-auer, line 252: I'm thinking that subsequently we may want to make this a request flag instead.

jan-auer (changes requested):
After discussing offline, two final change requests:

- Please move the flag to the request struct (for all request types)
- Please default it to off

src/actors/symbolication.rs
Swatinem: I think its more about using `self.serialize_dif_candidates` and moving self in there? unless `do_symbolicate_impl` takes ownership?

flub:
Apologies for the force-pushing.  I hard resetted the last config commit without realising that's probably not the best way for the PR.  Anyway, the DIF info is now only enabled if in the request as can be seen in the integration tests.

This also means the matching PR for the sentry snapshot tests doesn't need to happen as those responses should not change at all.

untitaker (approved):

src/utils/multipart.rs
```diff
@@ -46,3 +47,11 @@ pub fn read_multipart_sources(
             .and_then(|data| Ok(serde_json::from_slice(&data)?)),
     )
 }
+
+pub fn read_multipart_request_data(
```
untitaker, line 51: This + read_multipart_sources can be one generic fn

src/endpoints/symbolicate.rs
```diff
@@ -44,6 +46,8 @@ struct SymbolicationRequestBody {
     pub stacktraces: Vec<RawStacktrace>,
     #[serde(default)]
     pub modules: Vec<RawObjectInfo>,
+    #[serde(default)]
+    pub options: Option<RequestOptions>,
```
jan-auer, line 50: You can make this `RequestOptions` right away, since serde will then add the default for you.

jan-auer (approved):

src/utils/multipart.rs
flub: Yes, that would seem sane but it appears I'm not smart enough to understand serde's lifetimes.  I'm just living with this for now instead of wasting loads of time.  I'd appreciate any hints though.

jan-auer (approved):
Now for real.

PR #310, opened 2020-11-28
meta: Add a squared icon

jan-auer:
Adds a square crop of the full symbolicator decoder ring as icon.

icon

---

#skip-changelog


Swatinem (approved):
🎉

PR #312, opened 2020-11-30
style(clippy): Unused result and contains

flub:
Some nightly clippy warnings that seem sensible.

This also gives a unique name for each clippy run, as this is used for
some caching and is probably the reason we don't get nice clippy
results currently.

----

#skip-changelog

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -11,6 +11,7 @@ jobs:
   test:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         rust: [nightly, beta]
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           override: true
           components: clippy
 
-      - name: Run clipppy
+      - name: Run clipppy ${{ matrix.rust }}
```
jan-auer, line 26:
```suggestion
      - name: Run clippy ${{ matrix.rust }}
```

jan-auer (approved):
Let's check if clippy lints are directly promoted from nightly to stable. Otherwise, running beta only would be enough, and it would be more reliable than nightly.

PR #313, opened 2020-12-01
ref(symbolication): Remove stale error kind

jan-auer:
The error kind has only been used in tests and did not convey any semantic meaning. By switching to `anyhow::Result` for tests, we can avoid it alltogether.

#skip-changelog

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -78,7 +78,6 @@ impl From<&SymbolicationError> for SymbolicationResponse {
     fn from(err: &SymbolicationError) -> SymbolicationResponse {
         match err {
             SymbolicationError::Timeout => SymbolicationResponse::Timeout,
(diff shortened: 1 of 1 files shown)


src/actors/symbolication.rs
```diff
@@ -81,7 +78,7 @@ impl From<&SymbolicationError> for SymbolicationResponse {
     fn from(err: &SymbolicationError) -> SymbolicationResponse {
         match err {
             SymbolicationError::Timeout => SymbolicationResponse::Timeout,
-            SymbolicationError::Io(_) => SymbolicationResponse::InternalError,
+            // SymbolicationError::Io(_) => SymbolicationResponse::InternalError,
```
Swatinem, line 81:
```suggestion
```

Swatinem (approved):

flub (approved):

PR #316, opened 2020-12-03
feat(difs): Add debug status to dif candidates info

flub:


src/types/mod.rs
```diff
@@ -540,17 +547,15 @@ pub enum ObjectUseInfo {
...
-impl ObjectUseInfo {
-    pub fn is_none(&self) -> bool {
-        matches!(self, Self::None)
-    }
-}
+/// Newtype around a collection of [`ObjectCandidate`] structs.
+///
+/// This abstracts away some common operations needed on this collection.  Because it needs
+/// to be passed through objects implementing the [`CacheItemRequest`] trait while avoiding
+/// copying and yet being able to mutate, we need to have inner mutability.
```
Swatinem, line 554: this comment does not reflect reality, since this will be cloned, and it has no internal mutability

src/types/objects.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+impl ObjectUseInfo {
+    pub fn is_none(&self) -> bool {
+        matches!(self, Self::None)
+    }
+}
+
+impl From<Vec<ObjectCandidate>> for AllObjectCandidates {
```
Swatinem, line 21: Maybe `IntoIterator<ObjectCandidate>` would be a better bound?

src/types/objects.rs
```diff
@@ -0,0 +1,85 @@
...
+    fn from(mut source: Vec<ObjectCandidate>) -> Self {
+        source.sort_by_key(|candidate| (candidate.source.clone(), candidate.location.clone()));
```
Swatinem, line 23: The docs say:

> For expensive key functions (e.g. functions that are not simple property accesses or basic operations), sort_by_cached_key is likely to be significantly faster, as it does not recompute element keys.

Since at least `location.clone()` allocates a new string, you should try to optimize this.

src/types/objects.rs
```diff
@@ -0,0 +1,85 @@
...
+impl AllObjectCandidates {
+    /// Sets the [`ObjectCandidate::debug`] field for the specified DIF object.
+    ///
+    /// You can only request symcaches from a DIF object that was already in the metadata
+    /// candidate list, therefore if the candidate is missing it is treated as an error.
+    pub fn set_debug(&mut self, source: SourceId, location: SourceLocation, info: ObjectUseInfo) {
+        match self
+            .0
+            .binary_search_by_key(&(source, location), |candidate| {
+                (candidate.source.clone(), candidate.location.clone())
```
Swatinem, line 37: I think the same applies here. Maybe factoring out the sorting function would be a good idea, as that would also format this a lot nicer.

src/types/objects.rs
```diff
@@ -0,0 +1,85 @@
...
+            }) {
+            Ok(index) => {
+                if let Some(mut candidate) = self.0.get_mut(index) {
+                    candidate.debug = info;
+                } else {
+                    // TODO(flub): remove before PR
+                    panic!("wtf");
```
Swatinem, line 44: 🤣

Swatinem:
just curious, since you have so many snapshot changes: is the sorting of the candidates stable? should it be? or just for the tests at least?

src/types/mod.rs
flub: Huh, yes.  the implementation changed a bit since this was written :/
thanks for the catch

src/types/objects.rs
flub: This way this can move the Vec rather than re-create it.  And this happens to already be created as a Vec, so I don't think there's anything to be gained from being more generic here.

Maybe this should be a `::new` method intead?  I'm not really sure when new is preferred.

src/types/objects.rs
flub: Oh yes, didn't spot the cached version.

This cloning is annoying, I didn't know what to do better about this.

src/types/objects.rs
flub: Not sure what your suggestion is here.  There is no cached variant, which is probably because this never needs to compute the same key twice.  As this is currently only in the `From` and here I don't think it's worth further factoring out.  Once this grows more methods, like `.set_unwind()` that'll probably change though at which point I think it makes more sense to do that.

src/types/objects.rs
flub: lol, the type system made me do this.  At least I never actually saw that panic so all is good :)

flub:
> just curious, since you have so many snapshot changes: is the sorting of the candidates stable? should it be? or just for the tests at least?

Since this commit it explicitly is, since it now sorts on (sourceid, location).  So this is hopefully the last time this order is completely changed in all the snapshot changes.

src/types/objects.rs
Swatinem: hm, why can’t you just use `(&candidate.source, &candidate.location)`? oh yes, because sort is `mut`, or does it basically take care of that internally?

flub: I think it's the tuple that makes this impossible.

flub:
So this went through a bit of a git hell after the target branch got merged into master.   The diff is now correct again and is reviewable.  Please do so. 👼

Swatinem (approved):
Would be nice to find a solution to the excessive cloning, but I think this is fine for a first pass

PR #317, opened 2020-12-09
ci(clippy): Enable clippy::all all in source code

flub:
This makes sure everyone invokes clippy locally the same way as on CI
and doesn't get any surprises.

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
         uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-features --workspace --tests --examples -- -D clippy::all
+          args: --all-features --workspace --tests --examples
 
   unit-test:
     name: Unit Tests
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -28,7 +28,7 @@ jobs:
         uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-features --workspace --tests --examples -- -D clippy::all
+          args: --all-features --workspace --tests --examples
 
       - uses: actions-rs/cargo@v1
         with:
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ lint-python: .venv/bin/python
 
 lint-rust:
 	@rustup component add clippy --toolchain stable 2> /dev/null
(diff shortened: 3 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --all-features --workspace --tests --examples -- -D clippy::all
+          args: --all-features --workspace --tests --examples
```
jan-auer, line 57: You should still leave `-D` here, as otherwise the build won't fail with clippy lint violations.

jan-auer (changes requested):
I don't necessarily mind adding `warn(clippy::all)` in Symbolicator, but there is a number of reasons why we chose the current approach across all Sentry projects:

- We mostly use Rust Analyzer, which allows to set `clippy` as the check command. This is also documented in our internal Rust documentation.
- In multi-workspace projects, this has to be repeated in every crate of the workspace.
- We want to only warn during development, but deny in CI

.github/workflows/ci.yml
flub: isn't the `env: RUSTFLAGS: -Dwarnings` taking care of that?  main.rs turns them into warnings, the env flag turns them into errors for CI.  That's what I'm expecting to happen at least.

jan-auer: Ah, you're right. I completely forgot we had that, too! I'm still slightly in favor of having it in a central place for workspaces (which symbolicator will soon become), and since this is accepted practice in all our Rust repositories. 

If you still prefer to go ahead with this, please also add `warn(clippy::all)` to the other two crates in this repository, and remove from [here](https://github.com/getsentry/symbolicator/blob/b65856cde87b6e1de6472dec485c3037d0f8fd25/Makefile#L89).

flub: > (which symbolicator will soon become)

Noooo
 
> and since this is accepted practice in all our Rust repositories.

We already document you should add `warn(missingdocs, missing_debug_implementations)` to the crates.  I'd suggest we add `clippy::all` to that.

> If you still prefer to go ahead with this, please also add `warn(clippy::all)` to the other two crates in this repository, and remove from [here](https://github.com/getsentry/symbolicator/blob/b65856cde87b6e1de6472dec485c3037d0f8fd25/Makefile#L89).

yes, and done

jan-auer (approved):
`symsorter` still needs the new annotation, then this is G2G.

PR #322, opened 2020-12-17
ref(objects): Make objects actor async/await

flub:
I tried to update the objects actor to handle the new download meta information and I ended up with nested `future::Either` structs and it was generally really hard to modify.  The desired modifications didn't end up working and we had another chat about approaches to this and some new plans.  But in the mean time this refactor makes working on this code much easier.  So submitting this as a separate refactor-only PR.  This does:

- Split the objects actor into multiple modules, with the meta cache and data cache each having their own module.
- Rename the `ObjectMetaFile` into `ObjectMetaHandle` to match the already existing `ObjectHandle`.
- Make the objects actor use futures 0.3
- Convert the objects actor to async/await

Changes make sense in individual commits, so per-commit reviewing should work pretty well.

#skip-changelog


untitaker (approved):

src/actors/cficaches.rs
```diff
@@ -201,6 +202,7 @@ impl CfiCacheActor {
     ) -> impl Future<Item = Arc<CfiCacheFile>, Error = Arc<CfiCacheError>> {
         let object = self
             .objects
+            .clone()
```
Swatinem, line 205: I hope this is fast to clone…

src/actors/objects/mod.rs
```diff
@@ -596,242 +233,235 @@ impl ObjectsActor {
...
+    ///
+    /// TODO(flub): Once all the callers are async/await this should take `&self` again
+    /// instead of requiring `self`.
+    ///
+    // TODO(flub): this function should inline already fetch the meta (what
+    // `fetch_file_metas()` does now) for each file.  Currently we synchronise all the
+    // futures at the end of the listing for no good reason.
+    async fn list_files(
+        self,
+        sources: Arc<[SourceConfig]>,
```
Swatinem, line 259: I guess with async/await, this can also be `&[SourceConfig]` then.

src/actors/objects/mod.rs
```diff
@@ -596,242 +233,235 @@ impl ObjectsActor {
...
-                        };
-                        (score, *i)
-                    })
-                    .map(|(_i, response)| response)
-                    .transpose()
-                    .map(|meta| FoundObject { meta, candidates: candidates.into() })
-            },
-        )
+            };
+            queries.push(query);
```
Swatinem, line 347: you could keep the `file_ids.into_iter().map()`. Also, maybe in the future, `file_ids` could be a slice or an `&dyn Iterator` / `impl Iterator` instead of a `Vec`

Swatinem (approved):

src/actors/cficaches.rs
flub: It's 4 `Arc`s so not too bad.  Also, if you looked into this this was already being cloned many times (still is) so this isn't that bad of a *change* compared to before.

src/actors/objects/mod.rs
flub: I'd love to pair on this with you :) I'm also unsure why I failed to make this take `&self`

src/actors/objects/mod.rs
flub: I think in the future this should become `fetch_file_meta()` which operates on just one file_id so that it can be chained on the future which creates the file_id and the join only happens at the end of this.  But I didn't want to change the logic for now and left this as a more pure async/await refactor.

PR #323, opened 2020-12-21
ref(cache): Convert to futures 0.3

flub:
This refactors the cache actor to futures 0.3, as this involves a
trait this requires touching everything using this trait.  No attempt
at changing the code-style itself or moving to async/await has been
made, now that everything is on futures 0.3 this can be done in
smaller parts where more careful thought can be put into this.
Instead changes have been tried to keep to just the minimum required
for the futures conversion, only laying the groundwork for more
improvements later.

The trait itself has been changed to return Pin<Box<dyn Future ...>>,
since you can not do async/await in traits this matches the signature
that the async-trait crate generates.  This allows seamless
integration in future async/await code, without the Pin currently a
few Unpin requirements would already break.

#skip-changelog

### DIFF
--- a/src/actors/cficaches.rs
+++ b/src/actors/cficaches.rs
@@ -1,13 +1,12 @@
 use std::fs::File;
 use std::io::{self, BufWriter};
 use std::path::Path;
-use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
 
 use futures::compat::Future01CompatExt;
 use futures::future::{self, Either};
-use futures::{Future, FutureExt, TryFutureExt};
+use futures::{FutureExt, TryFutureExt};
 use sentry::{configure_scope, Hub, SentryFutureExt};
 use symbolic::{
     common::ByteView,
@@ -15,7 +14,7 @@ use symbolic::{
 };
 use thiserror::Error;
 
-use crate::actors::common::cache::{CacheItemRequest, CachePath, Cacher};
+use crate::actors::common::cache::{BoxedFuture, CacheItemRequest, CachePath, Cacher};
 use crate::actors::objects::{
     FindObject, ObjectError, ObjectHandle, ObjectMetaHandle, ObjectPurpose, ObjectsActor,
 };
@@ -112,10 +111,7 @@ impl CacheItemRequest for FetchCfiCacheInternal {
     ///
     /// The extracted CFI is written to `path` in symbolic's
     /// [`CfiCache`](symbolic::minidump::cfi::CfiCache) format.
-    fn compute(
-        &self,
-        path: &Path,
-    ) -> Pin<Box<dyn Future<Output = Result<CacheStatus, Self::Error>>>> {
+    fn compute(&self, path: &Path) -> BoxedFuture<Result<CacheStatus, Self::Error>> {
         let path = path.to_owned();
         let object = self
             .objects_actor
@@ -203,7 +199,7 @@ impl CfiCacheActor {
     pub fn fetch(
         &self,
         request: FetchCfiCache,
-    ) -> Pin<Box<dyn Future<Output = Result<Arc<CfiCacheFile>, Arc<CfiCacheError>>>>> {
+    ) -> BoxedFuture<Result<Arc<CfiCacheFile>, Arc<CfiCacheError>>> {
         let object = self
             .objects
             .clone()
--- a/src/actors/cficaches.rs
+++ b/src/actors/cficaches.rs
@@ -14,14 +14,14 @@ use symbolic::{
 };
 use thiserror::Error;
 
-use crate::actors::common::cache::{BoxedFuture, CacheItemRequest, CachePath, Cacher};
+use crate::actors::common::cache::{CacheItemRequest, CachePath, Cacher};
 use crate::actors::objects::{
     FindObject, ObjectError, ObjectHandle, ObjectMetaHandle, ObjectPurpose, ObjectsActor,
 };
 use crate::cache::{Cache, CacheKey, CacheStatus};
 use crate::sources::{FileType, SourceConfig};
 use crate::types::{ObjectFeatures, ObjectId, ObjectType, Scope};
-use crate::utils::futures::ThreadPool;
+use crate::utils::futures::{BoxedFuture, ThreadPool};
 use crate::utils::sentry::WriteSentryScope;
 
 /// Errors happening while generating a cficache
--- a/src/actors/cficaches.rs
+++ b/src/actors/cficaches.rs
@@ -8,7 +8,6 @@ use std::time::Duration;
 use futures::compat::Future01CompatExt;
 use futures::future::{self, Either};
 use futures::{Future, FutureExt, TryFutureExt};
-// use futures01::future::{Either, Future, IntoFuture};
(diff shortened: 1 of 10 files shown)


src/actors/cficaches.rs
```diff
@@ -1,12 +1,15 @@
...
 use std::sync::Arc;
 use std::time::Duration;
 
-use futures::{compat::Future01CompatExt, FutureExt, TryFutureExt};
-use futures01::future::{Either, Future, IntoFuture};
-use sentry::configure_scope;
+use futures::compat::Future01CompatExt;
+use futures::future::{self, Either};
+use futures::{Future, FutureExt, TryFutureExt};
+// use futures01::future::{Either, Future, IntoFuture};
```
jan-auer, line 11:
```suggestion
```

src/actors/cficaches.rs
```diff
@@ -199,7 +204,7 @@ impl CfiCacheActor {
     pub fn fetch(
         &self,
         request: FetchCfiCache,
-    ) -> impl Future<Item = Arc<CfiCacheFile>, Error = Arc<CfiCacheError>> {
+    ) -> Pin<Box<dyn Future<Output = Result<Arc<CfiCacheFile>, Arc<CfiCacheError>>>>> {
```
jan-auer, line 207: The clippy warning here is a good one. Symbolicator often uses such a type:

```rust
type SharedResult<T, E> = Result<Arc<T>, Arc<E>>;
pub fn fetch(...) -> Box<Pin<SharedResult<CfiCacheFile, CfiCacheError>>>;
```

Alternatively, probably for our async traits, this one may be a better fit:

```rust
type AsyncResult<T, E> = Pin<Box<dyn Future<Output Result<T, E>>>>;
pub fn fetch(...) -> AsyncResult<Arc<CfiCacheFile>, Arc<CfiCacheError>>;
```

src/actors/cficaches.rs
```diff
@@ -220,32 +225,31 @@ impl CfiCacheActor {
...
-                        objects_actor: objects,
-                        object_meta,
-                        threadpool,
-                    }))
-                })
-                .unwrap_or_else(move || {
-                    Either::B(
-                        Ok(Arc::new(CfiCacheFile {
+        object
+            .and_then(move |object| {
```
jan-auer, line 229: Consider this optional for this PR, but I would like to have this addressed in follow-ups: 

This function is large and uses unidiomatic combinators where async/await could be used. We could rewrite this into an `async fn`, split it up in smaller parts with speaking names, and then call a single `async fn` in the trait impl, surrounded by `Box::pin`.

src/actors/common/cache.rs
```diff
@@ -2,19 +2,21 @@ use std::collections::BTreeMap;
...
+use std::pin::Pin;
 use std::sync::Arc;
 
-use actix::ResponseFuture;
-use futures01::future::{self, Future, Shared};
-use futures01::sync::oneshot;
+use futures::channel::oneshot;
+use futures::future::{self, Future, FutureExt, Shared, TryFutureExt};
+// use futures01::future::{self, Future, Shared};
+// use futures01::sync::oneshot;
```
jan-auer, line 11:
```suggestion
```

src/actors/common/cache.rs
```diff
@@ -325,13 +348,15 @@ impl<T: CacheItemRequest> Cacher<T> {
...
-            })
-            .and_then(|shared| (*shared).clone());
+        let future = channel.unwrap_or_else(move |_cancelled_error| {
+            let message = format!("{} computation channel dropped", name);
+            Err(Arc::new(
+                io::Error::new(io::ErrorKind::Interrupted, message).into(),
+            ))
+        });
+        // TODO(flub): what was this cloning??
+        // .and_then(|shared| (*shared).clone());
```
jan-auer, line 358: This is a [shared](https://docs.rs/futures/0.1.30/futures/future/trait.Future.html#method.shared) receiver of a oneshot channel, and it used to clone the inner data. In futures 0.3, this has changed, and the data has to be clonable to begin with.

src/actors/cficaches.rs
flub: lol, i even specifically looked for things like this before committing... ah well.

src/actors/cficaches.rs
flub: I'm not a big fan of type aliases they often make things harder to read when coming back to the code.  In this case it's all decoration and the real signature is only the `Result` part.  But if clippy says I'll obey, I think the `AsyncResult` is my preferred option, or maybe even more general `PinedFuture<Result<...>>` (though given `FutureExt::boxed()` naming maybe `BoxedFuture` is a better name)

src/actors/cficaches.rs
flub: Yes, this is definitely something that should happen, but definitely as a followup.

src/actors/cficaches.rs
jan-auer: I think being opinionated with these types and having a semantic `AsyncResult` can be a good thing. I've made good experience with that in Relay, since they are quick to learn and then improve readability while also helping to streamline the code.

src/actors/common/cache.rs
```diff
@@ -273,19 +281,29 @@ impl<T: CacheItemRequest> Cacher<T> {
...
+                Err(err) => Err(Arc::new(err)),
+            };
+            drop(remove_computation_token);
+            sender.send(result).ok();
+            Ok(())
+        }
+        .bind_hub(Hub::new_from_top(Hub::current()));
+        // let channel = future::lazy(move |_| slf.compute(request, key))
+        //     .then(move |result| {
+        //         // Drop the token first to evict from the map. This ensures that callers either
```
Swatinem: please move the comment to the right place

src/actors/common/cache.rs
flub: thanks

Swatinem (commented):
👍 looks like a good first step. Moving this out of the trait method, and having that just be a wrapper around the async-fn call sounds like a good idea!

src/actors/common/cache.rs
flub: I forgot to remove this whole block, it was the alternative that I didn't get to work.  The async move block above does the same.

flub: huh, never mind.  you're right and i re-added the comment.

Swatinem (approved):

src/actors/common/cache.rs
```diff
@@ -2,19 +2,28 @@ use std::collections::BTreeMap;
...
 use crate::cache::{get_scope_path, Cache, CacheKey, CacheStatus};
 use crate::types::Scope;
 use crate::utils::futures::CallOnDrop;
-use crate::utils::sentry::SentryFutureExt;
+
+/// A pinned, boxed future.
+///
+/// This is the type of future that [`futures::FutureExt::boxed`] would return.  This is
+/// pretty boring but clippy quickly complains about type complexity without this.
+pub type BoxedFuture<T> = Pin<Box<dyn Future<Output = T>>>;
```
jan-auer, line 23: This can probably be declared in some utility module since it's a pattern we'll use more in this code base.

jan-auer (approved):

PR #324, opened 2020-12-22
Feat/dif info unwind

flub:
This adds the unwind info to dif candidates.  For this the cficache actor
also exposes DIF info and the symbolication actor now correctly merges
the various DIF candidates info together.

sentry snapshots do not need updating

### DIFF
--- a/src/actors/cficaches.rs
+++ b/src/actors/cficaches.rs
@@ -120,44 +120,38 @@ impl CacheItemRequest for FetchCfiCacheInternal {
     /// The extracted CFI is written to `path` in symbolic's
     /// [`CfiCache`](symbolic::minidump::cfi::CfiCache) format.
     fn compute(&self, path: &Path) -> BoxedFuture<Result<CacheStatus, Self::Error>> {
-        let objects_actor = self.objects_actor.clone();
-        let meta_handle = self.meta_handle.clone();
-        let threadpool = self.threadpool.clone();
         let path = path.to_owned();
+        let object = self
+            .objects_actor
+            .fetch(self.meta_handle.clone())
+            .map_err(CfiCacheError::Fetching);
 
-        let result = async move {
-            let object_handle = objects_actor
-                .fetch(meta_handle.clone())
-                .await
-                .map_err(CfiCacheError::Fetching)?;
-
+        let threadpool = self.threadpool.clone();
+        let result = object.and_then(move |object| {
             let future = async move {
-                let status = match object_handle.status() {
-                    CacheStatus::Positive => match write_cficache(&path, &*object_handle) {
-                        Ok(()) => CacheStatus::Positive,
-                        Err(err) => {
-                            log::warn!("Could not write cficache: {}", err);
-                            sentry::capture_error(&err);
-                            CacheStatus::Malformed
-                        }
-                    },
-                    cache_status => cache_status,
+                if object.status() != CacheStatus::Positive {
+                    return Ok(object.status());
+                }
+
+                let status = if let Err(e) = write_cficache(&path, &*object) {
+                    log::warn!("Could not write cficache: {}", e);
+                    sentry::capture_error(&e);
+
+                    CacheStatus::Malformed
+                } else {
+                    CacheStatus::Positive
                 };
+
                 Ok(status)
             };
 
-            match threadpool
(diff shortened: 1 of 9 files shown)


src/actors/cficaches.rs
```diff
@@ -104,46 +112,52 @@ impl CacheItemRequest for FetchCfiCacheInternal {
...
+                let status = match object_handle.status() {
+                    CacheStatus::Positive => match write_cficache(&path, &*object_handle) {
+                        Ok(()) => CacheStatus::Positive,
+                        Err(err) => {
+                            log::warn!("Could not write cficache: {}", err);
+                            sentry::capture_error(&err);
+                            CacheStatus::Malformed
+                        }
+                    },
+                    cache_status => cache_status,
```
jan-auer, line 144: I preferred the former structure as it was easier to read and had less indentation.

src/actors/cficaches.rs
```diff
@@ -104,46 +112,52 @@ impl CacheItemRequest for FetchCfiCacheInternal {
...
 
-            threadpool
+            match threadpool
                 .spawn_handle(future.bind_hub(Hub::current()))
-                .unwrap_or_else(|_| Err(CfiCacheError::Canceled))
-        });
+                .await
+            {
+                Ok(result) => result,
+                Err(_) => Err(CfiCacheError::Canceled),
```
jan-auer, line 154: _nit_: Rewrite this match statement as `.map_err(|_| CfiCacheError::Canceled)`.

src/actors/cficaches.rs
```diff
@@ -104,46 +112,52 @@ impl CacheItemRequest for FetchCfiCacheInternal {
...
-                    sentry::capture_error(&e);
+        let path = path.to_owned();
 
-                    CacheStatus::Malformed
-                } else {
-                    CacheStatus::Positive
+        let result = async move {
+            let object_handle = objects_actor
+                .fetch(meta_handle.clone())
+                .await
```
Swatinem, line 131: this await here is the reason you have two nested futures. Only the inner one runs in the threadpool. Is that intentional? Maybe add a comment explaining the reasoning here.

src/actors/cficaches.rs
```diff
@@ -104,46 +112,52 @@ impl CacheItemRequest for FetchCfiCacheInternal {
...
+                        Err(err) => {
+                            log::warn!("Could not write cficache: {}", err);
+                            sentry::capture_error(&err);
+                            CacheStatus::Malformed
+                        }
+                    },
+                    cache_status => cache_status,
                 };
-
                 Ok(status)
```
jan-auer, line 146: This future does not have to return a result, you should be able to return `status` directly.

src/actors/cficaches.rs
```diff
@@ -168,14 +182,38 @@ impl CacheItemRequest for FetchCfiCacheInternal {
...
+                    }
+                } else {
+                    // No need to pretend that we were going to use this cficache if the
+                    // original object file was already not there, that status is already
+                    // reported.
+                    ObjectUseInfo::None
+                }
+            }
+            CacheStatus::Malformed => ObjectUseInfo::Malformed,
+        };
```
jan-auer, line 200: _nit_: This block could be moved into a function. This would be great since this will make unit testing it easier and we can add a descriptive doc comment that describes the edge cases it handles.

src/actors/symbolication.rs
```diff
@@ -1463,14 +1468,18 @@ impl SymbolicationActor {
...
         // Go through all the modules in the minidump and build a map of the modules with
         // missing or malformed CFI.  ObjectFileStatus::Found is only added when the file is
         // loaded as it could still be corrupted when reading from the cache.  Usable CFI
         // cache files are added to the frame_info_map.
         for (code_module_id, result) in &cfi_results {
             let cache_file = match result {
-                Ok(x) => x,
+                Ok(x) => {
+                    dif_candidates.insert(*code_module_id, x.candidates().clone());
+                    x
```
jan-auer, line 1481: _nit_: Since you're touching that already, can you give this variable a more speaking name?

jan-auer (approved):
A number of small suggestions below, but overall G2G from my end.

Swatinem (approved):

PR #325, opened 2020-12-22
build: Update symbolic to support MIPS and fix stack scanning

jan-auer:
Pulls in https://github.com/getsentry/symbolic/pull/281

This version of breakpad should compute the ELF debug_identifier
correctly for big-endian minidumps processed in a little-endian
system.


flub (approved):

jan-auer:
Due to code freeze we will merge this into the `freeze` branch. It will assume the snapshot warning, and snapshots can be updated in Sentry on 2021-01-04 after code freeze changes.

PR #329, opened 2021-01-03
ref(downloader): Make the GCS downloader a stateful struct

jan-auer:
This is a work-in-progress proof of concept for creating stateful downloaders rather than storing caches in global static data structures. This will allow to unit-test the downloaders more easily.

As part of this, I noticed that we largely do not to return `impl Future` and can use `async fn` directly. This is possible even in the case of the `download` function, although it is made harder by the thread pool. Since the pool requires `'static` futures, we cannot spawn any async function that takes `&self`. This requires one of the following workarounds:

1. Get rid of the thread pool and move resource delegation up the call hierarchy. Probably the most desirable solution, but also exceeds the scope of this PR.
2. Use [`async-scoped`](https://docs.rs/async-scoped) to spawn a future with a bound life time. In our case, this would be perfectly safe, since the downloaders do not have constrained lifetime in the application. However, it just layers more complexity on top of an existing problem.
3. Clone `self` inside the function. This creates the requirement that all fields internally be cloneable (i.e. likely put inside an `Arc`). We use this pattern in several places, but it is not easily spottable when looking at usage sites. I would rather not introduce another instance of this now.
4. Finally, require the caller to `Arc` the downloader. In fact, we already require this for the downloader, and by changing `Downloader::new`, we codify this contract. It can be reverted easily once the thread pool is dropped, and until then provides a clear interface.

If desired, I can split this up into two PRs: One refactoring to `async fn` and another one to introduce the stateful GCS downloader.

#skip-changelog

### DIFF
--- a/src/actors/objects/data_cache.rs
+++ b/src/actors/objects/data_cache.rs
@@ -138,83 +138,85 @@ impl CacheItemRequest for FetchFileDataRequest {
             self.0.file_id.write_sentry_scope(scope);
         });
 
+        let file_id = self.0.file_id.clone();
+        let downloader = self.0.download_svc.clone();
         let download_file = tryf03pin!(self.0.data_cache.tempfile());
         let download_dir =
             tryf03pin!(download_file.path().parent().ok_or(ObjectError::NoTempDir)).to_owned();
-        let request = self
-            .0
-            .download_svc
-            .download(self.0.file_id.clone(), download_file.path().to_owned())
-            .map_err(Into::into);
 
-        let result = request.and_then(move |status| {
-            async move {
-                match status {
-                    DownloadStatus::Completed => {
-                        log::trace!("Finished download of {}", cache_key);
-                        let decompress_result =
-                            decompress_object_file(&download_file, tempfile_in(download_dir)?);
+        let future = async move {
+            let status = downloader
+                .download(file_id, download_file.path().to_owned())
+                .await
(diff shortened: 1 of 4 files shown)


jan-auer:
I recommend to review this commit-by-commit and hide whitespace changes in the diff view:
image

src/actors/objects/data_cache.rs
```diff
@@ -138,83 +138,85 @@ impl CacheItemRequest for FetchFileDataRequest {
...
-                        } else {
-                            // Attempt to parse the object to capture errors. The result can be
-                            // discarded as the object's data is the entire ByteView.
-                            if archive.object_by_index(0).is_err() {
-                                return Ok(CacheStatus::Malformed);
-                            }
 
-                            io::copy(&mut view.as_ref(), &mut persist_file)?;
-                        }
+        let future = async move {
```
jan-auer, line 147: This was changed before I made `download()` require an `Arc`. Since the downloader (Arc) is moved into the async block, it's lifetime is contained within the future. The entire resulting block therefore did not expose the downloader's lifetime.

Since we now consume the `Arc`, there is no more reason to change this. However, the new block seems more readable with async await, which is why I kept it. In fact, some of the `tryf03pin!` calls could also be moved inside the block.

Swatinem (approved):

Swatinem (approved):

PR #331, opened 2021-01-04
test(downloader): Migrate GCS tests to Rust

jan-auer:
Migrates the GCS integration test to Rust and splits it into several smaller tests. Since the only responsibility of the GCS source is to connect to actual Google Cloud Storage, there is little point in mocking HTTP calls.

Authentication caching remains to be tested in a future PR.

#skip-changelog


src/services/download/gcs.rs
```diff
@@ -273,3 +273,153 @@ impl GcsDownloader {
...
+            None
+        } else {
+            Some(GcsSourceKey {
+                private_key,
+                client_email,
+            })
+        }
+    }
+
+    macro_rules! gcs_source_key {
```
Swatinem, line 300: interesting that you can have name conflicits between functions and macros…
Maybe its worth inlining?

Swatinem (approved):

src/services/download/gcs.rs
```diff
@@ -273,3 +273,153 @@ impl GcsDownloader {
...
+                    return;
+                }
+            }
+        }
+    }
+
+    fn gcs_source(source_key: GcsSourceKey) -> Arc<GcsSourceConfig> {
+        Arc::new(GcsSourceConfig {
+            id: SourceId::new("gcs-test"),
+            bucket: "sentryio-system-symbols-0".to_owned(),
```
flub, line 315: Not related to this test, i am somewhat worried about us connecting to a real system bucket for tests.  I hope our credentials are at least read only?  but i suspect they might not be

src/services/download/gcs.rs
```diff
@@ -273,3 +273,153 @@ impl GcsDownloader {
...
+        digest.update(std::fs::read(target_path).unwrap());
+        let hash = digest.finalize();
+
+        assert_eq!(
+            hash.as_ref(),
+            [
+                0x20, 0x6e, 0x63, 0xc0, 0x6d, 0xa1, 0x35, 0xbe, 0x18, 0x58, 0xdd, 0xe0, 0x37, 0x78,
+                0xca, 0xf2, 0x5f, 0x84, 0x65, 0xb8,
+            ]
+        );
```
flub, line 377: l: more for cosmetic reasons than anything else, but comparing hexdigests would be nicer to my eyes.

flub (approved):

src/services/download/gcs.rs
Swatinem: I had the same concerns re depending on external services in the tests. We looked up how to do mocking for the s3 case, and it would basically change the complete code that we test.

jan-auer: The credentials are read-only, checked. Thanks.

src/services/download/gcs.rs
jan-auer: Updated.

PR #335, opened 2021-01-07
ref(downloader): Upgrade downloaders to tokio 1.0

jan-auer:
Introduces Tokio 1 in the downloader module, migrates to reqwest, and updates to the latest version of rusoto for S3. Downloaders share the same HTTP client and run in a single-threaded runtime. At this point, the new runtime is only used by the downloaders.

Since reqwest does not allow to hook into the resolver, this uses a patched version: jan-auer/reqwest@62edd49 until reqwest adds the required APIs.

### DIFF
--- a/src/services/download/gcs.rs
+++ b/src/services/download/gcs.rs
@@ -5,12 +5,7 @@
 use std::path::PathBuf;
 use std::sync::Arc;
 
-use actix_web::error::JsonPayloadError;
-use actix_web::{client, HttpMessage};
 use chrono::{DateTime, Duration, Utc};
-use client::SendRequestError;
-use failure::Fail;
-use futures::compat::{Future01CompatExt, Stream01CompatExt};
 use futures::prelude::*;
 use parking_lot::Mutex;
 use serde::{Deserialize, Serialize};
@@ -72,10 +67,8 @@ pub enum GcsError {
     Base64(#[from] base64::DecodeError),
     #[error("failed encoding JWT")]
     Jwt(#[from] jsonwebtoken::errors::Error),
-    #[error("failed to parse JSON response")]
-    Json(#[from] failure::Compat<JsonPayloadError>),
     #[error("failed to send authentication request")]
-    Auth(#[from] failure::Compat<SendRequestError>),
+    Auth(#[source] reqwest::Error),
 }
 
 /// Parses the given private key string into its binary representation.
@@ -118,12 +111,17 @@ fn get_auth_jwt(source_key: &GcsSourceKey, expiration: i64) -> Result<String, Gc
 #[derive(Debug)]
 pub struct GcsDownloader {
     token_cache: Mutex<GcsTokenCache>,
+    client: reqwest::Client,
 }
 
 impl GcsDownloader {
     pub fn new() -> Self {
         let token_cache = Mutex::new(GcsTokenCache::new(GCS_TOKEN_CACHE_SIZE));
-        Self { token_cache }
+        let client = reqwest::Client::new();
+        Self {
+            token_cache,
+            client,
+        }
     }
 
     /// Requests a new GCS OAuth token.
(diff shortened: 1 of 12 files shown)


src/services/download/gcs.rs
```diff
@@ -118,41 +111,45 @@ fn get_auth_jwt(source_key: &GcsSourceKey, expiration: i64) -> Result<String, Gc
...
 
-        let mut builder = client::post("https://www.googleapis.com/oauth2/v4/token");
-        // for some inexplicable reason we otherwise get gzipped data back that actix-web
-        // client has no idea what to do with.
-        builder.header("accept-encoding", "identity");
-        let response = builder
+        let request = self
+            .client
+            .post("https://www.googleapis.com/oauth2/v4/token")
+            // TODO(ja): check if this still holds
```
Swatinem, line 135: I think encodings are feature flags on reqwest, or they are builtin even?

jan-auer: According to the reqwest main readme those are off-by-default feature flags, correct. This PR is still very much in progress and needed some leg work to be merged first, so all of this still needs some time to get right :)

jan-auer:
This will be rebased on top of #336.

jan-auer:
There are some changes to the symbolication actor wrt `measure_task_timeout` which I can pull out into a separate PR. 

The only failing integration tests are IP/localhost gating. Everything else should be passing now.

src/utils/futures.rs
```diff
@@ -328,73 +303,73 @@ where
...
-    fn test_remote_thread_clone_drop() {
-        // When dropping a clone of the RemoteThread the other one needs to keep working.
-        let remote0 = setup_remote_thread();
-        let remote1 = remote0.clone();
-        std::mem::drop(remote0);
-        let fut = remote1.spawn("task", Duration::from_secs(10), || future::ready(42));
-        let ret = test::block_fn(|| fut);
-        assert_eq!(ret, Ok(42));
-    }
+    // use super::*;
```
Swatinem, line 306: do we still need these?

jan-auer: Apart from the timeout tests, no.

jan-auer:
@flub @Swatinem do you have any preferences regarding error handling? I think we should be fine with the current way of wrapping in a `::Reqwest` variant, although that is not the most semantic way to do it. I'm thinking we can touch this in a follow-up.

Swatinem:
> do you have any preferences regarding error handling?

I think its fine, either way we can revisit this.

flub:
Yeah, I think this is fine for now.  I thought in general we're heading to using plain `anyhow::Error` unless stuff is matched on anyway.

src/services/download/mod.rs
```diff
@@ -142,15 +138,16 @@ impl DownloadService {
         let hub = Hub::current();
         let slf = self.clone();
 
-        let spawn_result = self
-            .worker
-            .spawn("service.download", Duration::from_secs(300), || {
-                slf.dispatch_download(source, destination).bind_hub(hub)
-            })
-            .await;
+        let _guard = self.worker.enter();
```
Swatinem, line 140: note to self: read up on why this is needed

jan-auer:
This is now ready for final review. I would love to roll this soon with @tonyo.

There's still one integration test failing that needs fixing: `test_reserved_ip_addresses[127.0.0.1-False]`

Swatinem (approved):

src/services/download/mod.rs
jan-auer: Good point, I should add a comment on this. The timeout we create below requires an active runtime since it's based on `Sleep`, which internally stores a reference to the runtime's clock. `enter` registers the runtime in thread locals so that we can appropriately create the futures and then spawn them. 

This will go away as soon as we move the executor up and run the entire downloader in the tokio 1 runtime (read: "soon" :) )

jan-auer:
Sorry for the false-alarm. Turns out the integration test uncovered a major problem:

Reqwest uses `hyper::client::HttpConnector` under the hood, which skips the resolver if the URI given specifies a valid IP address instead of a host name. This is a problem, since we only hook into the resolver. As it turns out, it's not as straight forward to hook into the connector. 

I'll try to see if there is a solution which doesn't involve forking hyper. Worst case, we have to wrap all request calls and check IP addresses manually.

flub (approved):

PR #336, opened 2021-01-07
test(downloader): Migrate S3 tests to Rust

jan-auer:
Same as #331 for the S3 tests. As part of this, there are two changes:

- ~Environment variables have been simplified. They now match the AWS default credentials variable names and will be auto-discovered by the `aws s3` command.~
- Since the data in the bucket does not change usually, we treat it like an external lookup bucket. The setup routine only runs once to populate the bucket but then never again. This is to speed up tests. If we end up adding more files, we can perform individual checks for files, instead.

#skip-changelog


jan-auer:
I think I'll finish this before #335, since we just have to `compat()` some of the S3 client futures whereas the reqwest refactor will require more care.

src/services/download/s3.rs
```diff
@@ -144,3 +144,224 @@ impl S3Downloader {
...
+            rusoto_core::HttpClient::new().expect("create S3 HTTP client"),
+            rusoto_credential::StaticProvider::new_minimal(
+                source_key.access_key,
+                source_key.secret_key,
+            ),
+            source_key.region,
+        )
+    }
+
+    fn gcs_source(source_key: S3SourceKey) -> Arc<S3SourceConfig> {
```
flub, line 197: May want to rename this ;)

src/services/download/s3.rs
```diff
@@ -144,3 +144,224 @@ impl S3Downloader {
...
+            list[0].location(),
+            SourceLocation::new("50/2fc0a51ec13e479998684fa139dca7/debuginfo")
+        );
+    }
+
+    /// Loads mock data into the S3 test bucket.
+    ///
+    /// This function will exit early if the bucket is already populated. Eventually, this should be
+    /// replaced with a one-off setup script.
+    async fn setup_bucket(source_key: S3SourceKey) {
```
flub, line 235: How well does this work when two tests are racing each other to set this up?  Is it reliable?  Maybe this should use a `std::sync::Once` to make it more explicitly safe?

flub (approved):

src/services/download/s3.rs
```diff
@@ -144,3 +144,224 @@ impl S3Downloader {
...
+            })
+        }
+    }
+
+    macro_rules! s3_source_key {
+        () => {
+            match s3_source_key() {
+                Some(key) => key,
+                None => {
+                    println!("Skipping due to missing SENTRY_SYMBOLICATOR_TEST_AWS_ACCESS_KEY_ID or SENTRY_SYMBOLICATOR_TEST_AWS_SECRET_ACCESS_KEY");
```
Swatinem, line 179:
```suggestion
                    println!("Skipping due to missing SENTRY_SYMBOLICATOR_AWS_ACCESS_KEY_ID or SENTRY_SYMBOLICATOR_AWS_SECRET_ACCESS_KEY");
```

src/services/download/s3.rs
```diff
@@ -144,3 +144,224 @@ impl S3Downloader {
...
+    use crate::types::ObjectType;
+
+    use super::*;
+    use sha1::{Digest as _, Sha1};
+
+    const S3_BUCKET: &str = "symbolicator-test";
+
+    fn s3_source_key() -> Option<S3SourceKey> {
+        let access_key = std::env::var("AWS_ACCESS_KEY_ID").ok()?;
+        let secret_key = std::env::var("AWS_SECRET_ACCESS_KEY").ok()?;
```
Swatinem, line 161:
```suggestion
        let access_key = std::env::var("SENTRY_SYMBOLICATOR_AWS_ACCESS_KEY_ID").ok()?;
        let secret_key = std::env::var("SENTRY_SYMBOLICATOR_AWS_SECRET_ACCESS_KEY").ok()?;
```

src/services/download/s3.rs
Swatinem, line 197:
```suggestion
    fn s3_source(source_key: S3SourceKey) -> Arc<S3SourceConfig> {
```

src/services/download/s3.rs
Swatinem: my question is rather: why do we need this?

.github/workflows/ci.yml
```diff
@@ -81,6 +81,8 @@ jobs:
         env:
           SENTRY_SYMBOLICATOR_GCS_CLIENT_EMAIL: ${{ secrets.SENTRY_SYMBOLICATOR_GCS_CLIENT_EMAIL }}
           SENTRY_SYMBOLICATOR_GCS_PRIVATE_KEY: ${{ secrets.SENTRY_SYMBOLICATOR_GCS_PRIVATE_KEY }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
Swatinem, line 85:
```suggestion
          SENTRY_SYMBOLICATOR_AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
          SENTRY_SYMBOLICATOR_AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```

just to keep this consistent

src/services/download/s3.rs
jan-auer: We tried this out, but it didn't give much of a speedup. On the contrary, if our setup code panics we will just receive a poison error instead of the verbose error message.

src/services/download/s3.rs
jan-auer: This was an intentional change, but I've reverted it now.

src/services/download/s3.rs
jan-auer: @Swatinem we need this so that the tests are self-contained and externals can also test this easily with their own S3 accounts.

jan-auer:
auth env vars will be addressed in a separate PR.

PR #337, opened 2021-01-07
ref(downloader): Make HTTP and FS downloaders stateful structs

jan-auer:
Continues work from #330 for the HTTP download source. This is in preparation of moving to reqwest.

#skip-changelog


flub (approved):
i like how the PR is a small simple step :)

flub (approved):
I like how the PR is a small simple step :)

PR #340, opened 2021-01-08
ref(futures): Use futures' BoxFuture instead of our own alias

flub:
Turns out futures already has this type alias which works well with
the future combinators provided by the FutureExt trait.  So switch
over to this.

This ends up being more strongly typed, we need to be explicit about
lifetime and Send instead of letting the compiler resolve this.

**Note:** I started this to use an already defined type, but the
lifetime and Send specific versions of this do make it more
complicated.  The previous version was also closer to what async-trait
does and to what plain async/await does.  So **I actually am -1 on
merging this PR**.  Still, it's interesting to discuss and some code
does get a little bit more readable as now we can use the .boxed()
.boxed_local() combinators.

#skip-changelog

### DIFF
--- a/src/actors/cficaches.rs
+++ b/src/actors/cficaches.rs
@@ -5,6 +5,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use futures::compat::Future01CompatExt;
+use futures::future::LocalBoxFuture;
 use futures::prelude::*;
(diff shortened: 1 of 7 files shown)


Swatinem:
Hm, I’m also not convinced this is what we should do?

PR #343, opened 2021-01-11
ref(utils): Introduce measure util for futures

jan-auer:
Introduces two new future utilities:

- `measure` function allows to time futures and their results. It is
  supposed to eventually replace the other uses of the `future_metrics!`
  macro, too.
- `timeout_compat` creates a timeout around a future that can run on a
  `tokio 0.1` executor. This is to transition to the new `tokio 1.0`
  timeout function.

Both utilities are introduced in preparation of #335.

#skip-changelog

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -12,7 +12,7 @@ use apple_crash_report_parser::AppleCrashReport;
 use bytes::{Bytes, IntoBuf};
 use chrono::{DateTime, TimeZone, Utc};
 use futures::compat::Future01CompatExt;
-use futures::{future, select, FutureExt as _, TryFutureExt};
+use futures::{future, FutureExt as _, TryFutureExt};
 use futures01::future::{Future as _, Shared};
 use futures01::sync::oneshot;
 use parking_lot::Mutex;
@@ -45,7 +45,7 @@ use crate::types::{
     RequestId, RequestOptions, Scope, Signal, SymbolicatedFrame, SymbolicationResponse, SystemInfo,
 };
 use crate::utils::addr::AddrMode;
-use crate::utils::futures::{CallOnDrop, ThreadPool};
+use crate::utils::futures::{m, measure, timeout_compat, CallOnDrop, ThreadPool};
 use crate::utils::hex::HexValue;
 
 /// Options for demangling all symbols.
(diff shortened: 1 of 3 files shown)


Swatinem (approved):

src/utils/futures.rs
```diff
@@ -285,6 +285,30 @@ impl Drop for CallOnDrop {
     }
 }
 
+/// Error returned by [`timeout_compat`].
+#[derive(Debug, PartialEq, thiserror::Error)]
+#[error("deadline has elapsed")]
+pub struct Elapsed(());
```
flub, line 291: huh, i always wrote stuff like this as `struct Elapsed { }`.  guess it doesn't matter

src/actors/symbolication.rs
```diff
@@ -1011,53 +1011,24 @@ pub struct SymbolicateStacktraces {
...
-        .map(|mut response| {
-            if !serialize_dif_candidates {
-                response.clear_dif_candidates();
-            }
-            response
-        })
+
+        let f = self.do_symbolicate_impl(request);
+        let f = timeout_compat(Duration::from_secs(3600), f);
+        let f = measure("symbolicate", m::timed_result, f);
```
flub, line 1023: This emits a new metric, `futures.wait_time`.  And since this is not being spawned it will always be super-short since it's being polled right away.  This will seriously skew our `futures.wait_time` metric to the point it's not useful as a metric to look at.
Luckily our default dashboard groups by task_name already, but I still don't think this is nice.  Can we have a variation which doesn't measure `futures.wait_time`?  Operationally having to know when this metric tells something meaningful and when not is not nice, it's mere presence indicates it's showing spawn-time and this violates that.

src/actors/symbolication.rs
```diff
@@ -1752,7 +1723,9 @@ impl SymbolicationActor {
                 .await
         };
 
-        measure_task_timeout("minidump_stackwalk", future, Duration::from_secs(1200)).await
+        let future = timeout_compat(Duration::from_secs(3600), future);
+        let future = measure("minidump_stackwalk", m::timed_result, future);
```
flub, line 1727: same issue

src/actors/symbolication.rs
```diff
@@ -1933,19 +1906,16 @@ impl SymbolicationActor {
...
         };
 
-        measure_task_timeout(
-            "parse_apple_crash_report",
-            request_future,
-            Duration::from_secs(1200),
-        )
-        .await
+        let future = timeout_compat(Duration::from_secs(1200), future);
+        let future = measure("parse_apple_crash_report", m::timed_result, future);
```
flub, line 1917: same issue

src/services/download/sentry.rs
```diff
@@ -199,7 +199,7 @@ impl SentryDownloader {
         };
 
         let search = self.cached_sentry_search(query, config);
-        let entries = future_utils::time_task("downloads.sentry.index", search).await?;
+        let entries = measure("downloads.sentry.index", m::result, search).await?;
```
flub, line 202: same issue

src/utils/futures.rs
jan-auer: So did I, but turns out you can construct such a type.

src/actors/symbolication.rs
jan-auer: The `futures.wait_time` metric did exist and was only dropped in a recent refactor of how we report future metrics since we said it was not super crucial to us. It was never meant to be read across tasks, so nothing changes in the contract.

src/utils/futures.rs
Swatinem: … Now that you mention it, why not `struct Elapsed;` without any members. I think that makes the intention clearer, or do you actually need a member?

jan-auer: I copied tokio's [`Elapsed`](https://docs.rs/tokio/1.0.1/tokio/time/error/struct.Elapsed.html), that's why. But the reason is always the same: Prevent that the type is constructed by a user.

src/actors/symbolication.rs
flub: doesn't change the reasoning why this is a bad idea: `futures.wait_time` measures the time a future was waiting before it first gets polled, to add it to futures which are already being polled makes no sense and gives a false impression when looking at the metrics.

PR #344, opened 2021-01-11
feat(difs): Expose dif location as a URI in candidates

flub:
This adds a "uri" field to the dif candidates info.

To do this it firstly refactors the SourceFileId into an
ObjectFileSource to provide better structure and a unified place to
add the URI.

The second commit then uses this to add to the DIF candidates.  This
also needs to add insta redaction since the port number shows up in
the URI otherwise.

sentry snapshots are not affected by this as they do not enable dif candidates info.


Swatinem (approved):

src/actors/objects/mod.rs
```diff
@@ -126,10 +126,12 @@ impl From<debuginfo::ObjectError> for ObjectError {
...
-    /// The [`SourceId`] of the object which caused the [`ObjectError`] while being fetched.
-    source_id: SourceId,
-    /// The [`SourceLocation`] of the object which caused the [`ObjectError`] while being fetched.
-    source_location: SourceLocation,
+    /// The object file which was attempted to be fetched.
+    file_source: ObjectFileSource,
+    // /// The [`SourceId`] of the object which caused the [`ObjectError`] while being fetched.
+    // source_id: SourceId,
+    // /// The [`SourceLocation`] of the object which caused the [`ObjectError`] while being fetched.
+    // source_location: SourceLocation,
```
jan-auer, line 134:
```suggestion
```

src/actors/snapshots/symbolicator__actors__symbolication__tests__add_bucket-2.snap
```diff
@@ -31,6 +31,7 @@ modules:
     candidates:
       - source: local
         location: 502F/C0A5/1EC1/3E47/9998/684FA139DCA7
+        uri: "http://localhost:<port>/download/502F/C0A5/1EC1/3E47/9998/684FA139DCA7"
```
jan-auer, line 34: Q: Can we just make this the `location`?

src/services/download/s3.rs
```diff
@@ -78,15 +78,14 @@ impl S3Downloader {
 
     pub async fn download_source(
         &self,
-        source: Arc<S3SourceConfig>,
-        download_path: SourceLocation,
+        file_source: S3ObjectFileSource,
```
jan-auer, line 81: Would it be possible to keep those attributes separate in the downloader implementations? There's probably a good reason you combined them, but looking at the result now I think the previous signature was more straight forward and required less type definitions.

src/actors/snapshots/symbolicator__actors__symbolication__tests__add_bucket-2.snap
flub: yeah, i imagine the UI doesn't want both.  Do we:
1. store the value of `uri` above in `location`
2. leave `uri` as above and remove `location` (will break current UI)

just using location gives us a little more leeway as it does not imply a URI of any sort (and there's the weird `noent://` case to deal with).

src/services/download/s3.rs
flub: I found this easier to work with to be honest.  Plus it allows to implement things like `.url()` on the ObjectFileSource type instead of needing them to be free-standing functions having knowledge about this type.

So in short, I'd like to keep it like this (unsurprisingly).

src/actors/snapshots/symbolicator__actors__symbolication__tests__add_bucket-2.snap
jan-auer: I'd go with Option 1 since otherwise location and URI seem like duplicate information. Agreed, `noent://*` is strange. In which case would it appear?

src/actors/objects/mod.rs
```diff
@@ -445,6 +439,7 @@ fn create_candidates(
         let info = ObjectCandidate {
             source: source_id,
             location: SourceLocation::new("*"),
+            uri: String::from("noent://*"),
```
flub, line 442: @jan-auer This is the weird "noent" case.  It is used when a source, i.e. sentry, returns no candidates with `list_files()` in which case we create a NotFound entry with this URL.  Other options:
1. Use the base URI of the source.  Confusing how to interpret though.
2. empty string
3. if we go to only having a `location` field in an `ObjectCandidate` and no `uri` field we could leave the old behaviour of just a `*`.

src/services/download/s3.rs
jan-auer: Understood, and this is probably very subjective. Feel free to merge like this 👍  

It would not look as strange to me if the downloaders defined their types in the same file rather than all source types 
(+ configs) being defined in a single file. That way, at least a downloader is self contained, and everything composes that.

The interface change to the downloaders is very unfortunate. Passing config + location in a single argument seems like an additional indirection when looking at it in isolation. As an idea, would it be possible to destructure this on a higher level and at least keep `download/*`the same?

src/actors/objects/mod.rs
jan-auer: `noent://*` works for this, will it be clear to a consumer, though? Option 2 (empty string) or even `None` seems most correct to reflect the intention IMHO. I'd not go with just `*` since one might actually expect to parse this and then fail.

flub: Could also directly exploit that it's not guaranteed to be a URI of any kind of scheme and that it's directly displayed in the UI and simply say `No objects files listed on this source` which you can always argue is a URI ;)

src/services/download/s3.rs
flub: In summary, going to leave this.  Having them as two separate structs meant there was a lot of random guessing how two of these bits went together.  This way brings more type safety.  (it's a bit unfortunate that we use an enum rather than a trait for the overarching type, but I don't want to make everything that handles this generic just now)

src/actors/objects/mod.rs
flub: went with my last unorthodox^Wpractical suggestion here

src/services/download/filesystem.rs
```diff
@@ -9,11 +9,43 @@ use std::io;
...
+        self.source.path.join(&self.location.path())
+    }
+
+    /// Returns the `file://` URI from which to fetch this object file.
+    ///
+    /// This is a quick-and-dirty approximation, not fully RFC8089-compliant.  E.g. we do
+    /// not provide a hostname nor percent-encode.  Use this only for diagnostics and use
+    /// [`FilesystemObjectFileSource::path`] if the actual file location is needed.
+    pub fn uri(&self) -> ObjectFileSourceURI {
+        format!("file:///{path}", path = self.path().display()).into()
```
Swatinem, line 46: I mean, the doc comment says its not supposed to be compliant, but do we have tests for this on windows and linux?

flub: Which tests do you mean?

src/services/download/http.rs
```diff
@@ -127,21 +135,22 @@ impl HttpDownloader {
...
         }
-        .map(|loc| SourceFileId::Http(source.clone(), loc))
+        .map(|loc| HttpObjectFileSource::new(source.clone(), loc).into())
         .collect()
     }
 }
 
 #[cfg(test)]
 mod tests {
+    use super::super::locations::SourceLocation;
```
jan-auer, line 153:
```suggestion
    use crate::locations::SourceLocation;
```

There's another instance of this.

src/actors/symbolication.rs
```diff
@@ -2213,7 +2238,9 @@ mod tests {
             service.symbolication().get_response(request_id, None)
         })?;
 
-        insta::assert_yaml_snapshot!(response);
+        insta::assert_yaml_snapshot!(response, {
+            ".**.location" => insta::dynamic_redaction(redact_localhost_port)
+        });
```
jan-auer, line 2243: _nit_: Deduplicate these via `macro_rules!`.

src/services/download/http.rs
```diff
@@ -8,35 +8,50 @@ use std::time::Duration;
...
+
+    pub fn uri(&self) -> ObjectFileSourceURI {
+        match self.url() {
+            Ok(url) => url.as_ref().into(),
+            Err(_) => "".into(),
+        }
+    }
+
+    /// Returns the URL from which to download this object file.
+    pub fn url(&self) -> Result<Url> {
```
jan-auer, line 52: I would prefer to make the signature `Result<Url, anyhow::Error>`, since it's easier to see at a glance what this returns. That said, since this is a `join_url` operation, can we make this return a URL-specific error? I'd rather have the caller coerce this.

src/types/mod.rs
```diff
@@ -522,6 +522,17 @@ pub struct ObjectCandidate {
...
+/// A URI representing an [`ObjectFileSource`].
+///
+/// Note that this does not provide enough information to download the object file, for this
+/// you need the actual [`ObjectFileSource`].  The purpose of this URI is to be able to
+/// display to a user who might be able to use this in other tools.  E.g. for an S3 source
+/// this could be an `s3://` URI etc.
+///
+/// [`ObjectFileSource`]: crate::services::download::ObjectFileSource
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
+pub struct ObjectFileSourceURI(String);
```
jan-auer, line 534: _nit_: Would `ObjectUri` suffice as a shorter name?

src/types/objects.rs
```diff
@@ -59,12 +66,17 @@ impl AllObjectCandidates {
     ///
     /// You can only request cficaches from a DIF object that was already in the metadata
     /// candidate list, therefore if the candidate is missing it is treated as an error.
-    pub fn set_unwind(&mut self, source: SourceId, location: SourceLocation, info: ObjectUseInfo) {
+    pub fn set_unwind(
+        &mut self,
+        source: SourceId,
+        location: &ObjectFileSourceURI,
```
jan-auer, line 72:
```suggestion
        uri: &ObjectFileSourceURI,
```

src/types/objects.rs
```diff
@@ -287,3 +299,24 @@ mod tests {
         assert_eq!(all.0[0].debug, ObjectUseInfo::Ok);
     }
 }
+
+impl ObjectFileSourceURI {
```
jan-auer, line 303: Please move these implementations next to the definition of the type.

jan-auer (commented):
LGTM, final comments below.

src/services/download/http.rs
flub: I don't get this comment, that location doesn't even exist.

src/services/download/http.rs
flub: I thought our current approach was to not use custom errors unless we need to match on them?

The `Result<Url>` part is more just familiarity imho, once you're used to `anyhow::Result` it's a pretty obvious shortcut.

src/services/download/http.rs
jan-auer: Good point, I just quickly saw the `super::super::` and got confused. Turns out, the `locations.rs` file is missing completely from this branch. Could you please check it in?

src/services/download/http.rs
jan-auer: So if you write `anyhow::Result` rather than importing, I'm also fine with it 👍  

Point taken, let's go with anyhow for now. I'll propose something for this in our style guide. What I've often seen the community do, is:

- If there is a function that performs a single task, then throw the specific error for this task. For example, I/O functions should throw `std::io::Error`. Parse functions should throw a custom parse error.
- For propagating, libraries should create their own error type, either with a plain `ErrorKind` enum or with flag getters.
- For propagating, applications may use a boxed error such as anyhow as long as no error *handling* is needed.

src/types/mod.rs
flub: meh, doesn't really matter but mostly i think that's too short for such an auxiliary type

src/services/download/filesystem.rs
flub: oh, i know what you mean :). So we only run prod on linux and we don't even run symbolicator tests on windows (I don't think it even can run there)

jan-auer (approved):

PR #345, opened 2021-01-13
test(wasm): Port wasm tests to Rust

jan-auer:
#skip-changelog


src/actors/symbolication.rs
```diff
@@ -2189,6 +2189,53 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_wasm_payload() -> anyhow::Result<()> {
+        let (service, _cache_dir) = setup_service();
+        let (_symsrv, source) = test::symbol_server();
+
+        let modules: Vec<RawObjectInfo> = serde_json::from_str(
```
jan-auer, line 2197: Deserializing directly into `CompleteObjectInfo` would require us to specify a set of irrelevant attributes and would no longer match the input to normal requests. 

I will revisit test utilities for this when porting the actual E2E tests.

Swatinem (approved):

src/actors/snapshots/symbolicator__actors__symbolication__tests__wasm_payload.snap
```diff
@@ -0,0 +1,31 @@
...
+        addr_mode: "rel:0"
+        instruction_addr: 0x8c
+        lang: rust
+        symbol: internal_func
+        sym_addr: 0x8b
+        function: internal_func
+        filename: src/lib.rs
+        abs_path: /Users/mitsuhiko/Development/wasm-example/simple/src/lib.rs
+        lineno: 19
+modules:
```
flub, line 19: I'm in two minds on whether this should include the dif candidates section or not

flub (approved):

src/actors/snapshots/symbolicator__actors__symbolication__tests__wasm_payload.snap
jan-auer: Good question. It would be nice to have dedicated tests in this actor for candidates that also asserts the different states for unwind & debug.

PR #349, opened 2021-01-14
ci(periodic): Use cargo action for clippy

flub:
Giving up on this.

Run clippy with the cargo action.  The annotations from the clippy
action end up in the wrong place and github can't figure this out.
Running the cargo action still annotates via stdout plus it doesn't
run the actualy command with json output so humans can read it anyway.

#skip-changelog

### DIFF
--- a/.github/workflows/periodic.yml
+++ b/.github/workflows/periodic.yml
@@ -26,10 +26,9 @@ jobs:
           components: clippy
 
       - name: Run clippy ${{ matrix.rust }}
-        uses: actions-rs/clippy-check@v1
+        uses: actions-rs/cargo@v1
         with:
-          name: clippy-${{ matrix.rust }}-periodic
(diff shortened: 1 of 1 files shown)


flub:
Much nicer: https://github.com/getsentry/symbolicator/actions/runs/485049298

Swatinem (approved):

Swatinem:
I’m 👍 on this, but I guess @jan-auer will disagree.

Swatinem:
BTW, I think this is the ICE that we hit on nightly: https://github.com/rust-lang/rust/issues/81000

jan-auer:
>  I guess @jan-auer will disagree.

Not necessarily, but please point me in the right direction. Previously, clippy violations were annotated on the source code. This was a super useful feature, as it showed the suggested fix in line with the diff. Where is this now?

Separately, if you know I disagree, could we please hold off merging until we have reached agreement?

PR #350, opened 2021-01-15
ci: Run clippy via the cargo action

jan-auer:
Switches to the standard cargo action for running clippy lints since it
emits sane CI output while still creating annotations in GitHub.

#skip-changelog


jan-auer:
The annotation works beautifully

flub:
So we're changing all github actions to use this instead of the clippy one?

PR #351, opened 2021-01-15
test: Port Minidump and AppleCrashReport tests to Rust

jan-auer:
Ports some of the endpoint tests to Rust. Primarily, these tests are meant to
test the endpoints. For this reason, they do not use sources. Ultimately, we
need to investigate unit-testing the endpoint code without the full service in
the background.

However, the tests instanciate a full symbolicator application. With this, we
verify that middlewares and routing (via endpoint registration) are covered by
the tests.

One of the minidump tests used the microsoft symbol server. To retain this test,
it was labeled as "integration" test, but remains in the endpoint module for
now. Once we determine a better strategy for integration and E2E testing, it can
move to its more permanent location.

#skip-changelog


jan-auer:
The test seems to deadlock before shutdown.

src/endpoints/applecrashreport.rs
```diff
@@ -156,3 +156,69 @@ pub fn configure(app: ServiceApp) -> ServiceApp {
...
+    use crate::types::SymbolicationResponse;
+
+    #[tokio::test]
+    async fn test_basic() {
+        test::setup();
+
+        let service = ServiceState::create(Config::default()).unwrap();
+        let server = TestServer::with_factory(move || crate::server::create_app(service.clone()));
+
+        let file_contents = fs::read("tests/fixtures/apple_crash_report.txt").unwrap();
```
flub, line 179: Is this really safe?  AFAIK you could be running the test binary from any cwd so the only safe thing to do is `include_bytes`()`.  But I could be wrong.

src/endpoints/minidump.rs
```diff
@@ -162,3 +162,109 @@ pub fn configure(app: ServiceApp) -> ServiceApp {
...
+    }
+
+    #[test]
+    fn test_integration_microsoft() {
+        // TODO: Move this test to E2E tests
+        test::setup();
+
+        // XXX: This cannot be a `#[tokio::test]` because the service spawns a tokio runtime
+        // internally. Dropping this within another runtime created by `tokio::test` is not allowed.
+        let service = ServiceState::create(Config::default()).unwrap();
```
flub, line 214: I still don't get why this doesn't apply to the other tests?

src/endpoints/minidump.rs
```diff
@@ -162,3 +162,109 @@ pub fn configure(app: ServiceApp) -> ServiceApp {
...
+                .unwrap();
+
+            assert_eq!(response.status(), StatusCode::OK);
+
+            let body = response.text().await.unwrap();
+            let response = serde_json::from_str::<SymbolicationResponse>(&body).unwrap();
+            insta::assert_yaml_snapshot!(response);
+        });
+
+        drop(service); // ensure the service outlives the runtime.
```
flub, line 243: I suspect this comment isn't true.  I think needs to be dropped here because waiting on implicit drop makes it race with something else.  Maybe with something the test framework does to the other threads spawned by this thread or so?  I actually wonder when exactly the runtime is torn down, wonder if you tried assigning it and explicitly dropping it in various orders.  Obviously this is a massive pain because it only deadlocks on CI, so I don't mind that we don't know.

flub: Could be that dropping the `ServiceState` explicitly shuts down the procspawn pool?

flub (approved):

src/endpoints/applecrashreport.rs
jan-auer: All of our tests assume that you run the test binary via `cargo test`, and that guarantees a stable CWD.

src/endpoints/minidump.rs
jan-auer: Neither do I. My best guess is that either the runtime initializes lazily, or if no futures are spawned in the runtime its `Drop` does not acquire certain resources that lead to the drop.

src/endpoints/minidump.rs
jan-auer: There is no difference between implicit and explicit drop. My best guess is that by dropping here we ensure that there is no lexical drop inside of the runtime. We could replace this statement with just `service;` and it would still work.

PR #352, opened 2021-01-16
ref(downloader): Cleanup location utilities

jan-auer:
In preparation of upgrading the `url` crate, this PR applies a few
syntactic changes to location utilities and downloaders. There is no
change in behavior.

- Move constants to the top of the code file
- Make location utilities methods on `SourceLocation`
- For brevity, use shorter `format!` statements

#skip-changelog

### DIFF
--- a/src/services/download/s3.rs
+++ b/src/services/download/s3.rs
@@ -11,12 +11,23 @@ use futures::TryStreamExt;
 use parking_lot::Mutex;
 use rusoto_s3::S3;
 
-use super::locations::{join_prefix_location, SourceLocation};
+use super::locations::SourceLocation;
 use super::{DownloadError, DownloadStatus, ObjectFileSource, ObjectFileSourceURI};
 use crate::sources::{FileType, S3SourceConfig, S3SourceKey};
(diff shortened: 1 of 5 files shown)


src/services/download/s3.rs
```diff
@@ -11,13 +11,24 @@ use futures::TryStreamExt;
...
 
 type ClientCache = lru::LruCache<Arc<S3SourceKey>, Arc<rusoto_s3::S3Client>>;
 
+/// Maximum number of cached S3 clients.
+///
+/// This number defines the size of the internal cache for S3 clients and should be higher than
+/// expected concurrency across S3 buckets. If this number is too low, the downloader will
+/// re-authenticate between every request.
+///
+/// TODO(ja):
```
Swatinem, line 27: unrelated, but what is the TODO here?

Swatinem (approved):

src/services/download/s3.rs
jan-auer: To add the missing metrics. We had them for S3 but not for GCS.

PR #353, opened 2021-01-16
feat: Update `url` to 2.2.0

jan-auer:
Updates the URL crate to 2.2.0. This allows us to pass the URL directly to the web client without going through a second parse stage.

With version 2, the `url::percent_encoding` module is no longer exposed. Instead, we have to use `Url::path_segments_mut` to construct a URL with encoded path segments.

#skip-changelog

### DIFF
--- a/src/services/download/sentry.rs
+++ b/src/services/download/sentry.rs
@@ -115,7 +115,7 @@ impl SentryDownloader {
     ) -> Result<Vec<SearchResult>, SentryError> {
         let response = self
             .client
-            .get(query.index_url.as_str())
+            .get(query.index_url.clone())
             .header("Accept-Encoding", "identity")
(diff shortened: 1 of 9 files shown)


jan-auer:
Based on https://github.com/getsentry/symbolicator/pull/352

src/services/download/sentry.rs
```diff
@@ -115,7 +115,7 @@ impl SentryDownloader {
     ) -> Result<Vec<SearchResult>, SentryError> {
         let response = self
             .client
-            .get(query.index_url.as_str())
+            .get(query.index_url.clone())
```
Swatinem, line 118: just wondering why they won’t take a `&Url`?

Swatinem (approved):

PR #354, opened 2021-01-16
build: Update insta and yaml

jan-auer:
Insta has updated its dependency on `serde_yaml`, which ships a change in string
rendering. For this reason, this forces an update to `serde_yaml` and updates
all snapshots.

#skip-changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -533,15 +533,14 @@ dependencies = [
 
 [[package]]
 name = "console"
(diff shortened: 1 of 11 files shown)


jan-auer:
Snapshot changes do not require an update in Sentry, since we do not use symbolicator's version of serde_yaml for snapshotting.

Swatinem (approved):

PR #355, opened 2021-01-16
build: Update uncritical dependencies

jan-auer:
#skip-changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,53 +10,52 @@ readme = "README.md"
 [dependencies]
 actix = "0.7.9"
 actix-web = { version = "0.7.19", features = ["tls"], default-features = false }
-anyhow = "1.0.32"
+anyhow = "1.0.38"
 apple-crash-report-parser = "0.4.2"
-backtrace = "0.3.53"
-base64 = "0.11.0"
+backtrace = "0.3.55"
+base64 = "0.13.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -10,54 +10,53 @@ readme = "README.md"
...
+parking_lot = "0.11.1"
 pretty_env_logger = "0.4.0"
 procspawn = { version = "0.9.0", features = ["backtrace", "json"] }
-regex = "1.3.3"
+regex = "1.4.3"
 reqwest = { git = "https://github.com/jan-auer/reqwest", tag = "v0.11.0", features = ["gzip", "json", "stream", "trust-dns"] }
 rusoto_core = "0.46.0"
 rusoto_credential = "0.46.0"
 rusoto_s3 = "0.46.0"
 sentry = { version = "0.21.0", features = ["debug-images", "failure", "log"] }
```
Swatinem, line 45: followup: update this and vendor the failure integration

Swatinem (approved):

Cargo.toml
jan-auer: 0.22.0 removed the failure integration, which is why I did not run this update. We use `failure` in two places: the Sentry actix middleware and the error handling middleware. Since I was looking into replacing `actix-web` with `warp`, we can likely get rid of that entirely.

PR #356, opened 2021-01-18
build(docker): Upgrade base image to buster-slim

jan-auer:
Last time we attempted this in #112, we caused a performance regression in the
downloader.


jan-auer:
Pushed to `release/deploy-docker-buster` so that the Docker image builds.

Swatinem (approved):

jan-auer:
Quick update, since this has gone stale: Upgrading the docker image leads to a major increase in memory usage, and a seeming memory leak. After some research, it seems that `glibc` now ships an updated memory allocator (or different defaults), which may contribute to this.

In an attempt to isolate this, I've also switched the memory allocator back to `jemalloc`. However, this allocator only kicks in for allocations from Rust code. Especially openssl, used by our downloaders, will still use the system allocator if I'm not mistaken.

It's worth noting that we used to statically link the system's openssl, which also changed version between the two containers. I tried vendoring the openssl implementation (and verified that it is being used in all clients), but there was no difference.

Here are the permutations we tried so far:

- Original implementation on stretch: baseline memory behavior
- Original implementation on buster: high memory usage + rapid growth
- Vendored `openssl`: no difference
- Jemalloc on stretch: baseline memory behavior
- Jemalloc on buster: slightly lower memory usage, but the same rapid growth

Swatinem:
updated to bullseye

PR #361, opened 2021-01-22
ref: Remove all use of the actix crate

jan-auer:
Removes all use of the actix crate, specifically the `actix::spawn` function.
Since we no longer use actix web client, we can run on bare `tokio` runtimes.
This allows us to use `tokio::spawn` directly.

Additionally, this PR removes the futures test mode. It used to spawn all
futures on the same thread rather than thread pools. In Rust 1.49, capturing
test output from threads was fixed, so this is no longer necessary.

Finally, this replaces all use of `actix::ResponseFuture` with our own typedef,
until the transition to new-style futures is complete.

#skip-changelog

### DIFF
--- a/src/actors/common/cache.rs
+++ b/src/actors/common/cache.rs
@@ -13,7 +13,7 @@ use tempfile::NamedTempFile;
 
 use crate::cache::{get_scope_path, Cache, CacheKey, CacheStatus};
 use crate::types::Scope;
-use crate::utils::futures::{BoxedFuture, CallOnDrop};
+use crate::utils::futures::{spawn_compat, BoxedFuture, CallOnDrop};
 
 /// Result from [`Cacher::compute_memoized`].
 type CacheResultFuture<T, E> = BoxedFuture<Result<Arc<T>, Arc<E>>>;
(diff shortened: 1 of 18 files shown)


src/actors/common/cache.rs
```diff
@@ -285,13 +285,12 @@ impl<T: CacheItemRequest> Cacher<T> {
             // get a channel that will receive data, or they create a new channel.
             drop(remove_computation_token);
             sender.send(result).ok();
-            Ok(())
         }
         .bind_hub(Hub::new_from_top(Hub::current()));
 
         // TODO: This spawns into the arbiter of the caller. Consider more explicit resource
         // allocation here to separate CPU intensive work from I/O work.
```
flub, line 292: Could we update or remove this comment?

src/actors/common/cache.rs
```diff
@@ -285,13 +285,12 @@ impl<T: CacheItemRequest> Cacher<T> {
             // get a channel that will receive data, or they create a new channel.
             drop(remove_computation_token);
             sender.send(result).ok();
-            Ok(())
         }
         .bind_hub(Hub::new_from_top(Hub::current()));
 
         // TODO: This spawns into the arbiter of the caller. Consider more explicit resource
```
Swatinem, line 291: maybe update the comment?

Swatinem (approved):

src/actors/common/cache.rs
jan-auer: The comment still holds up, I'll replace "arbiter" with "single-threaded runtime" though.

flub (approved):

PR #362, opened 2021-01-22
ref(symbolication): Remove futures 01

flub:
#skip-changelog

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -10,10 +10,9 @@ use std::time::{Duration, Instant};
 use apple_crash_report_parser::AppleCrashReport;
 use bytes::{Bytes, IntoBuf};
 use chrono::{DateTime, TimeZone, Utc};
+use futures::channel::oneshot;
 use futures::compat::Future01CompatExt;
-use futures::future;
-use futures01::future::{Future as _, Shared};
-use futures01::sync::oneshot;
+use futures::{future, FutureExt as _};
 use parking_lot::Mutex;
 use regex::Regex;
 use sentry::protocol::SessionStatus;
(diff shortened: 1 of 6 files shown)


src/actors/symbolication.rs
```diff
@@ -340,6 +296,46 @@ impl SymbolicationActor {
...
+        }
+        None => channel.await,
+    };
+    match channel_result {
+        Ok((finished_at, response)) => {
+            metric!(timer("requests.response_idling") = finished_at.elapsed());
+            response
+        }
+        Err(futures::channel::oneshot::Canceled) => {
+            // Double capture: todo(flub): explain why and remove
```
flub, line 329: do this

src/actors/symbolication.rs
```diff
@@ -1015,6 +1011,13 @@ impl SymbolicationActor {
         self,
         request: SymbolicateStacktraces,
     ) -> Result<CompletedSymbolicationResponse, SymbolicationError> {
+        #[cfg(test)]
+        {
+            // For test_pending_request we need to yield to the executor because otherwise
+            // the entire symbolication completes without ever yielding.  But we need to
+            // test that responses.
+            crate::utils::futures::delay(std::time::Duration::from_millis(0)).await;
+        }
```
flub, line 1020: We should decided what we'd like to do.  The way the `create_symbolication_request()` currently works is it keeps the response for 90s anyway, even when it is completed and polled once.  So the test does not need to rely on a pending response once, it could check that it can request the completion more than once which would also mean the functionality is tested.

src/actors/symbolication.rs
```diff
@@ -1137,23 +1140,25 @@ impl SymbolicationActor {
...
-    pub fn get_response(
-        &self,
+    // TODO(flub): once the callers are updated to be `async fn` this can take `&self` again.
+    pub async fn get_response(
+        self,
         request_id: RequestId,
         timeout: Option<u64>,
-    ) -> ResponseFuture<Option<SymbolicationResponse>, SymbolicationError> {
+    ) -> Option<SymbolicationResponse> {
+        // TODO(flub): Check what callers do with None, might be pointless
```
flub, line 1149: do this

jan-auer (approved):

PR #365, opened 2021-01-25
ref(symbolication): Preserve and propagate errors better

flub:
The error reporting here was all truncated to "cancelled" or "timeout"
with no extra information, resulting in opaque errors like "internal
errror" with no other information.  This changes a number of functions
over to anyhow::Error with cause chainging.  Only top-level timeouts
are still propagated as SymbolicationError::Timeout with some inner
onces being mapped to anyhow::Error, which now maps into
SymbolicationError::Canceled.

Error reporting in the request future now also logs the error instead
of only capturing it in Sentry.  This allows also getting the errors
during test runs and other debugging.

#skip-changelog

### DIFF
--- a/src/actors/symbolication.rs
+++ b/src/actors/symbolication.rs
@@ -267,9 +267,9 @@ impl SymbolicationActor {
                     };
                     sentry::end_session_with_status(status);
 
-                    let ret = error.to_symbolication_response();
+                    let response = error.to_symbolication_response();
                     log::error!("Symbolication error: {:?}", anyhow::Error::new(error));
-                    ret
+                    response
                 }
             };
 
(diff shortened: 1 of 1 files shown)


Swatinem (approved):

src/actors/symbolication.rs
```diff
@@ -1315,30 +1318,23 @@ impl SymbolicationActor {
...
-                                            path.to_string_lossy().to_string(),
-                                        ),
-                                    );
-                                });
-                            }
-                        })
-                        .ok();
-                }
-                Err(kind)
+                Self::save_minidump(minidump, minidump_cache)
```
jan-auer, line 1322: Before, never saved the minidump for timeouts, and only for crashes, panics, and fatal errors in the child process.

jan-auer (approved):

PR #367, opened 2021-01-27
ref(downloaders): Introduce a NoPerm status

flub:
This introduces a NoPerm status for the download service and adopts
the downloaders to emit it when required.  For now the users ignore
this new status and map it back to NotFound as before.

- Sentry: This is an internal source, it never emits NoPerm

- S3: This is messy, authentication failures do not happen neatly and
  there is some ambiguity to the interpretation of the errors.

- GCS: Handles both authentication-time failures and access failures
  via HTTP status codes.

- HTTP: status codes are used.

- Filesystem: straight forward mapping.

### DIFF
--- a/src/services/download/s3.rs
+++ b/src/services/download/s3.rs
@@ -9,6 +9,8 @@ use std::sync::Arc;
 
 use futures::TryStreamExt;
 use parking_lot::Mutex;
+use reqwest::StatusCode;
+use rusoto_core::RusotoError;
 use rusoto_s3::S3;
 
 use super::locations::SourceLocation;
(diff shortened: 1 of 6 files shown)


src/services/download/s3.rs
```diff
@@ -131,6 +133,27 @@ impl S3Downloader {
...
+                // Forbidden error rather than a 404, so we could accidentally turn a
+                // DownloadStatus::NotFound into DownloadStatus::NoPerm.  However we also
+                // get 403 for genuine bad credentials, so we must include this.  Our docs
+                // say s3:ListBucket is optional though.
+                return Ok(DownloadStatus::NoPerm(format!(
+                    "{} {}\nIf sentry does not have s3:ListBucket permission the file could be simply not found\n{}",
+                    http_err.status.as_str(),
+                    http_err.status.canonical_reason().unwrap_or_default(),
+                    http_err.body_as_str()
+                )));
```
flub, line 155: So this entire bit is horrible and I'm not sure this is the right call.

flub:
@jan-auer So I still think this is the correct approach over using the Err variant:
1. The way this currently works is everything in the Ok variant is user-facing, everything in the Err variant are internal errors
2. Putting this in the Err variant means there would be two classes of Err variants, those that we want to propagate to the user and those that we don't
3. Having the Err variant purely internal error means you can safely use `?`
4. There is a need for internal errors, we don't want to expose e.g. filesystem errors to the users.
5. We want to be very deliberate about what we expose to the user, e.g. a generic http client error is too wide, maybe we'd be leaking some info from another customer by some bugs etc.  So we only want to expose errors that we know are from the server.

So that leaves us with a number of returns that are "valid outcomes from calling this" and these have some classifications, may as well make them a single Enum, which is the approach we're taking here.

jan-auer:
> 1. The way this currently works is everything in the Ok variant is user-facing, everything in the Err variant are internal errors
> 2. Putting this in the Err variant means there would be two classes of Err variants, those that we want to propagate to the user and those that we don't

I disagree with these points. We need to propagate all errors up and surface them in the API response, regardless of whether it's considered "user" error or "internal" error. There's no distinction from what symbolicator is concerned: It didn't get a file and needs to report on that. 

The only reason that there was a `None` (I still don't understand why we have an enum for this), is because this is like a large hashmap. The file either exists, or it doesn't. And trying to retrieve it may fail.

> 4. There is a need for internal errors, we don't want to expose e.g. filesystem errors to the users.
> 5. We want to be very deliberate about what we expose to the user, e.g. a generic http client error is too wide, maybe we'd be leaking some info from another customer by some bugs etc.  So we only want to expose errors that we know are from the server.

This is a valid concern for downstream, user-facing applications like Sentry, but not for an internal service like Symbolicator. It is currently a large pain that we do not have the precise error condition associated with the symbolication response.

flub:
> > 1. The way this currently works is everything in the Ok variant is user-facing, everything in the Err variant are internal errors
> > 2. Putting this in the Err variant means there would be two classes of Err variants, those that we want to propagate to the user and those that we don't
> 
> I disagree with these points. We need to propagate all errors up and surface them in the API response, regardless of whether it's considered "user" error or "internal" error. There's no distinction from what symbolicator is concerned: It didn't get a file and needs to report on that.

If there was some kind of internal error while downloading an object file, say it ran out of disk space and could not write the object file, then really that should be a fatal error for the entire symbolication request and the entire request should fail so that sentry will try it again on hopefully a more healthy symbolicator or once this one is healthy again.  At this point processing it degraded is not ideal.  Currently we don't do this and instead process it degraded, but that shouldn't stop us from building this.  Once the downloader can correctly identify error cases we can start propagating them better.  But I still believe it is a bad idea to mix internal errors and understood errors in the same `Err` variant and believe the current enum approach is right.

relaxolotl:
A major portion of the changes made here appear to have been added in #518, albeit the implementation differs significantly in some sections. It looks like there was a pretty important discussion about the way certain errors, such as a failure to download due to a lack of permissions should be viewed in symbolicator though.

@flub what do you think about this PR? Should the discussion above between yourself and @jan-auer maybe be moved to an issue? Do you feel that there are some changes in this PR that could and should be ported into the current implementation of the same feature?

PR #368, opened 2021-01-28
ref(downloaders): Use tokio's async filesystem operations

flub:
Now that all the downloaders run on tokio 1.0 they can and should use
the async filesystem operations provided.  Finally releasing us from
potentially blocking operations.

#skip-changelog

### DIFF
--- a/src/services/download/mod.rs
+++ b/src/services/download/mod.rs
@@ -196,7 +196,7 @@ impl DownloadService {
 /// This is common functionality used by many downloaders.
 async fn download_stream(
     source: impl Into<ObjectFileSource>,
-    mut stream: impl Stream<Item = Result<impl AsRef<[u8]>, DownloadError>> + Unpin,
+    stream: impl Stream<Item = Result<impl AsRef<[u8]>, DownloadError>>,
     destination: PathBuf,
 ) -> Result<DownloadStatus, DownloadError> {
     // All file I/O in this function is blocking!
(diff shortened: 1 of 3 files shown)


src/services/download/mod.rs
```diff
@@ -191,22 +193,25 @@ impl DownloadService {
...
 async fn download_stream(
     source: impl Into<ObjectFileSource>,
-    stream: impl Stream<Item = Result<impl AsRef<[u8]>, DownloadError>>,
+    mut stream: impl Stream<Item = Result<impl AsRef<[u8]>, DownloadError>> + Unpin,
     destination: PathBuf,
 ) -> Result<DownloadStatus, DownloadError> {
     // All file I/O in this function is blocking!
     log::trace!("Downloading from {}", source.into());
-    let mut file = File::create(&destination).map_err(DownloadError::BadDestination)?;
-    futures::pin_mut!(stream);
```
flub, line 203: also: bye bye `pin_mut!`, less wizardy, more simple things.

src/services/download/mod.rs
```diff
@@ -191,22 +193,25 @@ impl DownloadService {
 
 /// Download the source from a stream.
 ///
-/// This is common functionality used by all many downloaders.
+/// This is common functionality used by many downloaders.
 async fn download_stream(
     source: impl Into<ObjectFileSource>,
-    stream: impl Stream<Item = Result<impl AsRef<[u8]>, DownloadError>>,
+    mut stream: impl Stream<Item = Result<impl AsRef<[u8]>, DownloadError>> + Unpin,
```
jan-auer, line 199: Can we just leave the pin mut internally? It seems like an implementation detail of the function.

jan-auer (approved):

PR #370, opened 2021-01-29
ref(endpoints): Remove redundant I/O pools from functions

jan-auer:
The minidump and apple crash report endpoints pass an I/O threadpool without actually using it. Removing this allows to clean up the service state and remove I/O pools entirely from the public interface.

#skip-changelog


jan-auer:
@flub I left the downloader in the service state now as it might be useful to access it via the service state in near future rather than passing it into the other actor's constructors. It's not exposed though, so if you'd like me to remove that as well I don't mind doing so.

flub (approved):

PR #376, opened 2021-02-02
ref(deps): Upgrade jsonwebtoken crate

flub:
We used to be on a custom git version, but we should be able to use a
normal release by now.

#skip-changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -280,6 +280,12 @@ checksum = "b41b7ea54a0c9d92199de89e20e58d49f02f8e699814ef3fdf266f6f748d15c7"
 
 [[package]]
 name = "base64"
+version = "0.12.3"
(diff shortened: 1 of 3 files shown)


flub:
A note of warning, this was done in a very TDD-style: just imitate what cloud-storage-rs does with jwt until our tests passed.  So some careful rollout of this would be prudent.

jan-auer (approved):
LGTM. This is not stricter than it was before, and GCS documentation mentions that RSA keys are required. 

It would be good to validate that this parses even if the key scaffolding (`-----BEGIN PRIVATE KEY-----`) is missing, in case currently configured values rely on that.

flub:
rebased on master for CI fixes, please review this again :)

jan-auer (approved):

PR #378, opened 2021-02-06
ref(service): Remove unused I/O threadpool

jan-auer:
The service state used to create an I/O threadpool that was passed into the
objects and download services. The pool has been removed from the Download
service since, and is not used in the objects actor. Therefore, it is dead code
and can be removed.

#skip-changelog

### DIFF
--- a/src/services/mod.rs
+++ b/src/services/mod.rs
@@ -65,7 +65,6 @@ impl Service {
         let config = Arc::new(config);
 
         let cpu_pool = ThreadPool::new();
-        let io_pool = ThreadPool::new();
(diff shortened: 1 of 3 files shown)


jan-auer:
This is based on https://github.com/getsentry/symbolicator/pull/377.

Swatinem (approved):
I wonder why this was not flagged as dead code?

jan-auer:
That's an excellent question, I'm not entirely sure. `FetchFileMetaRequest` and its fields are `pub(super)`; maybe the compiler stops there with dead code detection.

PR #384, opened 2021-02-15
ref(deps): Upgrade jsonwebtoken crate

flub:
We used to be on a custom git version, but we should be able to use a
normal release by now.

#skip-changelog


jan-auer (approved):
Has been fixed in getsentry, so G2G.

PR #388, opened 2021-02-15
fix(tests): Fix a race condition in integration tests

flub:
The mechanims used to select a free port for symbolicator to listen on
was racy and lead to flaky tests due to ports being in used.  This
insteads configures symbolicator to use port 0, ensuring that its
server gets a unique port number from the kernel without races.  The
port number is then extracted from the logs printed.

To do this the simplest way involves starting a thread, in fact this
is one of the modes that subprocess.Popen.communicate() uses
internally anyway so this isn't terrible.  Forever reading line by
line once started is perhaps not great but these are just tests after
all.

#skip-changelog

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -3,7 +3,6 @@ import copy
 import json
 import os
 import re
-import socket
 import subprocess
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


untitaker (approved):
the intent was that [`SO_REUSEADDR`](https://stackoverflow.com/a/3233022) should take care of this but upon re-reading what it does I am not sure whether the os will actually identify pytest and symbolicator as "the same program that should reuse the port"

@mitsuhiko suggested to do it this way

flub:
Oh, I see.  But for that you'd presumably need to make realistic TCP connections on the port.  I'm guessing currently the kernel just never saw this socket transition to a state where it needs to linger so it makes it available right away?

flub:
Anyway, sometimes stupid code is good :)

jan-auer (approved):

PR #390, opened 2021-02-18
fix(cache): Update the mtime when it is 1h out of date

flub:
The code used to update the mtime wasn't actually updating the mtime
as nothing was ever written to the file, it was only opened.  This
pulls in a new library which seems to do a better job at this.

There is now an explicit unit test for this, before this was tried to
be covered by an extra iteration in an integration test but this was
occasionlly flaky and did not take into account the mtime needed to be
out of date for more than an hour.

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -265,9 +265,9 @@ impl Cache {
             && mtime.map(|x| x > Duration::from_secs(3600)).unwrap_or(true))
     }
 
-    /// Validates cachefile against expiration config and open a byteview on it.
+    /// Validates `cachefile` against expiration config and open a [`ByteView`] on it.
     ///
-    /// Takes care of bumping mtime.
+    /// Takes care of bumping `mtime`.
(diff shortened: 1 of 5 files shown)


Swatinem (approved):

loewenheim (approved):

untitaker (approved):

src/cache.rs
```diff
@@ -264,8 +265,9 @@ impl Cache {
             && mtime.map(|x| x > Duration::from_secs(3600)).unwrap_or(true))
     }
 
-    /// Validate cachefile against expiration config and open a byteview on it. Takes care of
-    /// bumping mtime.
+    /// Validates cachefile against expiration config and open a byteview on it.
+    ///
+    /// Takes care of bumping mtime.
```
jan-auer, line 270:
```suggestion
    /// Validates `cachefile` against expiration config and open a [`ByteView`] on it.
    ///
    /// Takes care of bumping `mtime`.
```

jan-auer (approved):

PR #392, opened 2021-02-22
ref: Unroll loop with two items

jan-auer:
Instead of using a for loop and flattening an iterator, we can simply unroll the
loop and match the two arms manually.

#skip-changelog


loewenheim (approved):

flub (approved):
Seeing this now I don't really think one version is better than the other.  But I also don't mind either.

PR #395, opened 2021-02-26
feat(download): Use circuit-breaker in sentry source

flub:
Introduces a circuit breaker to avoid hammering on sentry if
it had problems during outages. It takes a shortcut by only having one
global circuit-breaker instead of keeping one per source, because we
know all the sentry-sources will be the same.

**Todo:**

- [ ] Add tests
- [ ] Add metrics
- [ ] Consider whether this should be done on the top-level for all downloaders, indexed on the source.


flub:
@beezz @tonyo Is this configuration suitable and are we happy with it not being configurable at runtime?

PR #397, opened 2021-03-01
fix(download): Retry entire sentry download on failure

flub:
Currently we only retry failures during fetching of the headers
instead of also retrying after failures during streaming of the body.
This makes us not very resilient against sentry workers going down.

----

This is part of #395 but without adding the circuit breaker.

### DIFF
--- a/src/services/download/sentry.rs
+++ b/src/services/download/sentry.rs
@@ -118,7 +118,7 @@ impl SentryDownloader {
             .get(query.index_url.clone())
             .header("Accept-Encoding", "identity")
             .header("User-Agent", USER_AGENT)
-            .header("Authorization", format!("Bearer {}", &query.token))
+            .bearer_auth(&query.token)
             .send()
             .await?;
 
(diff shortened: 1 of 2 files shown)


src/services/download/sentry.rs
```diff
@@ -219,37 +219,60 @@ impl SentryDownloader {
...
+        source: SentryObjectFileSource,
+        destination: PathBuf,
+    ) -> Result<DownloadStatus, DownloadError> {
+        match self
+            .client
+            .get(source.url())
+            .header("User-Agent", USER_AGENT)
+            .bearer_auth(&source.source.token)
+            .send()
+            .await
```
jan-auer, line 257: _nit_: In previous code, this was assigned to a variable, which makes it nicer to read. Same, and more importantly, above with the retry block. We generally use this style more:

```rust
        let retries = future_utils::retry(|| {
            self.download_source_once(file_source.clone(), destination.clone())
        });

        match retries.await { ... }
```

src/services/download/sentry.rs
```diff
@@ -219,37 +219,60 @@ impl SentryDownloader {
...
+    async fn download_source_once(
+        &self,
+        source: SentryObjectFileSource,
+        destination: PathBuf,
+    ) -> Result<DownloadStatus, DownloadError> {
+        match self
+            .client
+            .get(source.url())
+            .header("User-Agent", USER_AGENT)
+            .bearer_auth(&source.source.token)
```
jan-auer, line 255: This changed from the previous implementation. It looks like it's not a big deal, but lets make sure that this doesn't ship an unintended change in behavior.

src/services/download/sentry.rs
```diff
@@ -219,37 +219,60 @@ impl SentryDownloader {
...
+            .send()
+            .await
+        {
             Ok(response) => {
                 if response.status().is_success() {
-                    log::trace!("Success hitting {}", download_url);
+                    log::trace!("Success hitting {}", source.url());
```
jan-auer, line 261: These calls to `url()` allocate each time. Can we keep this deduplicated as before?

jan-auer (approved):
See the comments below; generally good to go.

src/services/download/sentry.rs
flub: Yes, this uses a feature from reqwest which we didn't use yet.  I switched to it because it seems like a safer API to do the same.  It is something to watch during deployment though to make sure there's no regression (I do recall checking the code though and it seemed fine).

src/services/download/sentry.rs
flub: i always have mixed feelings about these.  Especially when variable names like `result` appear.  And also about using un-awaited futures in an async fn.  I could get used to it I guess (though hey, can you write custom clippy lints?  because that should be linted if we actually care)

I do dislike the match statement in one expression when it closes the braces/brackets on that line though like is the case on line 225.  And I guess this case is somewhat consistent if you start doing that there.

src/services/download/sentry.rs
flub: They don't duplicate allocations though.  Each branch calls it only once, and `self.client.get()` consumes anyway so you need to construct it twice in all cases.  (I originally had it in a variable as you suggested but it seemed noisier for no gain)

jan-auer: That's right. I should've put this comment on the retry closure.

`IntoUrl` from reqwest is implemented for `&str` (since it needs to construct an owned `Url` object anyway), so we could've just passed a reference by ensuring that `utils::retry` accepts an `FnMut`. Or, potentially even better, parse the URL once and then pass a clone (or reference?) to `get`.

flub: to fill in an offline conversation: `.url()` already returns `Url` type so the gains are very marginal because the retry can not be triggered by the wrong url type

PR #401, opened 2021-03-05
test: Run minidump e2e tests on CI

jan-auer:
Re-enables the minidump symbolication test that was previously locking up in CI. Also, updates the snapshots which have changed after updating symbolic.

#skip-changelog


flub (approved):

Swatinem:
I refreshed this in https://github.com/getsentry/symbolicator/pull/955

PR #402, opened 2021-03-08
build: Update workspace to fix build

jan-auer:
This fixes two issues with the workspace build:

1. Aligns the version of `symbolic` used in `symbolicator` and `symsorter`. This
	 allows to run a single `cargo update -p symbolic` to update both versions,
	 and ensures identical behavior for sorting and lookups.

2. Marks `wasm-split` as a workspace. Without, it refuses to build as cargo
	 thinks the crate is a workspace member that is not listed in the workspace
	 root. In a follow-up, we can make this a workspace member.

#skip-changelog

### DIFF
--- a/symsorter/Cargo.toml
+++ b/symsorter/Cargo.toml
@@ -14,7 +14,7 @@ regex = "1.4.3"
 serde = { version = "1.0.119", features = ["derive"] }
 serde_json = "1.0.61"
 structopt = "0.3.21"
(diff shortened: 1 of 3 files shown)


jan-auer:
I didn't mark wasm-split as a workspace member, as we'd have to add it to the Dockerfile again. If I get a bit more time, I'll reorganize this workspace to avoid that.

Swatinem (approved):

PR #403, opened 2021-03-08
BCSymbolMap support in symbolicator

flub:
~~So far adds support to symsorter, not nice code at all yet, but it works.~~


Swatinem (commented):
lgtm so far. Would be nice to extract some unrelated refactoring out of this PR

flub:
Currently no attempt is done to invalidate the symcache if one was created earlier with no BCSymbolMap but later a BCSymbolMap became available.

Swatinem (approved):

flub:
> lgtm so far. Would be nice to extract some unrelated refactoring out of this PR

This is true, the changes in the object actor are not really needed (apart from moving decompression to a utility) and only there as i was originally was trying to use the object actor.  Some changes I don't care about (the data_cache) but some I'd like to keep as I think it makes it simpler (the meta_cache changes)

flub:
PTAL, now plist parsing has moved back to symbolic.  This PR is basically ready.

Swatinem (approved):

PR #404, opened 2021-03-08
fix: Fix issue on Docker where main becomes a noop

untitaker:
Trying something out, not sure if this fixes it. https://github.com/rust-lang/cargo/issues/7982

Also added copy of symsorter into image.

#skip-changelog


jan-auer:
Thanks, @untitaker! Created #406 to fix the Rust build, and will use your approach if the current one doesn't work.

PR #405, opened 2021-03-08
fix: Revert making symsorter a workspace crate

untitaker:
We have issues with Docker images not producing functional binaries.

See also #404 

#skip-changelog


flub (approved):

jan-auer:
Rolling forward in #406, since we have to fix the Rust build anyway.

PR #406, opened 2021-03-08
fix(build): Update build order and build with latest Rust

jan-auer:
Follow-up to #398 and #400, which aligns the build sequence with the original script. The initial change lead to an empty binary.

Also switches from the `rust/slim-stretch` base image to `debian/stretch-slim` and installs Rust manually. Since the official Rust image is no longer updated, we have to install Rust ourselves to build with the latest stable compiler. This is now required after #399.

#skip-changelog

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,25 @@
 FROM getsentry/sentry-cli:1 AS sentry-cli
-FROM rust:slim-stretch AS symbolicator-build
+FROM debian:stretch-slim AS symbolicator-build
 
 WORKDIR /work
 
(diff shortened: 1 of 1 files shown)


flub (approved):
would be lovely if the commit message described what the order does here to cargo

jan-auer:
@flub I wish I knew that. Your patch doesn't look like it's doing anything different, to be honest. I'll have to investigate this further.

Swatinem (approved):

jan-auer:
Merging this to unblock Sentry builds. Please coordinate with me before deploying.

PR #407, opened 2021-03-22
ref(clippy): Use <dyn Fail> for trait

flub:
The bare trait usage syntax is deprecated.

#skip-changelog

### DIFF
--- a/src/middlewares.rs
+++ b/src/middlewares.rs
@@ -52,7 +52,7 @@ impl ApiErrorResponse {
     pub fn from_fail(fail: &dyn Fail) -> ApiErrorResponse {
         let mut messages = vec![];
 
-        for cause in Fail::iter_chain(fail) {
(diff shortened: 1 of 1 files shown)


src/middlewares.rs
```diff
@@ -52,7 +52,7 @@ impl ApiErrorResponse {
     pub fn from_fail(fail: &dyn Fail) -> ApiErrorResponse {
         let mut messages = vec![];
 
-        for cause in Fail::iter_chain(fail) {
+        for cause in <dyn Fail>::iter_chain(fail) {
```
Swatinem, line 55: btw, this is:
https://github.com/rust-lang/rust/pull/82868 (fixing https://github.com/rust-lang/rust/issues/65371)

It seemed weird at first, but does make sense. Also looks like that these will become hard errors in Rust 2021 soon: https://github.com/rust-lang/rust/pull/83213

Swatinem (approved):

PR #409, opened 2021-04-08
fix(symcache): Better error reporting for bcsymbolmap failures

flub:
This logs a single sentry error for bcsymbolmap failures.  We lose the
log entry though.  But the error contains both the full error chain as
well as more context on what went wrong with it.

To make the functions more robust it changes the file_type arg to not
take a list, this way the error logging doesn't need to take care of
this and is also more type-strict.  It has some lifetime implications
in the downloader which gets cleaned up a little bit.

#skip-changelog

### DIFF
--- a/src/services/bitcode.rs
+++ b/src/services/bitcode.rs
@@ -193,7 +193,7 @@ impl BitcodeService {
     ) -> Result<Option<BcSymbolMapHandle>, Error> {
         // First find the PList.
         let find_plist = self
-            .fetch_file_from_all_sources(uuid, &[FileType::PList], scope.clone(), sources.clone())
+            .fetch_file_from_all_sources(uuid, FileType::PList, scope.clone(), sources.clone())
             .await?;
         let plist_handle = match find_plist {
             Some(handle) => handle,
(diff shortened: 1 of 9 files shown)


src/services/bitcode.rs
```diff
@@ -225,10 +225,14 @@ impl BitcodeService {
...
         &self,
         uuid: DebugId,
-        file_type: &'static [FileType],
+        file_type: FileType,
         scope: Scope,
         sources: Arc<[SourceConfig]>,
     ) -> Result<Option<Arc<CacheHandle>>, Error> {
+        sentry::configure_scope(|scope| {
+            scope.set_tag("auxdif.debugid", uuid);
+            scope.set_extra("auxdif.file", file_type.as_ref().into());
```
Swatinem, line 234: might as well use a `tag` for all of this, thats also done here: https://github.com/getsentry/symbolicator/blob/326177214d323227ee605899596bdfcba529b5ec/src/types/mod.rs#L747-L773

That way you should also be able to search for it.

flub: my understanding is that tags are searchable, but the flipside is they're expensive.  here the file_type is either going to be "plist" or "bcsymbolmap" and never a custom name, I don't ever imagine I'd want to search for this.  The DebugId OTOH is useful to search on.  Hence my choice to not use `.set_tag()` for the file type.

Swatinem (approved):

Swatinem (approved):

PR #411, opened 2021-04-15
feat(difs): percent-encode URIs shown in DIF candidates

flub:
The S3 and GCS URIs can have keys which may need percent-encoding.
This yields equivalent but slightly nicer URIs to copy-paste.

#skip-changelog

### DIFF
--- a/src/services/download/gcs.rs
+++ b/src/services/download/gcs.rs
@@ -60,7 +60,15 @@ impl GcsRemoteDif {
 
     /// Returns the `gs://` URI from which to download this object file.
     pub fn uri(&self) -> RemoteDifUri {
-        format!("gs://{}/{}", self.source.bucket, self.key()).into()
+        Url::parse(&format!("gs://{}/", self.source.bucket))
+            .and_then(|base| base.join(&self.key()))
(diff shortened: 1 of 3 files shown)


src/services/download/gcs.rs
```diff
@@ -60,7 +60,15 @@ impl GcsRemoteDif {
...
-        format!("gs://{}/{}", self.source.bucket, self.key()).into()
+        Url::parse(&format!("gs://{}/", self.source.bucket))
+            .and_then(|base| base.join(&self.key()))
+            .map(|url| RemoteDifUri::new(url.into_string()))
+            .unwrap_or_else(|_| {
+                // All these Result-returning operations *should* be infallible and this
+                // branch should never be used.  Nevertheless, for panic-safety we default
+                // to something infallible that's also pretty correct.
+                RemoteDifUri::new(format!("gs://{}/{}", self.source.bucket, self.key()))
+            })
```
jan-auer, line 71: Since you're doing this several times, this looks like a great reusable constructor on `RemoteDifUri` with a signature like:

```rust
fn from_parts(scheme: &str, host: &str, path: &str) -> Self;
```

jan-auer (approved):

PR #412, opened 2021-04-16
Add --strip-names option to wasm-split

samsartor:
The `--strip` option of wasm-split leaves symbol names intact. Specifically, it does not remove the ["name" custom section
](https://webassembly.github.io/spec/core/appendix/custom.html#name-section). This may be desired in some cases, but it significantly increases the size of the wasm bundle and potentially leaks source info to users.

This PR adds a `--strip-names` flag to remove the "name" section as well.

I considered adding a flag (e.g. `--strip-by [NAME]`) to strip arbitrary custom sections matching NAME (which confusingly, could be "name"), but decided against it. Let me know if that would be preferable.

I haven't yet had the chance to verify that our wasm bundle still works with sentry itself after removing the name section, but I'm guessing it will. I won't merge here until that is tested, but I figured it would be good to get reviewed ASAP.

### DIFF
--- a/wasm-split/src/main.rs
+++ b/wasm-split/src/main.rs
@@ -45,6 +45,9 @@ pub struct Cli {
     /// strip the file of debug info.
     #[argh(switch, long = "strip")]
     strip: bool,
+    /// strip the file of symbol names.
+    #[argh(switch, long = "strip-names")]
+    strip_names: bool,
     /// do not print the build id.
     #[argh(switch, short = 'q', long = "quiet")]
     quiet: bool,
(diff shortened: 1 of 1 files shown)


samsartor:
I couldn't find a contributing guide for this repo, so I just did my best. It seems like there is some naming convention needed for CI? @flub what should I do to adhear to that?

flub (approved):

flub:
@samsartor CI is happy, the Danger action is trying to be smart and ask for a changelog entry if necessary (there is no changelog for wasm-split, so not needed I'd say).  However it fails because it tries to get permission to post to this PR but doesn't have access to the secret of this repo.  We should fix that :(

samsartor:
Alas, Sentry is barfing a little bit on our wasm bundle. I think I have a fix for that here: https://github.com/rustwasm/walrus/pull/209. However, this change appears to be working as intended! I'm ready for it to be merged as soon as y'all are.

jan-auer:
Thanks, overriding status checks and adding a changelog entry on master directly.

PR #414, opened 2021-04-28
fix(bcsymbolmap): downloading BCSymbolMaps from sentry

flub:
These are fixes to improve the downloading of BCSymbolMap and
associated PList files from the sentry source.  The sentry source did
not filter files by type, assuing they are all object files.  This is
no longer the case with BCSymbolMap and PList files introduced.

Related PRs:
https://github.com/getsentry/sentry/pull/25712
https://github.com/getsentry/sentry-cli/pull/952
https://github.com/getsentry/symbolic/pull/373

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/download/mod.rs
+++ b/crates/symbolicator/src/services/download/mod.rs
@@ -146,6 +146,10 @@ impl DownloadService {
     ///
     /// If the source needs to be contacted to get matching objects this may fail and
     /// returns a [`DownloadError`].
+    ///
+    /// Note that the `filetypes` argument is not more then a hint, not all source types
+    /// will respect this and they may return all DIFs matching the `object_id`.  After
+    /// downloading you may still need to filter the files.
     pub async fn list_files(
         self: Arc<Self>,
         source: SourceConfig,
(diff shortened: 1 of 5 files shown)


src/services/symbolication.rs
```diff
@@ -312,6 +312,7 @@ impl ModuleListBuilder {
         for thread in minidump_process_state.threads() {
             for frame_pair in thread.frames().windows(2) {
                 if let [prev_frame, next_frame] = frame_pair {
+                    // TODO(flub): This never marks the first frame!
```
Swatinem, line 315: what do you mean by this?

flub: whoops, this is not supposed to be part of this PR... but i suspect there is a bug here and need to check this.

Swatinem (approved):

crates/symbolicator/src/services/download/mod.rs
```diff
@@ -146,10 +146,14 @@ impl DownloadService {
...
     /// returns a [`DownloadError`].
+    ///
+    /// Note that the `filetypes` argument is not more then a hint, not all source types
+    /// will respect this and they may return all DIFs matching the `object_id`.  After
+    /// downloading you may still need to filter the files.
     pub async fn list_files(
         self: Arc<Self>,
         source: SourceConfig,
-        filetypes: &[FileType],
+        filetypes: Vec<FileType>,
```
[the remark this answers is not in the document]
flub: This was very deliberate: on line 181 you can see that this method actually spawns all the work, this forces any reference (what used to be `&[FileType]` needs to be static.  But we can't actually give it a static reference in all callers, the new bitcode service does create this list on the fly, so I resorted to a Vec as the least evil way to solve this.

Swatinem (approved):

PR #417, opened 2021-04-30
Add support for AWS credentials supplied by an ECS task's IAM role fo…

josb:
…r downloading S3 sources

This commit adds support for an optional use_container_credentials flag,
defaulting to false. If set to true, both access_key and secret_key
should NOT be specified, or left blank.
It sidesteps more complex handling of YAML
configuration data by simply detecting invalid configurations and
panic!'ing.

For details on the AWS side, see
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html.

### DIFF
--- a/src/services/download/s3.rs
+++ b/src/services/download/s3.rs
@@ -11,6 +11,9 @@ use futures::TryStreamExt;
 use parking_lot::Mutex;
 use rusoto_s3::S3;
 
+use rusoto_core::credential::ProvideAwsCredentials;
+use rusoto_core::region::Region;
+
 use super::locations::SourceLocation;
 use super::{DownloadError, DownloadStatus, RemoteDif, RemoteDifUri};
 use crate::sources::{FileType, S3SourceConfig, S3SourceKey};
@@ -88,7 +91,11 @@ impl S3Downloader {
         }
     }
 
-    fn get_s3_client(&self, key: &Arc<S3SourceKey>) -> Arc<rusoto_s3::S3Client> {
+    fn get_s3_client(
+        &self,
+        use_container_credentials: bool,
+        key: &Arc<S3SourceKey>,
+    ) -> Arc<rusoto_s3::S3Client> {
         let mut container = self.client_cache.lock();
         if let Some(client) = container.get(&*key) {
             metric!(counter("source.s3.client.cached") += 1);
@@ -96,18 +103,44 @@ impl S3Downloader {
         } else {
             metric!(counter("source.s3.client.create") += 1);
 
-            let s3 = Arc::new(rusoto_s3::S3Client::new_with(
-                self.http_client.clone(),
-                rusoto_credential::StaticProvider::new_minimal(
+            // if use_container_creds is false (the default), and either
+            // access_key or secret_key is empty/blank, we error out.
+            if !use_container_credentials && key.access_key.is_empty() || key.secret_key.is_empty()
+            {
+                panic!(
+                    "use_container_credentials=false but access_key or secret_key is empty/blank!"
+                );
+            }
+            // If use_container_credentials is true, and either
+            // access_key or secret_key are not empty/blank, we error out.
+            if use_container_credentials && !key.access_key.is_empty() || !key.secret_key.is_empty()
+            {
+                panic!("use_container_credentials=true but access_key and secret_key not both empty/blank!");
+            }
+
+            let region = key.region.clone();
+            let s3 = Arc::new(if use_container_credentials {
+                let provider = rusoto_credential::ContainerProvider::new();
+                self.create_s3_client(provider, region)
+            } else {
+                let provider = rusoto_credential::StaticProvider::new_minimal(
                     key.access_key.clone(),
                     key.secret_key.clone(),
-                ),
-                key.region.clone(),
-            ));
+                );
+                self.create_s3_client(provider, region)
+            });
 
             container.put(key.clone(), s3.clone());
             s3
         }
+    }
+
+    fn create_s3_client<P: ProvideAwsCredentials + Send + Sync + 'static>(
+        &self,
+        provider: P,
+        region: Region,
+    ) -> rusoto_s3::S3Client {
+        rusoto_s3::S3Client::new_with(self.http_client.clone(), provider, region)
     }
 
     pub async fn download_source(
@@ -120,8 +153,9 @@ impl S3Downloader {
         log::debug!("Fetching from s3: {} (from {})", &key, &bucket);
 
         let source_key = &file_source.source.source_key;
+        let use_container_credentials = file_source.source.use_container_credentials;
         let result = self
-            .get_s3_client(&source_key)
+            .get_s3_client(use_container_credentials, &source_key)
             .get_object(rusoto_s3::GetObjectRequest {
                 key: key.clone(),
                 bucket: bucket.clone(),
--- a/src/sources.rs
+++ b/src/sources.rs
@@ -145,9 +145,11 @@ pub struct S3SourceKey {
     pub region: rusoto_core::Region,
 
     /// S3 authorization key.
+    #[serde(default)]
     pub access_key: String,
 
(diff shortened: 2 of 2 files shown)


Swatinem:
Also, we sometimes get those credentials as part of the request, so we create them at runtime, where doing a `panic` is not the right way to handle this. I think credential failures are handled in another part of the code already.

josb:
Hi, thanks for the review! I made the changes you suggested, and I also implemented selecting the AWS credentials provider a little more flexible, so that adding other credentials providers down the road will be easier. I'm not actually sure what to do about the "invalid provider" case, so I left a `panic!`, but am happy to change it (print a warning message? do nothing?) if you can give some guidance.

Please let me know what you think.

crates/symbolicator/src/services/download/s3.rs
```diff
@@ -187,6 +214,8 @@ mod tests {
     const S3_BUCKET: &str = "symbolicator-test";
 
     fn s3_source_key() -> Option<S3SourceKey> {
+        let aws_credentials_provider =
+            std::env::var("SENTRY_SYMBOLICATOR_TEST_AWS_CREDENTIALS_PROVIDER").ok()?;
```
flub, line 218: As we are not testing the container provider (which is fair enough, it's up to rusoto to test that) I think getting this from the environment is not needed and it's fine to simply hardcode this.

crates/symbolicator/src/services/download/s3.rs
```diff
@@ -120,8 +146,9 @@ impl S3Downloader {
         log::debug!("Fetching from s3: {} (from {})", &key, &bucket);
 
         let source_key = &file_source.source.source_key;
+        let aws_credentials_provider = &source_key.aws_credentials_provider;
         let result = self
-            .get_s3_client(&source_key)
+            .get_s3_client(&aws_credentials_provider.to_string(), &source_key)
```
Swatinem, line 151: Since the provider is another property of the `source_key`, you don’t need to pass it in as a separate argument. Also, you do a `to_string` here and immediately taking a reference, which is needlessly allocating.

crates/symbolicator/src/services/download/s3.rs
```diff
@@ -88,28 +91,51 @@ impl S3Downloader {
...
+            log::debug!(
+                "Using AWS credentials provider: {:?}",
+                aws_credentials_provider
+            );
+            let s3 = Arc::new(match aws_credentials_provider {
+                "container" => {
+                    let provider = rusoto_credential::ContainerProvider::new();
+                    self.create_s3_client(provider, region)
+                }
+                "static" => {
```
Swatinem, line 116: IMO, this should just be the default branch, that way we would log an invalid access key error somewhere down the road. That should be more sensible than to panic.

crates/symbolicator/src/sources.rs
```diff
@@ -144,10 +148,16 @@ pub struct S3SourceKey {
     #[serde(deserialize_with = "deserialize_region")]
     pub region: rusoto_core::Region,
 
+    /// "static" (default)|"container"
+    #[serde(default = "default_aws_provider")]
+    pub aws_credentials_provider: String,
```
flub, line 153: I think this can be made a custom enum:
```
#[derive(Debug, Clone, Copy, Deserialize, Serialize)]
#[serde(rename_all = "lowercase"]
pub enum AwsCredentialProvider {
    Static,
    Container,
}
```
That way serde will handle the deserialisation and you don't have to panic because the deserialising error is already handled (it needs to be already for e.g. regions).  You can also directly implement `Default` for that and use `#[serde(default)]` on the field in here instead of having a custom default function.

As Arpad mentioned, this config can also come in per-request instead of being statically in the config file (apologies that i didn't fully point this out at first), so panicking is really not something we want to do here.

crates/symbolicator/src/services/download/s3.rs
josb: Fixed.

crates/symbolicator/src/services/download/s3.rs
josb: Done!

crates/symbolicator/src/services/download/s3.rs
josb: Done.

crates/symbolicator/src/sources.rs
josb: Done, implemented `Default` as suggested, and removed the `panic!`; instead `Static` is now the default provider value.

crates/symbolicator/src/services/download/s3.rs
```diff
@@ -96,20 +99,38 @@ impl S3Downloader {
...
+            log::debug!(
+                "Using AWS credentials provider: {:?}",
+                key.aws_credentials_provider
+            );
+            let s3 = Arc::new(match key.aws_credentials_provider {
+                AwsCredentialsProvider::Container => {
+                    let provider = rusoto_credential::ContainerProvider::new();
+                    self.create_s3_client(provider, region)
+                }
+                _ => {
```
flub, line 112: An exhaustive match is preferably here, I don't think catch-all behaviour would be very future-proof.

.github/workflows/ci.yml
```diff
@@ -81,6 +81,7 @@ jobs:
         env:
           SENTRY_SYMBOLICATOR_GCS_CLIENT_EMAIL: ${{ secrets.SENTRY_SYMBOLICATOR_GCS_CLIENT_EMAIL }}
           SENTRY_SYMBOLICATOR_GCS_PRIVATE_KEY: ${{ secrets.SENTRY_SYMBOLICATOR_GCS_PRIVATE_KEY }}
+          SENTRY_SYMBOLICATOR_TEST_AWS_CREDENTIALS_PROVIDER: static
```
flub, line 84: this is no longer needed

crates/symbolicator/src/sources.rs
```diff
@@ -137,17 +137,36 @@ where
         .map_err(|e| D::Error::custom(format!("region: {}", e)))
 }
 
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+#[serde(rename_all = "lowercase")]
+pub enum AwsCredentialsProvider {
```
flub, line 142: Do you mind giving this a doc comment as well?  I'd even include a link to https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html or some other relevant AWS docs as you did in the PR description.

flub (commented):
Looks good!  Would you mind adding a changelog entry as well?

crates/symbolicator/src/services/download/s3.rs
flub: oh, i realised i just made a comment that differs from this and now realise why you did this. 😄   I think my new comment still makes sense now that we have an enum, and this comment made sense if you would be matching strings.  But enums are a better choice here in my opinion, so exhaustive matching is preferred.

crates/symbolicator/src/services/download/s3.rs
josb: True; fixed.

.github/workflows/ci.yml
josb: Removed.

crates/symbolicator/src/sources.rs
josb: Done.

josb:
I updated the PR, please let me know your thoughts. And thanks again for the detailed feedback, it was very helpful for a Rust noob like myself! It made making these changes a breeze.

Swatinem (approved):
Looks good! Thanks for seeing this through ;-)

I think the changelog check would like to see a link to this PR.

flub:
Indeed, thanks for all the updates to this PR!

josb:
Bedankt, @flub, and thank you @Swatinem for helping me land my first Rust PR! :)

PR #422, opened 2021-04-30
build: Publish symsorter and wasm_split

jan-auer:
Performs a full workspace build and adds `symsorter` and `wasm-split` to the list of published binaries. They are not stripped for so that backtraces can be enabled when needed.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -43,3 +43,5 @@ requireNames:
   - /^gh-pages.zip$/
   - /^symbolicator-Linux-x86_64$/
   - /^symbolicator-Linux-x86_64-debug.zip$/
+  - /^symsorter-Linux-x86_64$/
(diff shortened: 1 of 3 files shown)


jan-auer:
See https://github.com/getsentry/symbolicator/pull/422/checks?check_run_id=2475287855. Includes the following artifacts:

```
386K   gh-pages.zip
 19M   symbolicator-Linux-x86_64
 89M   symbolicator-Linux-x86_64-debug.zip
 99M   symsorter-Linux-x86_64
 25M   wasm-split-Linux-x86_64
```

Call for input: Should we strip symsorter and wasm-split?

flub:
> Call for input: Should we strip symsorter and wasm-split?

We don't strip sentry-cli, so let's not strip these either.

Swatinem (approved):

Issue #415, opened 2021-04-29
symsorter does not create valid `unified` structure


kristjanvalur:
### Environment
I am using a symsorter from the last revision of this repo to create a symbol server image to upload to S3.
However, the output structure is unexpected.

### Steps to Reproduce

1. sort Win64 symbols using `symsorter.exe --output d:\scratch\output2 --prefix Win64 --bundle-id 123 --ignore-errors -z  e:\git\UEPaxDei\Engine\Binaries\Win64`
2. s3 sync to s3 bucket
3. create crash and try to find symbol

### Expected Result
When I create the crash, Sentry will tell me that it cannot fynd my symbol at 
`s3://mybucket/5e/3aa57163c94b9a89e0e1d0b2ab6c081/debuginfo`
This is no wonder, because the file was saved, by symsorter, as
`D:\scratch\output2\Win64\5e\3aa571-63c9-4b9a-89e0-e1d0b2ab6c08\debuginfo`

- Note the hyphens
- Note the missing `1` at the end of the ID string.

This makes the generated tree from symsorter unusable as a source symbol store for sentry.

kristjanvalur:
I've read a lot of rust code today.  I found that the filename is created using DebugId::guid().  Found the debugid as an external resource via the `symbolic` repo.
The missing "1" appears to be the __appendix__, that sentry expects to be 1.  This is apparently not generated.
Also, DebugId doesn't have a method to format things properly.  I think symbolicator needs to do it by removing the hyphens and adding the appendix.  Need to learn some rust to make it work....

jan-auer:
Thank you, this regressed in https://github.com/getsentry/symbolicator/pull/403. We're working on a fix.

kristjanvalur:
Ah, see #424

kristjanvalur:
Hm, now I can't bulid.  I'm sure im doing something wrong, but
```
crate update
crate build --release
```
gives me 
```
 Compiling simple_asn1 v0.4.1
error[E0425]: cannot find function `kill` in crate `libc`
   --> C:\Users\kristjan\.cargo\registry\src\github.com-1ecc6299db9ec823\procspawn-0.9.0\src\proc.rs:332:27
    |
332 |                     libc::kill(pid as i32, libc::SIGKILL);
    |                           ^^^^ not found in `libc`

error[E0425]: cannot find value `SIGKILL` in crate `libc`
[...]

PR #423, opened 2021-04-30
revert: BCSymbolMap support in symsorter (#403)

jan-auer:
This patch contained a regression in path generation. It included hyphens and
also skipped the appendix for PDB files.

#skip-changelog
Fixes https://github.com/getsentry/symbolicator/issues/415


flub (approved):

kristjanvalur:
see the alternative #424

PR #424, opened 2021-04-30
Fix the generated filename to be correct _unified_ format.  Fixes #415

### LINKED ISSUE #415: symsorter does not create valid `unified` structure

kristjanvalur:
Previous version left in the hyphens and didn't add the "appendix".  Now generates filenames which fit what sentry looks for on symbol servers

### DIFF
--- a/crates/symsorter/src/difs.rs
+++ b/crates/symsorter/src/difs.rs
@@ -56,7 +56,7 @@ impl DifType {
     ///
     /// This is a relative pathname to the root of the unified symbol server layout.
     pub fn unified_path_for(&self, unified_id: &DebugId) -> PathBuf {
-        let repr = format!("{:x}", unified_id.uuid());
+        let repr = unified_id.to_string().replace("-", "");
         let path = format!("{}/{}/{}", &repr[..2], &repr[2..], self.suffix());
         PathBuf::from(path)
     }


jan-auer (commented):
Thank you for your PR!

Unfortunately, this doesn't create correct paths for PE and PDB files, since it's still missing the appendix. We're reverting our regression in https://github.com/getsentry/symbolicator/pull/423 and later-on rewrite that patch since it needs larger refactoring.

kristjanvalur:
Actually it does, the to_string() method adds the appendix.  Tested locally.

jan-auer:
Sorry, my mistake, you're right. I missed that `.to_string()` no longer only takes the UUID. Now, this still doesn't work for ELF files.

For reference, the original code computing the path is here:
https://github.com/getsentry/symbolicator/blob/d7e43cb9fbe1e92f8f7c3b3dc9bbed1329f7a7c7/crates/symsorter/src/utils.rs#L29-L35

As you can see, we need to use the `code_id` in some cases instead of the debug identifier, namely for ELF files. The difference is that for little-endian ELF files, the byte order of the UUIDs is flipped. For MachO, it can be either or, since they are all the same. For PDBs and PEs, it has to be the `debug_id`. 

That's why I also mentioned a larger refactoring, it's all pretty convoluted. Sorry, #403 shouldn't have been merged like that.

kristjanvalur:
Thanks for explaining.  Had some fun learning rust this morning and navigating your repos :), will try again now with your applied fix.

jan-auer:
And thank you for the patch :) We're also going to publish symsorter soon and increase its code quality as a result of that. Stay tuned.

PR #434, opened 2021-05-07
ref(api): Rename the PList filetype

flub:
This FileType is actually part of the public API as it can be used in
the configuration of a source to only use a source for certain
filetypes.  So we should have some meaningful name which won't be
wrong and confuse in the future without being too restrictive.
Calling it "PList" is just too generic, so that's a bad choice.

Since all other filetype use the snake_case convention there's no need
to diverge as it would only create confusion (the sourcebundle boat
has sadly sailed).

These changes are fine currently since none of this has been released
yet.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/sources.rs
+++ b/crates/symbolicator/src/sources.rs
@@ -384,11 +384,12 @@ pub enum FileType {
     /// Source bundle
     #[serde(rename = "sourcebundle")]
     SourceBundle,
-    /// PropertyList, mapping a dSYM UUID to a BCSymbolMap UUID for MachO.
-    #[serde(rename = "plist")]
-    PList,
+    /// A file mapping a dSYM UUID to a BCSymbolMap UUID for MachO.
+    ///
+    /// At the time of writing this only supports the XML PropertyList format, but this does
+    /// not need to remain so.
+    BcSymbolMapUuidMap,
     /// BCSymbolMap, de-obfuscates symbol names for MachO.
-    #[serde(rename = "bcsymbolmap")]
     BcSymbolMap,
 }
 
@@ -408,7 +409,7 @@ impl FileType {
             WasmDebug,
             Breakpad,
             SourceBundle,
-            PList,
+            BcSymbolMapUuidMap,
             BcSymbolMap,
         ]
     }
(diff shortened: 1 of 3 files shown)


crates/symbolicator/src/utils/paths.rs
```diff
@@ -357,7 +358,7 @@ fn get_unified_path(filetype: FileType, identifier: &ObjectId) -> Option<String>
         }
         FileType::Breakpad => "breakpad",
         FileType::SourceBundle => "sourcebundle",
-        FileType::PList => "plist",
+        FileType::BcSymbolMapUuidMap => "plist",
```
Swatinem, line 361: why not change it here?

crates/symbolicator/src/sources.rs
```diff
@@ -384,11 +384,12 @@ pub enum FileType {
...
-    /// PropertyList, mapping a dSYM UUID to a BCSymbolMap UUID for MachO.
-    #[serde(rename = "plist")]
-    PList,
+    /// A file mapping a dSYM UUID to a BCSymbolMap UUID for MachO.
+    ///
+    /// At the time of writing this only supports the XML PropertyList format, but this does
+    /// not need to remain so.
+    BcSymbolMapUuidMap,
     /// BCSymbolMap, de-obfuscates symbol names for MachO.
-    #[serde(rename = "bcsymbolmap")]
```
Swatinem, line 391: I think you need this serde annotation for parsing? Now your `AsRef<str>` does not match the parser.

crates/symbolicator/src/utils/paths.rs
flub: This was an excellent question and made me re-think the entire naming.  😄 

The short answer for this was that this defines the filename used in the unified symbol server layout.  I've now renamed the entire thing to be consistently `uuidmap`, with the motivation written in the `FileType::UuidMap` doc comment.  Basically the tradeoff I'm making it to allow it to evolve for MachO UUID mappings, but not for other executable formats.  I'm also decoupling the name from BCSymbolMap as it can easily be more generic if they introduce something else.

crates/symbolicator/src/sources.rs
flub: Oh, I missed that this also has an `AsRef<str>` impl, you're right, they should match.

Swatinem (approved):

PR #439, opened 2021-05-12
feat: Fine-tune symbolication metrics

Swatinem:
- Tag all metrics with the platform (based on module type) and the origin (minidump vs other)
- Collect metrics for:
  - Unparsable modules
  - Short (<5 frames) stack traces
- Add more metrics for unsymbolicated frame by their frame trust
- Collect frame metrics after scanning heuristic

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/symbolication.rs
+++ b/crates/symbolicator/src/services/symbolication.rs
@@ -1201,11 +1201,13 @@ fn is_likely_base_frame(frame: &SymbolicatedFrame) -> bool {
 }
 
 fn record_symbolication_metrics(
-    origin: String,
+    origin: StacktraceOrigin,
     metrics: StacktraceMetrics,
     modules: &[CompleteObjectInfo],
     stacktraces: &[CompleteStacktrace],
 ) {
+    let origin = origin.to_string();
+
     let platform = modules
         .first()
         .map(|m| m.raw.ty)
@@ -1390,7 +1392,7 @@ fn symbolicate_stacktrace(
     stacktrace
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Copy, Clone)]
 /// Where the Stack Traces in the [`SymbolicateStacktraces`] originated from.
 pub enum StacktraceOrigin {
     /// The stack traces came from a direct request to symbolicate.
@@ -1476,7 +1478,7 @@ impl SymbolicationActor {
         let sources = request.sources.clone();
         let scope = request.scope.clone();
         let signal = request.signal;
-        let origin = request.origin.to_string();
+        let origin = request.origin;
 
         let symcache_lookup = symcache_lookup
             .fetch_symcaches(self.symcaches, request)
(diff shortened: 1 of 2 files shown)


crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1363,25 +1415,89 @@ impl SymbolicationActor {
...
+            metric!(
+                time_raw("symbolication.unsymbolicated_scanned_frames") =
+                    metrics.unsymbolicated_scanned_frames,
+                "platform" => &platform, "origin" => &origin,
             );
-            metric!(time_raw("symbolication.scanned_frames") = metrics.scanned_frames);
             metric!(
-                time_raw("symbolication.unsymbolicated_frames") = metrics.unsymbolicated_frames
+                time_raw("symbolication.unmapped_frames") = metrics.unmapped_frames,
+                "platform" => &platform, "origin" => &origin,
```
flub, line 1499: This function already mixes runtime maintenance (creating futures and spawning them in threadpools in a specific order and nesting) and business logic.  Adding all these metrics makes it even harder to follow what is going on.
Would it be possible to move all the metrics into a `record_symbolication_metrics()` method?  Or would the signature of such a thing be unreadable as well?

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1261,17 +1379,38 @@ fn symbolicate_stacktrace(
...
-        || metrics.unmapped_frames > 0
-    {
+    if metrics.scanned_frames > 0 || metrics.unsymbolicated_frames > 0 {
         metrics.bad_traces += 1;
     }
 
     stacktrace
 }
 
+#[derive(Debug, Clone)]
```
flub, line 1393: If you make this `Copy` then... (see other comment)

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1334,6 +1476,7 @@ impl SymbolicationActor {
         let sources = request.sources.clone();
         let scope = request.scope.clone();
         let signal = request.signal;
+        let origin = request.origin.to_string();
```
flub, line 1479: If you make `StacktraceOrigin` `Copy` then you don't need to create this variable here, but instead can pass `request.origin` straight into `record_symbolication_metrics()` if you update it to take the `StacktraceOrigin` instead of `String`.  I think that would be a nice improvement.
(strictly speaking you don't even need it to be copy and can pass a reference, but you might as well make it copy)

flub (approved):
Would be nice if you could still address the passing of the origin part, but otherwise LGTM.

crates/symbolicator/src/services/symbolication.rs
Swatinem: The problem is that I can’t `move` `request` into the closure, however I now pass it as a typed enum instead of as a `String`, so thats at least a small improvement.

flub: huh, oh yes.  that requires moving entire request into the async closure, i overlooked that 😞    we're getting partial moves soon i think, though not sure that'd help here.

Issue #447, opened 2021-05-31
Allow custom AWS region


brancz:
I would like to run symbolicator against a [minio](https://min.io/) object storage, but the way the [configuration works today](https://github.com/getsentry/symbolicator/blob/b32b7a414e6dabe6fff5e3b1bd32513d48b89524/crates/symbolicator/src/sources.rs#L161), it is not possible to specify a custom AWS region. [Rusoto is generally capable of this](https://github.com/rusoto/rusoto/blob/31cf1506f9f4bd7af7a1f86ced3cec436913d518/rusoto/signature/src/region.rs#L117-L124). If agreed to allow configuring this, it could be retrofitted without breaking the configuration using [string-or-struct](https://serde.rs/string-or-struct.html).

Let me know what you think, happy to contribute!

@jan-auer

jan-auer:
@brancz absolutely!

Interestingly, the `Deserialize` implementation of `Region` mentions that they chose a tuple serialization because of a lack of "flatten", which has been added to serde since. I wonder if the intent was to flatten the region's custom attributes into the surrounding struct. 

Either way, string-or-struct deserialization sounds good and we'd be happy about a contribution. Thanks!

PR #448, opened 2021-05-31
sources: Allow configuring a custom AWS region

brancz:
This can be useful when using an s3 compatible service, such as minio or
Ceph.

This takes a slightly simpler approach than initially suggested in https://github.com/getsentry/symbolicator/issues/447, since rusoto already supports parsing custom regions from tuples, I decided not to introduce yet-another-representation.

Tested with this configuration:

```yaml
sources:
- id: minio
  type: s3
  bucket: symbolicator
  region: ['minio', 'http://minio.minio.svc.cluster.local:9000']
  access_key: minio
  secret_key: minio123
  layout: { type: "unified" }
  is_public: true
```

Closes https://github.com/getsentry/symbolicator/issues/447

@jan-auer


Swatinem (approved):
lgtm, unless @jan-auer has objections?

brancz:
I'm afraid that would exceed my rust abilities, I'm somewhat surprised that I managed to get this to work :sweat_smile: 

(I'm missing them entirely, but I think I would have been able to add a single additional test case, but I can't find tests for deserializing, so creating entirely new tests probably exceeds my current rust abilities)

flub:
heh, fair enough.  Sad we don't have an existing test, I'm just a bit concerned that changes like this could easily break existing valid configs and we should have some way of avoiding this.

Could you maybe update https://github.com/getsentry/symbolicator/blob/master/docs/api/index.md#amazon-s3-bucket to represent the current state though.  People shouldn't have to read the source to figure this out.

brancz:
Updated documentation and added a changelog entry.

flub:
I took the liberty of pushing a simple test, hope you don't mind.

It might be simple but sadly it already shows that your change breaks the previous functionality, so all existing configurations will break.  Could investigate this?

brancz:
Thank you! No problem at all, this will help me investigate alternative solutions such as the initially proposed string-or-struct type of construct.

brancz:
Ok, I tried, but it seems that this is out of my breadth :sweat_smile: ... does anyone think they might be able to take this over the finish line?

relaxolotl:
I've pushed a few changes that should make this change compatible with old configs in addition to supporting custom AWS regions; @flub @Swatinem could I trouble the both of you to take a look at the changes? They're a little bit more involved now but it's honestly just a heavily modified version of https://serde.rs/string-or-struct.html.

@brancz sorry it's taken so long to get to this. Hopefully this is the last push needed to get support for this into symbolicator proper 🤞

docs/api/index.md
```diff
@@ -132,7 +132,9 @@ issues as the SSQP protocol demands case insensitive lookups.
 - `type`: `"s3"`
 - `bucket`: the name of the S3 bucket
 - `prefix`: a path prefix to put in front of all keys (eg: `/windows`)
-- `region`: the AWS region where the bucket is located
+- `region`: the AWS region where the bucket is located. In order to use a
+  custom region, for example to use an S3 compatible service such as Ceph or
+  minio, specify a tuple like: `["custom-region-name", "http://minio-address/"]`.
```
flub, line 137: maybe also give the simple sting example for non-custom regions?

flub (approved):

flub:
nice! 👍

docs/api/index.md
relaxolotl: good idea, done!

PR #450, opened 2021-06-02
fix: Convert stackwalking/cfi loading into a fixpoint iteration

Swatinem:
Instead of two times, it will now do up to 5 stack walks, trying to load increasing CFI on each iteration.

### DIFF
--- a/crates/symbolicator/src/services/symbolication.rs
+++ b/crates/symbolicator/src/services/symbolication.rs
@@ -139,6 +139,8 @@ impl<'a> SymCacheLookupResult<'a> {
 /// this.
 #[derive(Clone, Debug)]
 struct CfiCacheModules {
+    /// We have to make sure to hold onto a reference to the CfiCacheFile,
+    /// to make sure it will not be evicted in the middle of reading it in the procspawn
     cache_files: Vec<Arc<CfiCacheFile>>,
     inner: BTreeMap<CodeModuleId, CfiModule>,
 }
@@ -160,8 +162,6 @@ impl CfiCacheModules {
     /// Extend the CacheModules with the fetched caches represented by
     /// [`CfiCacheResult`].
     fn extend(&mut self, cfi_caches: Vec<CfiCacheResult>) {
-        // make sure to hold onto a reference to the CfiCacheFile, to make sure
-        // it will not be evicted in the middle of using it
         self.cache_files.extend(
             cfi_caches
                 .iter()
@@ -1939,7 +1939,7 @@ impl SymbolicationActor {
                     .stackwalk_minidump_with_cfi(minidump.clone(), &cfi_caches)
                     .await?;
 
-                let missing_modules: Vec<_> = result
+                let missing_modules: Vec<(CodeModuleId, &RawObjectInfo)> = result
                     .referenced_modules
                     .iter()
                     .filter(|(id, _)| !cfi_caches.has_module(id))
--- a/crates/symbolicator/src/services/symbolication.rs
+++ b/crates/symbolicator/src/services/symbolication.rs
@@ -343,28 +343,9 @@ impl ModuleListBuilder {
 
     fn process_stacktraces(&mut self, stacktraces: &[RawStacktrace]) {
         for trace in stacktraces {
-            if let Some(first_frame) = trace.frames.first() {
-                let addr = first_frame.instruction_addr.0;
+            for frame in &trace.frames {
+                let addr = frame.instruction_addr.0;
                 self.mark_referenced(addr);
-            }
-            for frame_pair in trace.frames.windows(2) {
-                if let [prev_frame, next_frame] = frame_pair {
-                    let addr = next_frame.instruction_addr.0;
-                    self.mark_referenced(addr);
-
-                    if let Some(info_index) = self.find_module_index(prev_frame.instruction_addr.0)
-                    {
-                        let (info, _) = &mut self.inner[info_index];
-                        if info.unwind_status.is_none() {
-                            // We report this error once per missing module in a minidump.
-                            sentry::capture_message(
-                                "Module marked as used but not found",
-                                sentry::Level::Error,
-                            );
-                            info.unwind_status = Some(ObjectFileStatus::Missing);
-                        }
-                    }
-                }
             }
         }
     }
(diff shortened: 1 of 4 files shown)


Swatinem:
Still to do:
The `ModuleListBuilder` is now run in every iteration, so it will over-report missing modules. I think I can move that part out of the loop eventually.

Swatinem:
While updating the snapshots, I noticed the following:

The CFI status is being changed from `unused` to `missing` for some modules, some are *more* correct, as the module is being referenced in the stack trace, but obviously not being found/loaded. 
For some though, the status is being set even though they are not being referenced from a stack trace. Maybe they are being referenced by `referenced_modules`, but the stack frames referencing that module are being discarded somewhere.

I think in this case, avoiding the false negatives (previously being marked as `unused` even though they are `missing`) is a good idea even though it might introduce some false positives (they are now being marked as `missing` even though they end up not being used).

loewenheim (approved):

Swatinem:
False positive:
In the fixture which is used for *both* snapshots, we do two passes, one for the initial scan, which *includes* `rpcrt4.dll` and `dbgcore.dll`, but the *second* pass with the added CFI from the main exe does not include those anymore.
So indeed we try to load them, but they are still unused in the *final* stack walk.

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -130,76 +130,99 @@ impl<'a> SymCacheLookupResult<'a> {
 /// The CFI modules referenced by a minidump for CFI processing.
 ///
 /// This is populated from the [`CfiCacheResult`], which is the result of looking up the
-/// modules determined to be referenced by this minidump by
-/// [`SymbolicationActor::get_referenced_modules_from_minidump`].  It contains the CFI cache
+/// modules determined to be referenced by this minidump.  It contains the CFI cache
```
flub, line 133: Isn't this statement `which is the result of looking up the modules determined to be referenced by this minidump` no longer true with your refactoring?  I think you create one cache which you keep adding to during the various iterations.  But you can end up with unused items in this cache now since a later iteration might no longer think it requires them.

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -130,76 +130,99 @@ impl<'a> SymCacheLookupResult<'a> {
...
+    /// Check if the Cache already contains the module with the given `id`.
+    fn has_module(&self, id: &CodeModuleId) -> bool {
+        self.inner.contains_key(id)
+    }
+
+    /// Extend the CacheModules with the fetched caches represented by
+    /// [`CfiCacheResult`].
+    fn extend(&mut self, cfi_caches: Vec<CfiCacheResult>) {
+        // make sure to hold onto a reference to the CfiCacheFile, to make sure
+        // it will not be evicted in the middle of using it
```
flub, line 165: Why does it suddenly become this cache's job to keep these files alive?

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -130,76 +130,99 @@ impl<'a> SymCacheLookupResult<'a> {
...
+                }
+            };
+            (code_id, cfi_module)
+        });
+        self.inner.extend(iter)
+    }
+
+    /// Returns a copy of the inner Module Map that can be used for processing.
+    fn for_processing(&self) -> BTreeMap<CodeModuleId, CfiModule> {
+        self.inner.clone()
```
flub, line 212: cloning a `BTreeMap` sounds expensive?

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -218,60 +240,14 @@ impl CfiCacheModules {
...
-        }
+    /// Returns the inner Map.
+    fn into_inner(self) -> BTreeMap<CodeModuleId, CfiModule> {
         self.inner
     }
 }
 
 /// A module which was referenced in a minidump and processing information for it.
-#[derive(Debug, Serialize, Deserialize, Default)]
+#[derive(Clone, Debug, Serialize, Deserialize, Default)]
```
flub, line 250: i assume this is needed because you want to clone the entire BTreeMap that contains this.  This still sounds expensive?

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -303,52 +279,23 @@ struct ModuleListBuilder {
...
                 // this status.
-                if let Some(code_id) = code_module.id() {
+                if let Some(code_id) = id {
                     match cfi_caches.get(&code_id) {
                         None => {
-                            // If it was not picked up from by initial
-                            // get_referenced_modules nor marked as scanned during
-                            // stackwalking, it can only be unused.
-                            obj_info.unwind_status = Some(ObjectFileStatus::Unused);
+                            obj_info.unwind_status = None;
```
flub, line 298: why no longer unused?

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -371,40 +314,91 @@ impl ModuleListBuilder {
         Self { inner }
     }
 
-    /// Marks the module loaded at the given address as referenced.
-    ///
-    /// The respective module will always be included in the final list of modules.
-    pub fn mark_referenced(&mut self, addr: u64) {
+    fn find_module_index(&self, addr: u64) -> Option<usize> {
```
flub, line 317: could we have a doc comment describing why this exists?

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -371,40 +314,91 @@ impl ModuleListBuilder {
...
-        let info_index = match search_index {
+        let info_idx = match search_index {
             Ok(index) => index,
-            Err(0) => return,
+            Err(0) => return None,
             Err(index) => index - 1,
         };
 
-        let (info, marked) = &mut self.inner[info_index];
+        let (info, _) = &self.inner[info_idx];
```
flub, line 328: nit: do you mind naming this `_marked`?  I find that a little clearer

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -371,40 +314,91 @@ impl ModuleListBuilder {
...
-        if should_mark {
-            *marked = true;
+        if includes_addr {
+            Some(info_idx)
+        } else {
+            None
         }
     }
 
+    fn process_stacktraces(&mut self, stacktraces: &[RawStacktrace]) {
```
flub, line 344: doc comment please, why does it exist?  what does it do?  when do you need it?

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -371,40 +314,91 @@ impl ModuleListBuilder {
...
     /// Returns the modules list to be used in the symbolication response.
     pub fn build(self) -> Vec<CompleteObjectInfo> {
         self.inner
             .into_iter()
-            .filter(|(info, marked)| *marked || info.raw.debug_id.is_some())
-            .map(|(info, _)| info)
+            .filter(|(info, marked)| {
+                metric!(
+                    counter("symbolication.unwind_status") += 1,
+                    "status" => info.unwind_status.unwrap_or(ObjectFileStatus::Unused).name(),
```
flub, line 388: `unwrap_or_default()` also works as Unused is the default value.  And if you also put the `.map()` before the `.filter()` that is probably a more succinct  way of expressing this.

flub (commented):
I need another pass over the logic of how to collect statusses i'm afraid.

The `CfiCacheModules` went from something that was purely used to build a report post-hoc to something that now is merging that functionality with also providing and adding info during the fixpoint iterations.  Which makes things more complicated?  Haven't figured out yet if this can be simpler and more separated yet.

crates/symbolicator/src/services/symbolication.rs
Swatinem: Previously, they were moved into the stackwalking method, and that had an explicit drop at the end to keep the cache alive. This just moves this around a bit.

crates/symbolicator/src/services/symbolication.rs
Swatinem: I kind-of piggyback on the `None` to raising errors about files being missing.

crates/symbolicator/src/services/symbolication.rs
Swatinem: This is indeed unfortunate. Previously, the `CfiCacheModules` (which internally was a vec of tuples) would be serialized and sent via procspawn. Now it is the btreemap, so not too different actually.

flub (approved):

flub:
@jan-auer @Swatinem so for some reason the bot never popped up on this PR, but it should have.  Is the bot ok?  It used to work in the past.

Swatinem:
which bot are we talking about?

PR #455, opened 2021-06-07
fix: Avoid auxdif props to leak to symcache Errors

Swatinem:
Wraps fetching bcsymbolmap files in a sentry scope so its scope props do not leak out to normal symcache errors that are being captured.

#skip-changelog


flub (approved):
nice. this has been on my todo list for a while but am never getting round to it!

PR #457, opened 2021-06-07
ref: Update symbolic

Swatinem:
This updates includes bumps to a few dependencies (goblin, gimli, wasm stuff),
as well as specific fixes to extract or hand-roll CFI for apple system symbols.

#skip-changelog


flub:
I'm actually wondering if we should be skipping changelog for most of these bumps.  Like, symbolic *is* a core part of symbolicator.  Does it make sense to summarise *why* we are bumping?  E.g. what feature we wanted?  Maybe even with a pointer to the changelog of the dep?

flub (approved):

PR #463, opened 2021-06-14
ref: Update dependencies/lockfile

Swatinem:
The recent switch of symbolic to nom parsing creates a lockfile conflict, so might as well do a `cargo update`.
Apart from this, this also updates a few packages explicitly. The new `cadence` and `url` version needed some code changes to avoid using deprecated features.

#skip-changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,11 +12,11 @@ dependencies = [
  "crossbeam-channel 0.3.9",
  "failure",
  "fnv",
- "futures 0.1.29",
+ "futures 0.1.31",
  "libc",
  "log",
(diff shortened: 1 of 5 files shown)


flub:
In the past we've avoided doing a generic "update everything" and always done more targeted updates.  I think we *should* be able to rely on CI to confidently do this kind of thing.  But as you already know e.g. the statsd stuff is untested so in reality we don't really have this luxury.

I'm not entirely against this, but you'd have to do a fair amount of testing and careful deployment.  and risk outages i guess?

flub (approved):

PR #464, opened 2021-06-14
feat(symsorter): Support for the dyld shared cache

relaxolotl:
Starter implementation to add in support for macOS/iOS's dyld shared cache, so stack tracing is more useful when system libraries are involved in those platforms.

This adds in gimli's object as a dependency to make use of their logic to recognize dyld cache files. Support for this file type is isolated to `symsorter` for now instead of extending `symbolic` just to get the ball rolling - It may be good to revisit this decision in the future and consider whether it is worth adding this type in as a variant on `FileFormat`.

Unlike the existing logic that handles zip archives, failures to parse specific images from the dyld cache will not immediately exit symsorter; Instead, the goal is to collect all of the failed images and provide a sensible amount of info to diagnose why parsing failed for those files given the current scope of this PR.

The diff also contains trivial fixes to phrasing in documentation (removing redundancy, capitalization, etc) that I figured should be fine to bundle into this PR, but I'm open to pulling those out into their own dedicated PR if needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,14 @@ __pycache__
 /local.yml
 /cache/
 /symbols/
+/symbols*/
+
+# Generated symbols
+executable
+meta
+
+# log files
+*.log
 
 # mkdocs
 site
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -267,7 +267,7 @@ dependencies = [
  "cfg-if 1.0.0",
  "libc",
  "miniz_oxide 0.4.3",
- "object",
+ "object 0.22.0",
  "rustc-demangle",
  "serde",
 ]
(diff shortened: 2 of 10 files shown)


relaxolotl:
I got this working but I'm not sure at what cost. @Swatinem let's talk about these changes tomorrow; I'm actually unsure if it's really worth going through the changes extensively with the diff being the way it is right now.

relaxolotl:
I'll be putting this on hold for a little to focus on other higher priority tasks. A summary of where I'm at and what's left:

I've completed a very basic, unoptimized solution that is able to extract individual images directly from a dyld cache file. It uses object to parse and extract the contents of the image; The caveat here is that the solution involves changes to object which deviate from the project's scope and goals.

object's write API is meant to only write `MH_OBJECT` files. This poses a bit of a problem for sentry, as it needs the debug ID which `MH_OBJECT`s do not include. To be specific, the `LC_UUID` load command needs to be included in the macho emitted by symsorter, or we'll have issues uniquely identifying it in various parts of sentry and symbolicator. `MH_OBJECT`s only contain one single, unnamed load command according to the Mach-O spec.

The next step is to either extend object to support writing executables/dylibs that contain the bare minimum that we need for symbolication, or we need to create our own writer that does this in symsorter.

mstange:
I would recommend against this approach. This format really isn't designed to be pulled apart. I think it would be easier to leave it together, treat it as a "bundle" or "archive", and teach the other parts of the system to deal with bundles and to pull symbols out of this format at the point where the symbol lookup happens, and no sooner.
Alternatively, you could batch-transform a shared cache file into multiple symcache files, for consumers that want symcache files. But I would never transform it into multiple mach-O objects.

Symbol lookup in the dyld shared cache format is really quite easy. Once you have the file offset for the library you need, you can just instruct any existing mach-O parser (e.g. goblin or obect) to load the mach-O object stored at that offset, and everything should work automatically.

Issue #465, opened 2021-06-15
symsorter sorts some files into 00/000000


kristjanvalur:
### Environment

Windows, symsorter @ d7e43cb9fbe1e92f8f7c3b3dc9bbed1329f7a7c7

### Steps to Reproduce

Run symsorter on some _so files (attached)

### Expected Result

Sorts into a proper hash or discartds

### Actual Result
It assigns debug info 00/000000 which clearly is nonsensical and just causes noise.

```
UE4Editor-Niagara.pdb (dbg, x86_64) -> symbols\77/c1454123ac4eab8237e1c8459a5e771/debuginfo
UE4Editor-NiagaraEditor.pdb (dbg, x86_64) -> symbols\1b/add497584049d7a0b80e23f8177e231/debuginfo
hdf5.dll (lib, x86_64) -> symbols\00/0000000000000000000000000000000/executable
UE4Editor-AlembicHairTranslatorModule.dll (lib, x86_64) -> symbols\5b/ddd381ba3e43f2850553832419ee351/executable
UE4Editor-AlembicHairTranslatorModule.pdb (dbg, x86_64) -> symbols\5b/ddd381ba3e43f2850553832419ee351/debuginfo
UE4Editor-RemoteImportLibrary.dll (lib, x86_64) -> symbols\4d/35b928af4949658ce43a6b0aadc2821/executable
UE4Editor-DatasmithRemoteImport.dll (lib, x86_64) -> symbols\14/8b0600610b4945b1c802860b8d4e691/executable
UE4Editor-DatasmithRemoteImport.pdb (dbg, x86_64) -> symbols\14/8b0600610b4945b1c802860b8d4e691/debuginfo
UE4Editor-RemoteImportLibrary.pdb (dbg, x86_64) -> symbols\4d/35b928af4949658ce43a6b0aadc2821/debuginfo
ar.dll (lib, x86_64) -> symbols\00/0000000000000000000000000000000/executable
boost_atomic-vc141-mt-x64-1_70.dll (lib, x86_64) -> symbols\00/0000000000000000000000000000000/executable
```

[hdf5 (2).zip](https://github.com/getsentry/symbolicator/files/6654410/hdf5.2.zip)

[boost_atomic-vc141-mt-gd-x64-1_70.zip](https://github.com/getsentry/symbolicator/files/6654405/boost_atomic-vc141-mt-gd-x64-1_70.zip)

Swatinem:
[...]

PR #469, opened 2021-06-16
fix(symsorter): Don't emit files with empty debug identifiers

relaxolotl:
Files that symsorter was unable to generate a valid, nonempty debug identifier (i.e. 000...000) for were being written to disk as if they were successfully parsed and sorted. Chances are if the debug identifier is empty, then the debug info is unusable. Not emitting these saves some manual cleanup work on the user's end since they don't have to prune these directories.

Related: #465

### DIFF
--- a/crates/symsorter/src/utils.rs
+++ b/crates/symsorter/src/utils.rs
@@ -1,7 +1,7 @@
 use std::io::Cursor;
 use std::path::{Path, PathBuf};
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, bail, Result};
 use lazy_static::lazy_static;
 use regex::Regex;
 use symbolic::common::ByteView;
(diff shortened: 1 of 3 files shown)


crates/symsorter/src/utils.rs
```diff
@@ -26,31 +26,30 @@ pub fn is_bundle_id(input: &str) -> bool {
...
-        obj.debug_id().breakpad().to_string().to_lowercase()
+        let debug_id = obj.debug_id();
+        if debug_id.is_nil() {
+            Err(anyhow!("failed to generate debug identifier"))
+        } else {
+            Ok(debug_id.breakpad().to_string().to_lowercase())
+        }
     } else {
-        obj.code_id().as_ref().unwrap().as_str().to_string()
+        Ok(obj.code_id().as_ref().unwrap().as_str().to_string())
```
jan-auer, line 38: Could you check that this identifier is non-empty, too?

jan-auer (approved):
Thank you!

crates/symsorter/src/utils.rs
relaxolotl: Good catch, this also tripped me up as well when I was working on this: L30 checks this (`|| obj.code_id().is_none()`) and enters the debug ID block if `.code_id()` is empty.

I was thinking of ways to make it more obvious that this `else` block runs with a non-empty `code_id`, but anything that came to mind felt like an overcomplication. For example, a match block was one of my ideas:

```rust
match (obj.file_format(), obj.code_id()) {
    (Pe, _) 
    | (_, None) => use debug_id,
    (_, _) => use code
}
```

Thoughts, opinions?

jan-auer: Heh, you know our code better than we do. An in fact, `code_id` can also never be empty since it always returns `None` instead. This is good to go then 👍

PR #471, opened 2021-06-18
fix(sources): Find debug session from module index

Hamled:
Commit 41b838f changed the `SourceLookup::get_object_index_by_addr`
function to return a stable module index value, independent of the inner
Vec's ordering. This updates the debug sessions collection to be keyed
on the same module index.

The specific commit making the above change in #301 is 2a19481.

#skip-changelog


Swatinem (approved):

jan-auer:
I'll do the second review so @mitsuhiko doesn't have to ;)

jan-auer (approved):
Thanks!

Swatinem:
The CI failure is https://github.com/getsentry/symbolicator/pull/472, after that things should be green

PR #473, opened 2021-06-21
ref(sentry): Remove source tags from object sentry scope

flub:
The source tags are relevant to the download part, but having them
also part of the object scope means they end up in errros about the
processing like CFI errors and write errors.  Yet they have nothing to
do with them.  This means that when you search for the source tags you
currently get these unrelated errors.  So remove this.

The source tags for downloads are added in
`<FetchFileDataRequest as CacheItemRequest>::compute()` and are
unchanged here.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/objects/data_cache.rs
+++ b/crates/symbolicator/src/services/objects/data_cache.rs
@@ -94,10 +94,7 @@ impl ObjectHandle {
 impl ConfigureScope for ObjectHandle {
     fn to_scope(&self, scope: &mut ::sentry::Scope) {
         self.object_id.to_scope(scope);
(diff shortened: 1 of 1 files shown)


Swatinem (approved):
have you checked that the source is still there when logging failed downloads?

flub:
> have you checked that the source is still there when logging failed downloads?

no, this is on par with testing metrics and generally any other sentry sdk behaviour: we don't really have good ways of doing this currently.

Though this comment makes me fairly happy that we're not going to lose them:
> The source tags for downloads are added in
<FetchFileDataRequest as CacheItemRequest>::compute() and are
unchanged here.

PR #474, opened 2021-06-23
fix: Reduce download timeout to 60 seconds

tonyo:
Temporary measure because of recent issues with some of symbol servers.

#skip-changelog


jasonious (approved):
failing faster, per incident INC-17

Swatinem (approved):

flub:
going to close this, I think we concluded this is counter-productive and makes us affect more projects than we otherwise would

PR #483, opened 2021-07-07
feat(download): Measure how long it takes to download sources

relaxolotl:
This begins measuring requests and download throughput related to source-fetching in order to determine what a good set of default timeouts for such requests should be.

The changes can be broken down into three major parts:

1. The addition of `MeasureSourceDownload`, which is a modified version of `MeasureGuard` focused on source downloads
2. Logging stream throughput and duration in `download_stream` via `MeasureSourceDownload`
3. Logging the initial GET request's duration in every type of download source

Source downloads from the local filesystem have been skipped as it is unlikely that any meaningful or useful information will come from logging information about those.

Open questions and notes will be left directly on the diff.

### DIFF
--- a/crates/symbolicator/src/metrics.rs
+++ b/crates/symbolicator/src/metrics.rs
@@ -101,17 +101,6 @@ macro_rules! metric {
                 .send();
         })
     }};
-    (timer($id:expr), $block:block $(, $k:expr => $v:expr)* $(,)?) => {{
-        use $crate::metrics::_pred::*;
-        let now = Instant::now();
-        let rv = {$block};
-        $crate::metrics::with_client(|client| {
-            client.time_duration_with_tags($id, now.elapsed())
-                $(.with_tag($k, $v))*
-                .send();
-        });
-        rv
-    }};
 
     // we use statsd timers to send things such as filesizes as well.
     (time_raw($id:expr) = $value:expr $(, $k:expr => $v:expr)* $(,)?) => {{
@@ -122,7 +111,6 @@ macro_rules! metric {
                 .send();
         })
     }};
-
 }
 
 macro_rules! future_metrics {
--- a/crates/symbolicator/src/metrics.rs
+++ b/crates/symbolicator/src/metrics.rs
@@ -111,6 +111,16 @@ macro_rules! metric {
                 .send();
         })
     }};
+
+    // histograms
+    (histogram($id:expr) = $value:expr $(, $k:expr => $v:expr)* $(,)?) => {{
+        use $crate::metrics::_pred::*;
+        $crate::metrics::with_client(|client| {
+            client.histogram_with_tags($id, $value)
+                $(.with_tag($k, $v))*
+                .send();
+        })
+    }};
 }
 
 macro_rules! future_metrics {
--- a/crates/symbolicator/src/services/download/http.rs
+++ b/crates/symbolicator/src/services/download/http.rs
@@ -111,7 +111,7 @@ impl HttpDownloader {
         }
         let source = RemoteDif::from(file_source);
         let request = builder.header(header::USER_AGENT, USER_AGENT).send();
-        let request = future_utils::measure_source_download(source.source_metric_key(), request);
+        let request = super::measure_download_time(source.source_metric_key(), request);
 
         match request.await {
             Ok(request) => {
--- a/crates/symbolicator/src/services/download/http.rs
+++ b/crates/symbolicator/src/services/download/http.rs
@@ -13,7 +13,7 @@ use url::Url;
 use super::{DownloadError, DownloadStatus, RemoteDif, RemoteDifUri, SourceLocation, USER_AGENT};
 use crate::sources::{FileType, HttpSourceConfig};
 use crate::types::ObjectId;
-use crate::utils::futures as future_utils;
+use crate::utils::futures::{self as future_utils, m};
 
 /// The HTTP-specific [`RemoteDif`].
 #[derive(Debug, Clone)]
@@ -68,6 +68,14 @@ impl HttpDownloader {
 
         let download_url = file_source.url().ok();
         match retries.await {
+            Ok(DownloadStatus::NotFound) => {
+                log::debug!(
+                    "Did not fetch debug file from {:?}: {:?}",
+                    download_url,
+                    DownloadStatus::NotFound
+                );
+                Ok(DownloadStatus::NotFound)
+            }
             Ok(status) => {
                 log::debug!("Fetched debug file from {:?}: {:?}", download_url, status);
                 Ok(status)
@@ -106,7 +114,7 @@ impl HttpDownloader {
         let request = future_utils::measure_source_download(
             "service.download.download_source",
             source.source_metric_key(),
-            future_utils::m::result,
+            m::result,
             request,
         );
 
--- a/crates/symbolicator/src/services/download/http.rs
+++ b/crates/symbolicator/src/services/download/http.rs
@@ -13,7 +13,7 @@ use url::Url;
 use super::{DownloadError, DownloadStatus, RemoteDif, RemoteDifUri, SourceLocation, USER_AGENT};
 use crate::sources::{FileType, HttpSourceConfig};
 use crate::types::ObjectId;
-use crate::utils::futures::{self as future_utils, m};
+use crate::utils::futures as future_utils;
 
 /// The HTTP-specific [`RemoteDif`].
 #[derive(Debug, Clone)]
@@ -111,12 +111,7 @@ impl HttpDownloader {
         }
         let source = RemoteDif::from(file_source);
         let request = builder.header(header::USER_AGENT, USER_AGENT).send();
-        let request = future_utils::measure_source_download(
-            "service.download.download_source",
-            source.source_metric_key(),
-            m::result,
-            request,
-        );
+        let request = future_utils::measure_source_download(source.source_metric_key(), request);
 
         match request.await {
             Ok(request) => {
--- a/crates/symbolicator/src/services/download/http.rs
+++ b/crates/symbolicator/src/services/download/http.rs
@@ -62,56 +62,25 @@ impl HttpDownloader {
         file_source: HttpRemoteDif,
         destination: PathBuf,
     ) -> Result<DownloadStatus, DownloadError> {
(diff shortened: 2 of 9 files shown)


crates/symbolicator/src/metrics.rs
```diff
@@ -101,28 +101,26 @@ macro_rules! metric {
                 .send();
         })
     }};
-    (timer($id:expr), $block:block $(, $k:expr => $v:expr)* $(,)?) => {{
```
relaxolotl, line 104: The diff is a little messy here, but an unused type of timer here `metrics!(timer(..))` has been removed since it currently isn't being used anywhere in the codebase, as far as I can tell.

crates/symbolicator/src/metrics.rs
```diff
@@ -101,28 +101,26 @@ macro_rules! metric {
...
+    // histograms
+    (histogram($id:expr) = $value:expr $(, $k:expr => $v:expr)* $(,)?) => {{
         use $crate::metrics::_pred::*;
         $crate::metrics::with_client(|client| {
-            client.time_with_tags($id, $value)
+            client.histogram_with_tags($id, $value)
                 $(.with_tag($k, $v))*
                 .send();
         })
     }};
```
relaxolotl, line 123: It looks like statsd [directly supports histograms](https://docs.rs/cadence/0.25.0/cadence/trait.Histogrammed.html#tymethod.histogram_with_tags).

The `metrics!` macro has been extended to include that functionality in this PR, but it appears that symbolicator is already emitting `StacktraceMetrics` as a histogram via `metrics!(timer_raw(...))`, which uses `time_with_tags`. 

Was there an explicit decision made not to use statsd's `histogram_with_tags()` for histograms?

crates/symbolicator/src/services/download/http.rs
```diff
@@ -61,37 +61,74 @@ impl HttpDownloader {
         &self,
         file_source: HttpRemoteDif,
         destination: PathBuf,
+    ) -> Result<DownloadStatus, DownloadError> {
```
relaxolotl, line 64: This needs a bit of an explanation.

In all non-Sentry, non-local FS sources only the initial GET request is/was being retried. In contrast, downloads from Sentry sources retry both the initial GET request and the stream attempts, which is what the diff in this file attempts to mimic. See: 
https://github.com/getsentry/symbolicator/blob/919aafb2abda6f150d8904b6958ad7b3fce71757/crates/symbolicator/src/services/download/sentry.rs#L236-L314

Is there any reason why the others (HTTP, S3, GCS) aren't retrying both the initial GET request as well as the stream?

crates/symbolicator/src/utils/futures.rs
```diff
@@ -235,6 +235,102 @@ where
     }
 }
 
+/// A guard to [`measure`] the amount of time it takes to download a source. This guard is also
```
relaxolotl, line 238: This was "heavily inspired" by `MeasureGuard` a few lines above in the same file. As far as I can tell, fairly invasive changes would need to be made to the existing `MeasureGuard` to work for the specific use case this PR needed. 

Some examples: I was unable to figure out how to adding an arbitrary number of tags to the metric it logs. It also doesn't support histograms out of the box. Throughput logging would have involved even more extensive changes to the struct. 

Given what was mentioned above, I opted to just use it as a loose template for a new type of `MeasureGuard` for source downloads.

crates/symbolicator/src/utils/futures.rs
```diff
@@ -235,6 +235,102 @@ where
...
+/// capable of calculating and reporting the the throughput of the connection. Two metrics are
```
Swatinem, line 239:
```suggestion
/// capable of calculating and reporting the throughput of the connection. Two metrics are
```

crates/symbolicator/src/utils/futures.rs
```diff
@@ -235,6 +235,102 @@ where
...
+    }
+
+    /// A checked add to the amount of bytes transferred during the download.
+    ///
+    /// This value will be emitted when the download's future is completed or cancelled.
+    pub fn add_bytes_transferred(&mut self, additional_bytes: u64) {
+        self.bytes_transferred = self
+            .bytes_transferred
+            .and_then(|old_count| old_count.checked_add(additional_bytes).or(Some(old_count)))
+            .or(Some(additional_bytes));
```
Swatinem, line 273: `Option` has `get_or_insert_default` https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert_default which might be super useful here.
Also, I wouldn’t worry about using `checked_add` here.

Swatinem (approved):
lgtm!
I guess retrying the whole download vs only the initial connection is fine.
However I can’t really answer your question regarding the time_raw vs histogram.

crates/symbolicator/src/utils/futures.rs
```diff
@@ -235,6 +235,102 @@ where
...
+///
+/// This function reports a single metric corresponding to the task name. This metric is reported
+/// regardless of the future's return value.
+///
+/// The metric is tagged with a status derived with the `get_status` function. See the [`m`] module
+/// for status helpers.
+///
+/// An additional tag for the source name is also added to the metric.
+pub fn measure_source_download<'a, S, F>(
+    task_name: &'a str,
```
flub, line 317: I think this is the same in all callers as well, so you could also remove this parameter?

crates/symbolicator/src/services/download/http.rs
relaxolotl: I've removed retry-related logic in this PR to keep it focused. A new PR branched off of the changes in this one has been opened which extends retrying on both the header and the stream to all sources: https://github.com/getsentry/symbolicator/pull/485. It's a pretty naive approach right now, but I'd love your feedback on the diff there as well @flub.

crates/symbolicator/src/utils/futures.rs
relaxolotl: Done: I've moved `MeasureSourceDownload` to `services::download` and renamed it to an even wordier `MeasureSourceDownloadGuard`. Per your other suggestions the API has also been slimmed down to remove some of the unneeded flexibility inherited from the original `MeasureGuard` that isn't needed for this particular measure guard's use case.

crates/symbolicator/src/utils/futures.rs
relaxolotl: Thank you for this! I've done something a little different here: `get_or_insert(Default::default())` is being used over the suggeted `get_or_insert_default()` since the latter method hasn't hit stable yet.

Could I bug you to elaborate on why you suppose `checked_add` isn't needed? For now I've opted to use `saturated_add` courtesy of Floris's suggestion, but perhaps I'm being overly cautious here when I don't need to be.

crates/symbolicator/src/utils/futures.rs
```diff
@@ -235,6 +235,102 @@ where
...
+impl Drop for MeasureSourceDownload<'_> {
+    fn drop(&mut self) {
+        let status = match self.state {
+            MeasureState::Pending => "canceled",
+            MeasureState::Done(status) => status,
+        };
+
+        let duration = self.creation_time.elapsed().as_secs();
+        metric!(
+            histogram(self.task_name) = duration,
```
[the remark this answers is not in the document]
relaxolotl: Spoke to @flub about this on the side, and it turns out I misunderstood things here. It looks like timers (`metric!(timer(...))`) under-the-hood are simply histograms, so it makes more sense to just use that here, instead of shoving the duration into a histogram. The diff has been updated to do exactly that.

relaxolotl:
Additional changes worth noting:
- I've changed the tag names on the new metrics being recorded to better match the existing ones based off of a quick skim through Datadog:
  - Initial GET: "source.download.duration"
  - Stream duration: "source.stream.duration"
  - Stream throughput: "source.stream.throughput"
 - The guard currently takes a task name and simply appends ".duration" or ".throughput" when reporting their respective metrics. Users of the guard are simply expected to just provide root tag that doesn't mention exactly what's being mentioned.

crates/symbolicator/src/services/download/mod.rs
```diff
@@ -199,23 +200,142 @@ async fn download_stream(
...
+            creation_time: Instant::now(),
+        }
+    }
+
+    /// A checked add to the amount of bytes transferred during the download.
+    ///
+    /// This value will be emitted when the download's future is completed or cancelled.
+    pub fn add_bytes_transferred(&mut self, additional_bytes: u64) {
+        let bytes = self.bytes_transferred.get_or_insert(Default::default());
+        *bytes = bytes.saturating_add(additional_bytes);
```
Swatinem, line 274: Might as well use `get_or_insert(0)`, since its a numeric value.

Since this is bytes that we add, we will overflow storage sooner than wrapping around a `u64`. But the `saturating_add` is a lot easier to read than the `checked_add` you had before, that was my main concern.

crates/symbolicator/src/services/download/mod.rs
```diff
@@ -199,23 +200,142 @@ async fn download_stream(
...
+/// A tag with the source name is also added to the metric, in addition to a tag recording the
+/// status of the future.
+fn measure_download_time<'a, F, T, E>(
+    source_name: &'a str,
+    f: F,
+) -> impl Future<Output = F::Output> + 'a
+where
+    F: 'a + Future<Output = Result<T, E>>,
+{
+    let guard = MeasureSourceDownloadGuard::new("source.download", source_name);
```
Swatinem, line 331: Maybe name this `source.connect`? because `download = connect + stream` :-D

Swatinem (approved):

crates/symbolicator/src/services/download/mod.rs
relaxolotl: Oh! Good point, it is done. What do you think about adding a common segment for those two as well so it's easy to know that they're related? ie `source.download.connect` and `source.download.stream`?

relaxolotl: Actually I'll make that change and we can update the tag later if there are any issues with it

relaxolotl:
Just to be safe: @flub you left quite a bit of feedback. Are the newer changes in line with what you were thinking of?

flub (approved):
LGTM, apologies for my reviewing failures yesterday!

PR #492, opened 2021-07-13
feat(config): Add tests for config values

relaxolotl:
Adds some tests for symbolicator's config focused on download timeouts. Also includes some nitpicky changes to the changelog and documentation.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,9 @@
 - Update symbolic to correctly apply bcsymbolmaps to symbol and filenames coming from DWARF. ([#479](https://github.com/getsentry/symbolicator/pull/479))
 - Update symbolic to correctly restore registers when using compact unwinding. ([#487](https://github.com/getsentry/symbolicator/pull/487))
 - Make various timeouts related to downloading files configurable. ([#482](https://github.com/getsentry/symbolicator/pull/482), [#489](https://github.com/getsentry/symbolicator/pull/489), [#491](https://github.com/getsentry/symbolicator/pull/491))
-- Introduced the `max_download_timeout` config setting. ([#482](https://github.com/getsentry/symbolicator/pull/482), [#489](https://github.com/getsentry/symbolicator/pull/489))
-- Introduced the `streaming_timeout` config setting. ([#489](https://github.com/getsentry/symbolicator/pull/489))
-- Introduced the `connect_timeout` config setting. ([#491](https://github.com/getsentry/symbolicator/pull/491))
+- Introduced the `max_download_timeout` config setting for source downloads. ([#482](https://github.com/getsentry/symbolicator/pull/482), [#489](https://github.com/getsentry/symbolicator/pull/489))
+- Introduced the `streaming_timeout` config setting for source downloads. ([#489](https://github.com/getsentry/symbolicator/pull/489))
+- Introduced the `connect_timeout` config setting for source downloads. ([#491](https://github.com/getsentry/symbolicator/pull/491))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,9 @@
 - Update symbolic to correctly apply bcsymbolmaps to symbol and filenames coming from DWARF. ([#479](https://github.com/getsentry/symbolicator/pull/479))
 - Update symbolic to correctly restore registers when using compact unwinding. ([#487](https://github.com/getsentry/symbolicator/pull/487))
 - Make various timeouts related to downloading files configurable. ([#482](https://github.com/getsentry/symbolicator/pull/482), [#489](https://github.com/getsentry/symbolicator/pull/489), [#491](https://github.com/getsentry/symbolicator/pull/491))
+- Introduced the `max_download_timeout` config setting. ([#482](https://github.com/getsentry/symbolicator/pull/482), [#489](https://github.com/getsentry/symbolicator/pull/489))
```
relaxolotl, line 13: There was something I wasn't too sure about regarding this config value. It looks like symbolicator hasn't been released in about a month; I assume this means that nobody in production or on-prem have started to use `download_timeout` in their configs, is that correct?

crates/symbolicator/src/config.rs
```diff
@@ -420,6 +420,44 @@ mod tests {
...
+        "#;
+        let cfg = Config::from_reader(yaml.as_bytes()).unwrap();
+        let default_cfg = Config::default();
+        assert_eq!(cfg.max_download_timeout, default_cfg.max_download_timeout);
+        assert_eq!(cfg.connect_timeout, default_cfg.connect_timeout);
+        assert_eq!(cfg.streaming_timeout, default_cfg.streaming_timeout);
+    }
+
+    #[test]
+    fn test_zero_second_dl_timeouts() {
```
relaxolotl, line 436: I'd like to double check to make sure this is actually intentional or desired. If I understand things (ie tokio's timeout implementation) correctly, these timeout ASAP. Do we want to allow this? 

The alternative would be to set these to their default values if they're initialized to 0s, but it does open up the question of how low timeouts are allowed to be in a config before we start replacing them with the default value. Allowing 0 here does let us dodge this problem.

crates/symbolicator/src/config.rs
```diff
@@ -420,6 +420,44 @@ mod tests {
...
+            streaming_timeout: 0s
+        "#;
+        let cfg = Config::from_reader(yaml.as_bytes()).unwrap();
+        assert_eq!(cfg.max_download_timeout, Duration::from_secs(0));
+        assert_eq!(cfg.connect_timeout, Duration::from_secs(0));
+        assert_eq!(cfg.streaming_timeout, Duration::from_secs(0));
+    }
+
+    #[test]
+    fn test_null_dl_timeouts() {
```
relaxolotl, line 450: Should these timeouts actually be optional like the cache timeouts in case somebody does want disable download timeouts? If not, I can remove this test. If we do want them, I can open up a follow-up PR that adds that functionality in.

crates/symbolicator/src/config.rs
```diff
@@ -431,4 +469,12 @@ mod tests {
         let cfg = Config::from_reader(yaml.as_bytes());
         assert!(cfg.is_ok());
     }
+
+    #[test]
+    fn test_empty_file() {
```
relaxolotl, line 474: Is this what we actually want to do? If an empty file is passed in, should we just use the defaults or is it too much of a pain to detect when the reader is empty?

Swatinem (approved):
Since this whole thing started by downloads locking up symbolicator, I would not allow disabling the timeouts alltogether. Also, I wouldn’t worry about misconfiguration. If you set it to `0s`, well meh.

PR #496, opened 2021-07-16
feat(download): Set default timeout for a download's HEAD request to 15s

relaxolotl:
See title. This obviously isn't final, but based on about a week's worth of data so far it seems like lowering the timeout to 15s should be reasonable and allow the 95th percentile of requests to complete without timing out.


flub:
So this will result in about 5% of requests starting to time out?  That's still a lot... I think we should aim for more than that.  Do we have any idea of the health of those that we'd be timing out?  I mean, are they really bad sources or are they just fine but a bit slow at times?

flub (approved):
For the record, the slowest successful connect we've seen in a weeks time is about 11 seconds.  The average is around 100-200ms.

jan-auer:
Can we set this to somewhere < 15 seconds? The timeout the former web client for connecting was at 5s. It's totally fine to time out as long as the requests are retried in principle.

relaxolotl:
This was on me; @flub and I looked at some of the metrics yesterday and the timeout was lowered to 15s. Thank you for updating the description!

PR #498, opened 2021-07-19
ref: Increase internal download concurrency

jan-auer:
#skip-changelog

### DIFF
--- a/crates/symbolicator/src/server.rs
+++ b/crates/symbolicator/src/server.rs
@@ -24,7 +24,7 @@ pub fn run(config: Config) -> Result<()> {
     metric!(counter("server.starting") += 1);
 
     let runtime = tokio::runtime::Builder::new_multi_thread()
-        .worker_threads(1)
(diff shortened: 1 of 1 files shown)


flub (approved):

flub:
We should merge this to master: https://app.datadoghq.com/dashboard/gbx-2ai-7ud/symbolicator?fullscreen_end_ts=1626766845972&fullscreen_paused=false&fullscreen_section=overview&fullscreen_start_ts=1626594045972&fullscreen_widget=8253687310346696&utc_override=true&from_ts=1626592967244&to_ts=1626765767244&live=true

The metrics clearly show that this is a bottleneck during normal operation, even outside of an incident and during the time this was deployed this situation was much better.

We may want to consider making it configurable.  I'm not sure if that's worth it.

jan-auer:
Merging as is for now, eventually we'll move to a default thread count that uses the number of cores.

PR #499, opened 2021-07-19
Revert "ref(download): Retry downloads to S3/GCS/HTTP/local FS sources (#485)"

relaxolotl:
This reverts commit f8ac70506666fc5633f3416d8ba4b8d5ae6fbef0.

#skip-changelog


flub:
I'm not convinced of this.  Large files would have timed out with max_download_duration during their first attempt and not been retried so not resulted in much worse behaviour.

I'd much rather see metrics around this:
* A count of number of downloads tagged by:
  * source type
  * result (ok, cancelled, write, ...)
  * attempt count (1, 2, 3)

And I'd also like to see sentry.capture_error inside the retries.

There must be a better place to write this, but it's here for now 😄

relaxolotl:
Closing this in favour of waiting to see if there are metrics that demonstrate that download retries are causing increased CPU load.

PR #500, opened 2021-07-20
 feat(ci): Run Sentry's Symbolicator integration tests

armenzg:
When we make changes to Symbolicator, we should test against the latest Sentry. Up until now, we would have to merge the change to Symbolicator's `master` branch and then open a PR on Sentry to see if anything would break.

This change ensures that every PR and `master` change is tested against the latest Sentry.

#skip-changelog


armenzg:
I believe this is working:
https://github.com/getsentry/symbolicator/runs/3116790297

.github/workflows/ci.yml
```diff
@@ -115,6 +115,69 @@ jobs:
...
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - name: Cache rust cargo artifacts
+        uses: swatinem/rust-cache@v1
+
+      - name: Build Relay
```
billyvg, line 136:
```suggestion
      - name: Build Symbolicator
```

.github/workflows/ci.yml
```diff
@@ -115,6 +115,69 @@ jobs:
...
+        run: cargo build --locked
+
+      - name: Run Relay
```
billyvg, line 139:
```suggestion
      - name: Run Symbolicator
```

.github/workflows/ci.yml
```diff
@@ -115,6 +115,69 @@ jobs:
...
+          path: ${{ steps.setup.outputs.pip-cache-dir }}
+          key: sentry-deps-${{ hashFiles('sentry/requirements**.txt') }}
+          restore-keys: sentry-deps-
+
+      - name: Setup Sentry
+        uses: ./.github/actions/setup-sentry
+        with:
+          workdir: sentry
+          snuba: true
+          kafka: true
```
billyvg, line 174: Does symbolicator need these?

armenzg: Let us try. I grabbed it from Sentry's workflow.

flub: symbolicator itself does not use kafka at all.  it only receives http requests from the `symbolicate_event` celery task.

armenzg: It definitely did not like the change and run for 6 hours:
image

armenzg: We need both services:
image

billyvg (approved):
LGTM, let's get a final approval stamp from a symbolicator owner though

flub (approved):

flub:
Looks good, thanks for doing this work for us!

PR #502, opened 2021-07-21
Fill `CacheStatus::Malformed` with more info about the exact error

relaxolotl:
Malformed is now overloaded to mean both malformed object/debug file and download error. This is a first shot at trying to add in a way to distinguish between the two. This is mostly a POC and is fairly messy, dirty code.

### DIFF
--- a/crates/symbolicator/src/cache.rs
+++ b/crates/symbolicator/src/cache.rs
@@ -283,42 +283,17 @@ impl Cache {
         //   - `b"malformed"` + b"some error description"
         //   - `b"malformedbadobject"` + b"some error description"
         //   - `b"malformeddownloaderror"` + b"some error description"
-        //   Like negative/empty, it represents a failed download and mtime represents when the
-        //   failure occurred. This covers all non-400 download errors.
+        //   Like negative/empty, it represents a failed download or a failed symcache conversion.
+        //   mtime represents when the failure occurred. This covers all non-400 download errors.
         // * ok (don't really have a name): File has any other content, mtime is used to keep track
         //   of last use.
         let metadata = path.metadata()?;
 
         log::trace!("File length: {}", metadata.len());
 
-        // TODO: there's probably a cleaner way to write the entire block below
-        let (is_malformed, is_retryable) = if metadata.len() >= MALFORMED_MARKER.len() as u64 {
-            let max_to_read = MALFORMED_MARKER.len() + DOWNLOAD_ERROR_MARKER.len();
-            let amount_to_read = max_to_read.min(metadata.len() as usize);
+        let strategy = cleanup_strategy(path)?;
 
-            let mut file = File::open(path)?;
-            let mut buf = vec![0; amount_to_read];
-            file.read_exact(&mut buf)?;
-
-            log::trace!("First {} bytes: {:?}", buf.len(), buf);
-            if buf.starts_with(&[MALFORMED_MARKER, DOWNLOAD_ERROR_MARKER].concat())
-                || buf.starts_with(&[MALFORMED_MARKER, TIMEOUT_MARKER].concat())
-            {
-                (false, true)
-            } else if buf.starts_with(&[MALFORMED_MARKER, BAD_OBJECT_MARKER].concat())
-                || buf.starts_with(MALFORMED_MARKER)
-            {
-                (true, false)
-            } else {
-                (false, false)
-            }
-        } else {
-            (false, false)
-        };
-
-        let is_negative = metadata.len() == 0;
-
-        if is_malformed {
+        if strategy == CleanupStrategy::Malformed {
             // Immediately expire malformed items that have been created before this process started.
             // See docstring of MALFORMED_MARKER
 
@@ -339,7 +314,7 @@ impl Cache {
             }
         }
 
-        let max_mtime = if is_negative || is_retryable {
+        let max_mtime = if strategy == CleanupStrategy::Negative {
             self.cache_config.retry_misses_after()
         } else {
             self.cache_config.max_unused_for()
@@ -357,8 +332,7 @@ impl Cache {
             None
         };
 
-        Ok(!is_negative
-            && !is_malformed
+        Ok(strategy == CleanupStrategy::None
             && mtime.map(|x| x > Duration::from_secs(3600)).unwrap_or(true))
     }
 
@@ -389,6 +363,62 @@ impl Cache {
             }
             None => Ok(NamedTempFile::new()?),
         }
+    }
+}
+
+/// Cleanup strategies for cache items. These aren't named after the strategies themselves right now
+/// due to how difficult it is to put a name to them. They're loosely named after the type of cache
+/// entry they should be used on instead.
+#[derive(Debug, PartialEq)]
+enum CleanupStrategy {
+    /// Clean up after it is untouched for a fixed period of time.
+    None,
+    /// Clean up after a forced cool-off period so it can be re-downloaded.
+    Negative,
+    /// Clean up after it is untouched for a fixed period of time. Immediately clean up if the item
+    /// was last touched before the process executing cleanup started.
+    Malformed,
+}
+
+/// Reads a cache item at a given path and returns the cleanup strategy that should be used
+/// for the item.
+fn cleanup_strategy(path: &Path) -> io::Result<CleanupStrategy> {
+    let metadata = path.metadata()?;
+    if metadata.len() == 0 {
+        return Ok(CleanupStrategy::Negative);
+    }
+    // Whatever it is, it's not malformed or negative so just leave it be
+    if metadata.len() < MALFORMED_MARKER.len() as u64 {
+        return Ok(CleanupStrategy::None);
+    }
+
+    let mut file = File::open(path)?;
+    let mut buf = vec![0; MALFORMED_MARKER.len()];
+    file.read_exact(&mut buf)?;
+
+    log::trace!("First {} bytes: {:?}", buf.len(), buf);
+
+    // Not malformed so just leave it be
+    if buf != MALFORMED_MARKER {
+        return Ok(CleanupStrategy::None);
+    }
+
+    // We're cheating here and working off of the knowledge that
+    // DOWNLOAD_ERROR_MARKER is the longest of all known malformed cause markers
+    let max_to_read = DOWNLOAD_ERROR_MARKER.len();
+    let left_to_read = metadata.len() as usize - MALFORMED_MARKER.len();
+    let amount_to_read = max_to_read.min(left_to_read);
+    buf = vec![0; amount_to_read];
+
+    file.read_exact(&mut buf)?;
+    log::trace!("Next {} bytes: {:?}", buf.len(), buf);
+
+    if buf.starts_with(DOWNLOAD_ERROR_MARKER) || buf.starts_with(TIMEOUT_MARKER) {
+        Ok(CleanupStrategy::Negative)
+    } else {
+        // It's either a genuinely malformed object, or we have no idea kind of
+        // malformed it is so just going to treat it like it's malformed
+        Ok(CleanupStrategy::Malformed)
     }
 }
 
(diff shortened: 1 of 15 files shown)


crates/symbolicator/src/cache.rs
```diff
@@ -26,6 +26,37 @@ use crate::types::Scope;
...
+}
+
+impl From<MalformedCause> for &[u8] {
+    fn from(val: MalformedCause) -> Self {
+        match val {
+            MalformedCause::BadObject => b"badobject",
+            MalformedCause::DownloadError => b"downloaderror",
+            // TODO: This could cause us to erase valid causes in the future if we introduce new
+            // variants. Maybe we should store the unrecognized cause inside of Unknown?
+            MalformedCause::Unknown => &[],
```
flub, line 55: You could solve this by using `MalformedCause::Other(String)` instead of `MalformedCause::Unknown`?

crates/symbolicator/src/cache.rs
```diff
@@ -26,6 +26,37 @@ use crate::types::Scope;
 /// yet.
 pub const MALFORMED_MARKER: &[u8] = b"malformed";
 
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub enum MalformedCause {
+    BadObject,
+    DownloadError,
```
flub, line 32: How bad would it be to go straight to `DownloadError(String)` (same for `BadObject`)?  That way we can even add a message.

flub (commented):
partial review only so far

crates/symbolicator/src/cache.rs
relaxolotl: So I'm a little hesitant to commit to storing a full string here for two reasons:
1. It would make reading the malformed file a little bit more difficult, and easier to game/fool since the length of the file would be variable at this point.
2. The dream would be to figure out a way to restore the underlying download or object error based off of this string, which `DownloadError(String)` doesn't really solve ): Key places where it would be useful to restore the original error when we have access to only the string read from the file would be in `SymCacheFile::parse()` and `CfiCacheModule::extend()`, both of which are changed in this diff. 

I figured a good compromise was to store the general category of error without preserving specifics via this enum. The malformed file's contents continue to be restricted to a very small, predictable set of values at the cost of losing some detail. Something to note is that this does loosely relate to and/or address the comment above as well. 

I'm open to being convinced that using `BadObject(String)`, `DownloadError(String)`, `Other(String)` might be a better idea. Alternative approaches where we could store a string that allows for a more meaningful conversion from `CacheStatus` to `ObjectFileStatus` would be very welcome, since the latter does have a variant that wraps around `DownloadError`s and whatnot as well.

crates/symbolicator/src/services/download/gcs.rs
```diff
@@ -205,6 +205,14 @@ impl GcsDownloader {
...
     }
 
+    /// Downloads a source hosted on GCS.
+    ///
+    /// # Errors
+    /// - [`GcsError::InvalidUrl`]: the url was malformed and/or could not be parsed
+    /// - [`DownloadError::Reqwest`]: unable to begin streaming the source, or the initial HEAD
+    ///   request encountered an error
+    /// - [`DownloadError::Rejected`]: the initial HEAD request received a non-successful response
+    /// - [`DownloadError::Cancelled`]: the request timed out
```
flub, line 215: in general these descriptions should probably be documented on the error types instead?  Not that i mind too much, but having this in every use might become tricky to all keep up to date.

crates/symbolicator/src/services/objects/data_cache.rs
```diff
@@ -71,7 +71,12 @@ impl ObjectHandle {
         match self.status {
             CacheStatus::Positive => Ok(Some(Object::parse(&self.data)?)),
             CacheStatus::Negative => Ok(None),
-            CacheStatus::Malformed => Err(ObjectError::Malformed),
+            CacheStatus::Malformed(MalformedCause::BadObject)
+            | CacheStatus::Malformed(MalformedCause::Unknown) => Err(ObjectError::Malformed),
+            CacheStatus::Malformed(MalformedCause::DownloadError) => {
+                // TODO: how can we restore the original error?
+                Err(ObjectError::Download(DownloadError::Canceled))
```
flub, line 78: We can't (or well, it might be possible but too complicated.  I'd try and add a message to the cache as explained above and then add a `DownloadError::Cached(String)` variant, that way we can propagate a details message.  But we don't need a full error message.  If it needed capturing into a sentry error it only needed capturing the non-cached version.

crates/symbolicator/src/services/download/gcs.rs
relaxolotl: Good point; I added all of these in because I started finding it a bit tedious constantly comparing the `download_source` implementations across all sources and seeing which errors were actually thrown by them while I was updating its consumers (`dispatch_download`, `download`, data cache itself, etc.)

I'll double check to see if all of the download error variants are consistently being used in the same spots, and move them if it doesn't differ from one source's `download_source` to another (e.g. does HTTP return `DownloadError::Request` for other things? does GCS? does S3?)

crates/symbolicator/src/cache.rs
```diff
@@ -201,24 +234,37 @@ impl Cache {
...
+        let malformed_markers = &[
+            MALFORMED_MARKER,
+            &[MALFORMED_MARKER, MalformedCause::BadObject.into()].concat(),
+            &[MALFORMED_MARKER, MalformedCause::DownloadError.into()].concat(),
+            &[MALFORMED_MARKER, MalformedCause::Unknown.into()].concat(),
+        ];
+
+        let mut is_malformed = false;
+        for marker in malformed_markers {
+            if marker.len() as u64 != metadata.len() {
```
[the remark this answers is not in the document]
relaxolotl: Would it be safe to say that Sentry exclusively controls the cache files, and they're mostly protected with the right permissions and whatnot to prevent somebody from going in and mucking around in their contents? 

My main fear is that allowing "anything" to be written and read is a pretty great way to ask for this to be exploited. Regardless, I'm thinking maybe we could just add in reasonable read/write limits as an initial preventative measure if the files are difficult to manipulate in a malicious way and go from there, thoughts?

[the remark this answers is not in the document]
relaxolotl: spoke to @flub about this last week and confirmed that cache files on sentry at least are only controlled by sentry and there exists reasonable measures to prevent unauthorized access to them. taking this and the fact that symbolicator is the only thing that's writing to these cache files, it should be safe to do basic checking and not worry too much about the contents of cache entries for malformed files.

relaxolotl:
reworked this significantly:
- rebased off of #505 
- `MalformedCause`s now wrap around a string that contains a brief description of the error
- `MalformedCause`s related to download errors use the same retry logic as negative caches
- sources cached as malformed due to download errors will now show up in the list of candidates during symbolication

crates/symbolicator/src/cache.rs
relaxolotl: see https://github.com/getsentry/symbolicator/pull/502#discussion_r677972226

crates/symbolicator/src/cache.rs
relaxolotl: it is done

crates/symbolicator/src/cache.rs
```diff
@@ -201,23 +274,42 @@ impl Cache {
...
         // * negative/empty: An empty file. Represents a failed download. mtime is used to indicate
         //   when the failed download happened (when the file was created)
-        // * malformed: A file with the content `b"malformed"`. Represents a failed symcache
-        //   conversion. mtime indicates when we attempted to convert.
+        // * malformed: A file that starts with `b"malformed"`. Its contents should be one of the
+        //   following permutations:
+        //   - `b"malformed"` + b"some error description"
+        //   - `b"malformedbadobject"` + b"some error description"
+        //   - `b"malformeddownloaderror"` + b"some error description"
+        //   Like negative/empty, it represents a failed download and mtime represents when the
```
Swatinem, line 282: as well as failed symcache conversion

crates/symbolicator/src/cache.rs
```diff
@@ -201,23 +274,42 @@ impl Cache {
...
+            let amount_to_read = max_to_read.min(metadata.len() as usize);
+
             let mut file = File::open(path)?;
-            let mut buf = vec![0; MALFORMED_MARKER.len()];
+            let mut buf = vec![0; amount_to_read];
             file.read_exact(&mut buf)?;
 
             log::trace!("First {} bytes: {:?}", buf.len(), buf);
-            buf == MALFORMED_MARKER
+            if buf.starts_with(&[MALFORMED_MARKER, DOWNLOAD_ERROR_MARKER].concat())
```
Swatinem, line 300: `.concat()` allocates, so we probably shouldn’t do it like this.

I can imagine something like this:

```
let mut is_retryable = false;
let is_malformed = buf.starts_with(...);
if is_malformed {
is_retryable = buf.starts_with(...) || buf.starts_with(...);
}
```

or similar. I think that is also easier to read than the `.concat()` variant.

crates/symbolicator/src/services/download/mod.rs
```diff
@@ -131,16 +138,15 @@ impl DownloadService {
...
-            }
             Ok(status) => {
-                log::debug!("Fetched debug file from {:?}: {:?}", source, status);
+                match status {
+                    DownloadStatus::Completed => {
+                        log::debug!("Did not fetch debug file from {:?}: {:?}", source, status);
+                    }
+                    DownloadStatus::NotFound => {
+                        log::debug!("Fetched debug file from {:?}: {:?}", source, status);
+                    }
```
Swatinem, line 148: these are switched around

crates/symbolicator/src/services/objects/data_cache.rs
```diff
@@ -337,7 +373,7 @@ mod tests {
 
             // for each of the different symbol sources, we assert that:
             // * we get a negative cache result no matter how often we try
-            // * we hit the symbol source exactly once for the initial request
+            // * we hit the symbol source once for the initial request, then another 3 times to retry
```
Swatinem, line 376: *another* 3 times, or 3 in total? I was under the impression that we try 3 times in total, so this might be wrong in a different place?

crates/symbolicator/src/services/objects/data_cache.rs
```diff
@@ -424,10 +475,17 @@ mod tests {
                 ..find_object
             };
             let result = objects_actor.find(find_object.clone()).await.unwrap();
-            assert_eq!(result.meta.unwrap().status, CacheStatus::Malformed);
+            assert_eq!(
+                result.meta.unwrap().status,
+                CacheStatus::Malformed(MalformedCause::Timeout)
+            );
+            // Timeouts don't retry ):
```
Swatinem, line 482: I think you can tweak the per-connection vs per-request timeouts to trigger a retry here

crates/symbolicator/src/services/objects/data_cache.rs
relaxolotl: It's another three times, I believe: 
https://github.com/getsentry/symbolicator/blob/1aa8a7a0aa9af4259a27a3a736a0cf874709bf6d/crates/symbolicator/src/utils/futures.rs#L290-L297

The check happens in the helper after the task is executed, so this retries 3 times in total.

crates/symbolicator/src/cache.rs
relaxolotl: the logic here is pretty complex, so i've reworked this using a slightly different approach to track what type of cache file we have (look for `CacheStrategy` in the diff).

crates/symbolicator/src/services/objects/data_cache.rs
relaxolotl: resolved via the addition of the error string into the cache status

relaxolotl:
i'm going to approach this differently: the changes in #502 and #504 will be consolidated into a series of 3-4 small PRs that will:
- gradually migrate symbolicator into a state that can distinguish between unexpected download errors and genuine malformed errors
- both malformed and download cache entries will contain some small amount of diagnostic information 
- download cache entries will use the same expiration strategy as negative cache entries.

i'm going to close this PR in favour of this new approach.

PR #505, opened 2021-07-27
feat(cache): Be less strict about the contents of malformed cache files so they're easier to extend

relaxolotl:
This makes the checking on cache files a little more lenient: as long as the file begins with the correct set of bytes then it's a valid malformed cache entry (despite its name).

This should make it easier to merge and deploy https://github.com/getsentry/symbolicator/pull/502, particularly if it goes bad: if it turns out that #502's changes need to be reverted then the cache won't be completely broken if this PR's changes are around.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/cache.rs
+++ b/crates/symbolicator/src/cache.rs
@@ -51,7 +51,7 @@ impl AsRef<str> for CacheStatus {
 
 impl CacheStatus {
     pub fn from_content(s: &[u8]) -> CacheStatus {
-        if s == MALFORMED_MARKER {
+        if s.starts_with(MALFORMED_MARKER) {
             CacheStatus::Malformed
         } else if s.is_empty() {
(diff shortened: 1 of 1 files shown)


crates/symbolicator/src/cache.rs
```diff
@@ -620,8 +620,10 @@ mod tests {
         // File has same amount of chars as "malformed", check that optimization works
         File::create(tempdir.path().join("foo/keepthis"))?.write_all(b"addictive")?;
         File::create(tempdir.path().join("foo/keepthis2"))?.write_all(b"hi")?;
+        File::create(tempdir.path().join("foo/keepthis3"))?.write_all(b"honkhonkbeepbeep")?;
```
Swatinem, line 623: honkhonk

Swatinem (approved):

loewenheim (approved):

PR #508, opened 2021-07-30
fix: Avoid serializing minidumps over procspawn

Swatinem:
Instead of passing (and serializing) a byte buffer to procspawn, this will rather create a tempfile and pass that path via procspawn.

This seems to be a small win when processing a 20M minidump, though note that I was timing `cargo run -p process-event` so timings include overhead from both cargo and uploading of the minidump.

before:
  Time (mean ± σ):      1.611 s ±  0.041 s    [User: 274.8 ms, System: 102.0 ms]
  Range (min … max):    1.548 s …  1.697 s    10 runs
after:
  Time (mean ± σ):      1.526 s ±  0.024 s    [User: 275.7 ms, System: 101.5 ms]
  Range (min … max):    1.496 s …  1.573 s    10 runs

Due to some error handling and borrowcheck shenanigans, the minidumps are now persisted in the diagnostics cache for *all* errors during processing, while previously timeouts were ignored.

### DIFF
--- a/crates/symbolicator/src/endpoints/minidump.rs
+++ b/crates/symbolicator/src/endpoints/minidump.rs
@@ -33,10 +33,13 @@ async fn handle_minidump_request(
         let content_disposition = field.content_disposition();
         match content_disposition.as_ref().and_then(|d| d.get_name()) {
             Some("upload_file_minidump") => {
-                let minidump_file = tempfile::Builder::new()
-                    .prefix("minidump")
-                    .suffix(".dmp")
-                    .tempfile()?;
+                let mut minidump_file = tempfile::Builder::new();
+                minidump_file.prefix("minidump").suffix(".dmp");
+                let minidump_file = if let Some(tmp_dir) = state.config().cache_dir("tmp") {
+                    minidump_file.tempfile_in(tmp_dir)
+                } else {
+                    minidump_file.tempfile()
+                }?;
                 let (mut file, temp_path) = minidump_file.into_parts();
 
                 read_multipart_file(field, &mut file).await?;
--- a/crates/symbolicator/src/endpoints/minidump.rs
+++ b/crates/symbolicator/src/endpoints/minidump.rs
@@ -1,12 +1,12 @@
+use std::io::Write;
+
 use actix_web::{error, multipart, App, Error, HttpMessage, HttpRequest, Json, Query, State};
 use futures::{compat::Stream01CompatExt, StreamExt};
 
 use crate::endpoints::symbolicate::SymbolicationRequestQueryParams;
 use crate::services::Service;
 use crate::types::{RequestOptions, SymbolicationResponse};
-use crate::utils::multipart::{
-    read_multipart_file, read_multipart_request_options, read_multipart_sources,
-};
+use crate::utils::multipart::{read_multipart_request_options, read_multipart_sources};
 use crate::utils::sentry::ConfigureScope;
 
 async fn handle_minidump_request(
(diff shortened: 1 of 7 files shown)


crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1939,12 +1903,49 @@ impl SymbolicationActor {
...
+                    Err(err) => {
+                        if self.diagnostics_cache.cache_dir().is_some() {
+                            match minidump_file.keep() {
+                                Ok((_file, path)) => {
+                                    sentry::configure_scope(|scope| {
+                                        scope.set_extra(
+                                            "crashed_minidump",
+                                            sentry::protocol::Value::String(
+                                                path.to_string_lossy().to_string(),
+                                            ),
```
flub, line 1934: ugh, we should make the SDK nicer to use ;)

flub (approved):

jan-auer (changes requested):
1. Rather than passing in a named temp file, pass around the `File`, if possible. The OS should still keep it open.
2. Stream into a file directly in the endpoint, then pass around the file. @flub we've had various versions of this in the past already, do you remember where this was?

flub:
> 2. Stream into a file directly in the endpoint, then pass around the file. @flub we've had various versions of this in the past already, do you remember where this was?

hmm, no i don't.  sorry

Swatinem:
now streaming directly to a file (and updating rustc), I get these timings, though I noticed they have high variability:

before:
  Time (mean ± σ):      1.694 s ±  0.028 s    [User: 288.2 ms, System: 106.9 ms]
  Range (min … max):    1.656 s …  1.749 s    10 runs
after:
  Time (mean ± σ):      1.621 s ±  0.030 s    [User: 294.3 ms, System: 110.5 ms]
  Range (min … max):    1.585 s …  1.676 s    10 runs

loewenheim (approved):

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1831,7 +1792,11 @@ impl SymbolicationActor {
 
                     // Stackwalk the minidump.
                     let cfi = load_cfi_for_processor(cfi_caches);
-                    let minidump = ByteView::from_slice(&minidump);
+                    // we cannot map an `io::Error` into `MinidumpNotFound` since there is no public
+                    // constructor on `ProcessResult`. Passing in an empty buffer should result in
+                    // the same error though.
+                    let minidump =
+                        ByteView::open(minidump_path).unwrap_or_else(|_| ByteView::from_slice(b""));
```
jan-auer, line 1799: An empty buffer will now result in a generic "failed to read miniump" error with little information for debugging.

Is there no way you can wrap the error somehow?

crates/symbolicator/src/endpoints/minidump.rs
```diff
@@ -32,17 +32,34 @@ async fn handle_minidump_request(
...
+                let (mut file, temp_path) = minidump_file.into_parts();
+
+                let mut stream = field.compat();
+                // it would have been nice to use `tokio::io::copy`, or at least the non-blocking
+                // `tokio::fs::File` / `tokio::io::AsyncWriteExt`, but that needs to run in the
+                // context of a tokio 1 executor, which these futures do not.
+                //let mut file = tokio::fs::File::from_std(file);
+                while let Some(chunk) = stream.next().await {
+                    file.write_all(&chunk?)?;
+                }
```
jan-auer, line 49: This is basically the intent of `read_multipart_file`. It would be nice to move this logic there so it's used by all places dealing with files.

The contents of the comment here are true, although I'm not sure if it's worth writing such a long rant for such a simple loop.

jan-auer (approved):

crates/symbolicator/src/services/symbolication.rs
Swatinem: I think "minidump not found" is the appropriate error here. Also, not sure what to wrap this with, as also everything needs to go through serde, which `std::io::Error` does not.

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1942,9 +1904,38 @@ impl SymbolicationActor {
...
+                let result = match self
+                    .stackwalk_minidump_with_cfi(&minidump_file, &cfi_caches)
+                    .await
+                {
+                    Ok(res) => res,
+                    Err(err) => {
+                        if let Some(dir) = self.diagnostics_cache.cache_dir() {
+                            if let Some(file_name) = minidump_file.file_name() {
+                                let path = dir.join(file_name);
+                                match minidump_file.persist(&path) {
```
flub, line 1916: > Note: Temporary files cannot be persisted across filesystems.

Are we sure this is never going to be the case?  It often is the case that `/tmp` is a different filesystem.  And afaik our cache directory is a persistent volume so very likely to be a different filesystem to wherever tempfile defaults.

/cc @beezz: can you confirm whether this would work or not?

flub (changes requested):
There is some elegance to writing it to disk straight away... but I'm not sure I really like this more than the solution before which didn't have this filesystem problem because it could directly save in the diagnostics cache.  Though thinking of it, this made it impossible to detect what files in the diagnostics cache were actually failed and which ones were life, so maybe that was also not a good solution.

Sorry, only creating problems here it seems, not providing solutions 😕

crates/symbolicator/src/services/symbolication.rs
Swatinem: I changed this to save the files in `$CACHE/tmp` which is also used for other tmp files already, so that won’t run into this particular problem.

flub (approved):

PR #509, opened 2021-07-30
feat(cache): Add a new cache status `CacheSpecificError`

relaxolotl:
This is the start of a series of PRs whose end goal is to allow symbolicator to separate download errors from IO and conversion errors. The end goal is to be able to apply different expiration strategies and include additional diagnostic information to these entries in order to make it easier to debug why a source file might be missing. 

The PRs are being broken down as follows:
1. 👉 Introduce the new cache status without changing the existing functionality, convert the new status to Malformed if needed (#509) 👈 you are here
2. Extend both the new cache status and the Malformed cache status to accept an arbitrary string that describes the root cause of the issue (#510)
3. Identify and split out code paths that should be using the new cache status without changing their return value (#511)
4. Begin actually using the new cache status by returning it in places identified in the previous step and writing + reading those to and from the cache (#512)

Each PR incrementally builds upon the changes of the previous one, but is split up in a way to make it easy to revert an entire PR's changes without breaking the cache.

Currently download (most non-400s) and conversion are both stored as Malformed cache entries, and this is a proposal to create a dedicated cache status to represent these download errors. There are a certain set of errors where the file is known to be present, but have failed to download due to transient/ephemeral errors during the download process. This PR makes no functional changes to symbolicator.

### DIFF
--- a/crates/symbolicator/src/cache.rs
+++ b/crates/symbolicator/src/cache.rs
@@ -29,14 +29,17 @@ pub const MALFORMED_MARKER: &[u8] = b"malformed";
 /// **Cache items of this type are currently not being created. This exists to make it easier to
 /// roll back changes.**
 ///
-/// Content of cache items where downloading the file failed.
+/// Content of cache items where an error has occurred, typically related to a cache-specific
+/// operation. For example, this would be download errors for download caches, and conversion errors
+/// for derived caches.
 ///
 /// Items with this value will be expired after an hour.
 ///
-/// The download error state is used as a way to distinguish between the absence of a file and an
-/// the failure to fetch a file that is known to be present, but unfetchable due to circumstantial
-/// errors. The former scenario is covered by the negative cache state.
-pub const DOWNLOAD_ERROR_MARKER: &[u8] = b"dlerr";
+/// Additional notes for download caches:
+/// This state is used as a way to distinguish between the absence of a file and the failure to
+/// fetch a file that is known to be present, but unfetchable due to transient errors. Absent files
+/// are covered by the negative cache state.
+pub const CACHE_SPECIFIC_ERROR_MARKER: &[u8] = b"cachespecificerror";
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq)]
 pub enum CacheStatus {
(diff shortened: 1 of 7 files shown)


crates/symbolicator/src/cache.rs
```diff
@@ -26,6 +26,18 @@ use crate::types::Scope;
...
+/// **Cache items of this type are currently not being created. This exists to make it easier to
+/// roll back changes.**
+///
+/// Content of cache items where downloading the file failed.
+///
+/// Items with this value will be expired after an hour.
+///
+/// The download error state is used as a way to distinguish between the absence of a file and an
+/// the failure to fetch a file that is known to be present, but unfetchable due to circumstantial
+/// errors. The former scenario is covered by the negative cache state.
```
Swatinem, line 38:
```suggestion
/// The download error state is used as a way to distinguish between the absence of a file and
/// a failure to fetch a file that is known to be present, but unfetchable due to circumstantial
/// errors. The former scenario is covered by the negative cache state.
```

Swatinem (approved):

loewenheim (approved):

flub:
So.. I've been struggling with the question of whether this really should be a DownloadError on the cache level, and I still do not think this is right.  I'm still more of the opinion that it should be a generic "Error" cache status and we currently do weird things.

relaxolotl:
Spoke with @flub about this on a call and I have a better idea of what he'd like to see. His suggestion makes sense: the cache enum being extended here is used by all of the caches, but the meaning of `DownloadError` actually varies based on the cache you're working with. 

For example, derived caches don't ever directly download anything - they may rely on the download cache to do so, but they're not personally responsible for nor do they perform any downloading. Derived caches however may have errors related the conversion of the contents of an entry in the download cache for their own use, and that would be their equivalent of a `DownloadError`. I will be calling this `CacheSpecificError` instead of `DownloadError`.

Without going into the details, the timeouts for this new type of error do sort themselves out nicely and existing propagation/conversion rules in caches that depend on other caches should work fine as well. I'll expand on those in their respective PRs.

relaxolotl:
Actually, this PR covers timeouts. I'll open up a follow-up PR that updates those to line up with the new definition of the cache status being introduced here.

relaxolotl:
Actually, this PR covers timeout strategies. I'll open up a follow-up PR that updates those to match the new definition of the cache status being introduced here: `CacheSpecificErrors` associated with the download caches should retry after an hour (negative strat), but `CacheSpecificErrors` for derived caches should use the same strategy as malformed entries. The current timeout in the diff should be fine as is, especially since this new cache status isn't being used anywhere.

PR #510, opened 2021-07-30
feat(cache): Stuff error messages into `Malformed` and `CacheSpecificError` cache files

relaxolotl:
Welcome to part 2 of my PR series on making symbolicator's cache statuses gooder. **Before reviewing or merging this PR, you may be interested in taking a look at part 1 first for some context: https://github.com/getsentry/symbolicator/pull/509**

If you've already seen #509 or you're brave enough to be look at these changes as-is because you reviewed an eerily similar set of changes in a previous PR, then let's continue.

This belongs to the following chain of PRs:

1. Introduce the new cache status without changing the existing functionality, convert the new status to Malformed if needed (#509)
1. Update cache expiration strategy selection so different caches will expire entries triggered by cache-specific problems after a timeout that makes sense for the cache (#516) 
1. 👉 Extend both the new cache status and the Malformed cache status to accept an arbitrary string that describes the root cause of the issue (#510) 👈 you are here
1. Identify and split out code paths that should be using the new cache status without changing their return value
1. Begin actually using the new cache status by returning it in places identified in the previous step and writing + reading those to and from the cache

This PR does one thing: it stuffs an error string returned by the `Display` impl of the error into `Malformed` and `CacheSpecificError` entries, and adds all of the plumbing needed to write and read them from their respective cache files. This makes no functional changes to symbolicator.

### DIFF
--- a/crates/symbolicator/src/services/symbolication.rs
+++ b/crates/symbolicator/src/services/symbolication.rs
@@ -175,9 +175,13 @@ impl CfiCacheModules {
                     let cfi_status = match cfi_cache.status() {
                         CacheStatus::Positive => ObjectFileStatus::Found,
                         CacheStatus::Negative => ObjectFileStatus::Missing,
-                        CacheStatus::Malformed => {
+                        CacheStatus::Malformed(details) => {
                             let err = CfiCacheError::ObjectParsing(ObjectError::Malformed);
-                            log::warn!("Error while parsing cficache: {}", LogError(&err));
+                            log::warn!(
+                                "Error while parsing cficache: {} ({})",
+                                LogError(&err),
+                                details
+                            );
                             ObjectFileStatus::from(&err)
                         }
                         CacheStatus::CacheSpecificError => ObjectFileStatus::Malformed,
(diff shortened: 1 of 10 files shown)


crates/symbolicator/src/services/symbolication.rs
```diff
@@ -175,12 +175,17 @@ impl CfiCacheModules {
                     let cfi_status = match cfi_cache.status() {
                         CacheStatus::Positive => ObjectFileStatus::Found,
                         CacheStatus::Negative => ObjectFileStatus::Missing,
-                        CacheStatus::Malformed => {
+                        CacheStatus::Malformed(details) => {
                             let err = CfiCacheError::ObjectParsing(ObjectError::Malformed);
-                            log::warn!("Error while parsing cficache: {}", LogError(&err));
+                            log::warn!(
+                                "Error while parsing cficache: {} ({})",
+                                LogError(&err),
```
Swatinem, line 182: doesn’t `LogError` already contain the `details`?

Swatinem (approved):

loewenheim (approved):

PR #515, opened 2021-08-04
WIP: Use tracing::instrument on some processing functions

Swatinem:
This can generate traces like this one: https://sentry.io/organizations/sentry-test/performance/rust:daa7a6a37bde40c48165da73693a21a5/?project=1041156

Some problems with this though:
* the `tracing_subscriber::fmt` clashes with env_logger I think. I think rewriting our logging to be completely based on tracing would be a nice idea. But gotta do it in such a way that we don’t break the loggers/configs that do exist.
* we gotta explicitly list all the args as `skip` that we don’t want. In our case, we really want none of them, so its kind of a pain in the rear to instrument. There are a couple of issues/feature requests in the tracing repo for this and even a WIP PR to skip all the args, but it takes some time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -162,6 +162,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi_term"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
+dependencies = [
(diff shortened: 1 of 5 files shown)


flub:
I feel like we should have samping figured out before merging something like this.  we might just start to hurt s4s really fast and need to be able to undo it quickly

Swatinem:
Well obviously I set that to 100% for testing. Otherwise we can sample way down. Also whats not done yet is to correlate this with a parent trace/span that comes from the workers. That would be interesting as well.

PR #518, opened 2021-08-07
feat(cache): Mention when a download failure is due to insufficient permissions

relaxolotl:
If we know that the reason why we weren't able to download a file was due to permissions we now record that fact and return it in the symbolication response.

This bases itself off of #512 to avoid conflicts with the download error work that's currently open in PRs.

### DIFF
--- a/crates/symbolicator/src/services/download/mod.rs
+++ b/crates/symbolicator/src/services/download/mod.rs
@@ -64,7 +64,7 @@ pub enum DownloadError {
 
 impl DownloadError {
     /// This produces a user-facing string representation of a download error if it is a variant
-    /// that needs to be stored as a acheStatusCacheSpecificError entry in the download cache.
+    /// that needs to be stored as a `CacheStatus::CacheSpecificError` entry in the download cache.
     pub fn for_cache(&self) -> String {
         match self {
             DownloadError::Gcs(inner) => format!("{}: {}", self, inner),
@@ -76,8 +76,9 @@ impl DownloadError {
         }
     }
 
-    /// This produces a user-facing string representation of a download error if it is a variant
-    /// that needs to be stored as a CacheSpecificError entry in the download cache.
+    /// If a given cache entry is `CacheStatus::CacheSpecificError`, this parses and extracts its
+    /// contents into a `DownloadError`. This will return none if a
+    /// non-`CacheStatus::CacheSpecificError` is provided.
     pub fn from_cache(status: &CacheStatus) -> Option<Self> {
         match status {
             CacheStatus::Positive => None,
--- a/crates/symbolicator/src/services/download/mod.rs
+++ b/crates/symbolicator/src/services/download/mod.rs
@@ -15,6 +15,7 @@ use thiserror::Error;
 use tokio::fs::File;
 use tokio::io::AsyncWriteExt;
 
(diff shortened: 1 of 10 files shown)


Swatinem (approved):
lgtm, but please run some tests through sentry to make sure it can deal with these noperm errors. we ran into a couple of situations where we broke sentry because we changed the results that we return, and the validation for these is quite strict.

crates/symbolicator/src/services/download/mod.rs
```diff
@@ -65,10 +68,26 @@ impl DownloadError {
             DownloadError::Gcs(inner) => format!("{}: {}", self, inner),
             DownloadError::Sentry(inner) => format!("{}: {}", self, inner),
             DownloadError::S3(inner) => format!("{}: {}", self, inner),
+            DownloadError::Permissions => String::from("insufficient permissions"),
```
relaxolotl, line 71: we can just use the string representation of `Download::Permissions` here, any micro-optimization achieved by using this marginally shorter string isn't really worth it, and we no longer need to track yet another magic string by removing `"insufficient permissions"`

loewenheim (approved):

crates/symbolicator/src/services/download/mod.rs
```diff
@@ -63,24 +63,28 @@ pub enum DownloadError {
 }
 
 impl DownloadError {
+    /// This produces a user-facing string representation of a download error if it is a variant
+    /// that needs to be stored as a acheStatusCacheSpecificError entry in the download cache.
```
Swatinem, line 67: a what now? :-D

relaxolotl: oh geez i just left that half-completed thank you

relaxolotl:
@Swatinem:

> please run some tests through sentry to make sure it can deal with these noperm errors.

I quickly ran through some example errors in my local build of sentry to make sure nothing broke, but I believe armenzg very kindly added in integration tests to our pipeline to test this out, i.e. if this https://github.com/getsentry/symbolicator/runs/3349120741 job is passing we are good.

crates/symbolicator/src/services/download/mod.rs
```diff
@@ -60,15 +63,36 @@ pub enum DownloadError {
 }
 
 impl DownloadError {
+    /// This produces a user-facing string representation of a download error if it is a variant
+    /// that needs to be stored as a `CacheStatus::CacheSpecificError` entry in the download cache.
```
Swatinem, line 67:
```suggestion
    /// that needs to be stored as a [`CacheStatus::CacheSpecificError`] entry in the download cache.
```

This turns these into links in rustdoc btw.

Swatinem (commented):
I don’t think we have an integration test specifically for this. Might be a good idea to add one though. Just configure an obviously bogus s3 source, and see that surfaces the permission error. Though not sure if S3 gives you different errors if you provide a completely bogus bucket, or just if the keys are invalid.

relaxolotl:
also checked this locally to make sure sentry didn't reject the new-ish payload, and i can confirm that it does not (and it even mentions "permission")

Screenshot 2021-08-19 at 8 51 10 AM

Swatinem:
nice! so the UI was able to deal with these statuses after all ;-)

PR #525, opened 2021-08-23
ref: Move deduplication logic to CacheItemRequest::compute

Swatinem:
Previously the whole cache lookup was being deduplicated by a channel indirection.
This splits that apart and makes only the actual computation deduplicated.
This opens up the road to be able to both eagerly compute, and lazily recompute cached items.

This does not completely avoid the `Arc`s, and it had to do quite a bit of extra work to avoid using `self`, as using that would get us into lifetime troubles.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/cacher.rs
+++ b/crates/symbolicator/src/services/cacher.rs
@@ -15,9 +15,10 @@ use crate::cache::{Cache, CacheStatus};
 use crate::types::Scope;
 use crate::utils::futures::{spawn_compat, BoxedFuture, CallOnDrop};
 
+type ComputationResult<T, E> = Result<Arc<T>, Arc<E>>;
 // Inner result necessary because `futures::Shared` won't give us `Arc`s but its own custom
 // newtype around it.
-type ComputationChannel<T, E> = Shared<oneshot::Receiver<Result<Arc<T>, Arc<E>>>>;
+type ComputationChannel<T, E> = Shared<oneshot::Receiver<ComputationResult<T, E>>>;
 
 type ComputationMap<T, E> = Arc<Mutex<BTreeMap<CacheKey, ComputationChannel<T, E>>>>;
 
@@ -126,7 +127,7 @@ fn safe_path_segment(s: &str) -> String {
         .replace(":", "_") // not a threat on POSIX filesystems, but confuses OS X Finder
 }
 
-pub trait CacheItemRequest: 'static + Send {
+pub trait CacheItemRequest: 'static + Send + Sync + Clone {
     type Item: 'static + Send + Sync;
 
     // XXX: Probably should have our own concrete error type for cacheactor instead of forcing our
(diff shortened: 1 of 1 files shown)


loewenheim (approved):

crates/symbolicator/src/services/cacher.rs
```diff
@@ -187,64 +188,58 @@ impl<T: CacheItemRequest> Cacher<T> {
...
-        let key = request.get_cache_key();
-
-        // A file was not found. If this spikes, it's possible that the filesystem cache
-        // just got pruned.
-        metric!(counter(&format!("caches.{}.file.miss", name)) += 1);
-
-        let temp_file = self.tempfile()?;
+    async fn compute(
+        name: &'static str,
+        tempfile: std::io::Result<NamedTempFile>,
```
flub, line 237: Your description says you run into lifetime issues when using `&self` and thats why these are here.  But the current arguments are a bit confusing and I had to look at both versions to understand them.  I wonder if passing `self` is another reasonable stopgap here?  It does mean you have to clone all of self, but `Cacher::current_computations` is already an `Arc` and `Cache` is... still clonable but not super cheap.

Swatinem: I think the problem basically was that somewhere, we require these futures to be `Send + 'static`, which meant that also `&self` needed to be `'static`, which essentially means you need to take it by value. I remember we had that problem in other places as well, but here I opted to just explicitly call the async fn (create the future) with just the types that are needed inside. The `tempfile` is a bit unfortunate though :-(

crates/symbolicator/src/services/cacher.rs
```diff
@@ -356,4 +352,38 @@ impl<T: CacheItemRequest> Cacher<T> {
...
+            if let Some(item) = self.lookup_cache(&request, &key, path)? {
+                return Ok(Arc::new(item));
+            }
+        }
+
+        // A file was not found. If this spikes, it's possible that the filesystem cache
+        // just got pruned.
+        metric!(counter(&format!("caches.{}.file.miss", name)) += 1);
+
+        Ok(self.spawn_computation(request.clone(), cache_path).await?)
```
flub, line 387: iiuc you can run into a race condition here where you checked the cache key, didn't find anything but a computation is already running, then spawn the computation but the previous computation already completed and now you're computing it twice.  I think this is something the old version avoided?
(sorry, didn't realise this yesterday when we were talking about this)

Swatinem: I think that is in theory possible, but wouldn’t worry about it too much, since these things are idempotent.
But we can do an additional cache fetch attempt inside of `compute` just in case, wdyt?

crates/symbolicator/src/services/cacher.rs
Swatinem: Oh, so the old code already cloned. Oh well, I changed it now so that it continues cloning. It looks like the cleanest solution right now.

crates/symbolicator/src/services/cacher.rs
flub: yes, that's probably a reasonable solution

flub (approved):

PR #527, opened 2021-08-24
test: Unitest for creating ELF DIF lookup paths

flub:
This adds a unittest which checks the paths for ELF DIFs are generated
correctly.  Trivial, but it helped me understand a user's issue.

#skip-changelog


Swatinem (approved):
are the unittest failures intermittent timing issues?

flub:
> are the unittest failures intermittent timing issues?

hum, I think so.  it's a test you wrote sometime and does not have anything to do with this.  It has a `TIMEOUT` of 500 milliseconds so probably flaky :(   I think we should address that in a separate PR though

PR #528, opened 2021-08-24
test: Increase timeout for less flaky test

flub:
This is a cheap attempt to make the test less flaky by simply giving
it a much larger timeout.

See #512 where this was flaky.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/symcaches.rs
+++ b/crates/symbolicator/src/services/symcaches.rs
@@ -450,7 +450,7 @@ mod tests {
     async fn test_symcache_refresh() {
         test::setup();
 
-        const TIMEOUT: Duration = Duration::from_millis(500);
+        const TIMEOUT: Duration = Duration::from_secs(5);
 
         let cache_dir = test::tempdir();
         let symbol_dir = test::tempdir();


crates/symbolicator/src/services/symcaches.rs
```diff
@@ -450,7 +450,7 @@ mod tests {
     async fn test_symcache_refresh() {
         test::setup();
 
-        const TIMEOUT: Duration = Duration::from_millis(500);
+        const TIMEOUT: Duration = Duration::from_secs(5);
```
Swatinem, line 453:
```suggestion
        const TIMEOUT: Duration = Duration::from_secs(2);
```

The flackyness might be related to filesystem timestamps. I don’t think we have to be overly generous with the timeout.

Swatinem (approved):

crates/symbolicator/src/services/symcaches.rs
flub: Hum, but what's wrong with being overly generous?  I'd prefer to give it plenty of time.  CI machines are busy sometimes.

PR #529, opened 2021-08-24
ref: Remove last use of tryf! macro

Swatinem:
This also removes the `clone` macro which has also been unused.

#skip-changelog


flub (approved):
yay!

PR #531, opened 2021-08-24
ref: Get rid of spawn_compat which assumes a tokio 0.1 Runtime

Swatinem:


loewenheim (approved):

crates/symbolicator/src/utils/futures.rs
```diff
@@ -17,7 +13,7 @@ use crate::metrics::{self, prelude::*};
 /// You would mainly use this if you are dealing with a trait methods which deals with
 /// futures.  Trait methods can not be async/await and using this type in their return value
 /// allows to integrate with async await code.
-pub type BoxedFuture<T> = Pin<Box<dyn Future<Output = T>>>;
+pub type BoxedFuture<T> = Pin<Box<dyn Future<Output = T> + Send + 'static>>;
```
flub, line 16: I wonder if you can now use `futures::future::BoxFuture` for this instead of our own type alias.  (it's also in futures' prelude)

crates/symbolicator/src/services/mod.rs
```diff
@@ -91,6 +91,7 @@ impl Service {
             symcaches,
             cficaches,
             caches.diagnostics,
+            io_pool,
```
flub, line 94: afaik this pool is only constructed with 4 threads at the moment because it is only used by the download service until now.  We should now construct it with the default number of threads I think, which is the number of available CPUs afaik.

flub (commented):
This is a very innocent title for quite an important architectural change 😓 
I think it should be titled "Run symbolication actor on tokio 1.0 IO threadpool" and should have a changelog item describing this.

And please be careful when deploying this.  It is probably best not to deploy this at the same time as the caching changes being introduced.

Looks like you've been having some fun though!

PR #532, opened 2021-08-24
ref: Replace last future_metrics usage with measure

Swatinem:
This is also removes that macro completely, and is a prerequisite to making cache computation futures `Send`, so we can get rid of the `spawn_compat` nonsense in a followup.

#skip-changelog


loewenheim (approved):

flub (approved):
I also hate measure, and even more after you added an extra argument.  but i hate future_metrics even more 😄

PR #540, opened 2021-08-27
fix: Clamp lazy-computation options to avoid misconfiguration

Swatinem:
#skip-changelog

### DIFF
--- a/crates/symbolicator/src/cache.rs
+++ b/crates/symbolicator/src/cache.rs
@@ -400,11 +400,14 @@ pub struct Caches {
 
 impl Caches {
     pub fn from_config(config: &Config) -> io::Result<Self> {
+        // The minimum value here is clamped to 1, as it would otherwise completely disable lazy
+        // re-generation. We might as well decide to hard `panic!` on startup if users have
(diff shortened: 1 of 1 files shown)


relaxolotl (approved):
Thanks for adding this in! 

I think if there's already some precedence for symbolicator to panic if an invalid config value is passed in, then I would strongly agree that we should just panic here. If not, then I'm open to leaving it as-is for the time being and maybe we could start an informal discussion about whether symbolicator should start doing this for invalid config values.

Regarding a "no limit" config value: If I understand correctly, it seems like we have empty-ish values we'd have to consider here, right?

1. Don't lazily recompute at all
2. There is no limit to how many lazy redownloads can be pushed to the queue

Did we want to include both? Or only one? I can see how both may be useful for testing purposes, but I'm unsure how those should be represented in a config.

Swatinem:
I don’t think we actually want to allow no lazy recompute at all. It is just that we want to throttle it reasonably. Also for the "no limit" case, symbolicator would fall over far ahead of reaching a limit of, say, 999. So I guess this should be good.

PR #544, opened 2021-08-31
ref: Remove actix-web in favor of axum (NATIVE-193)

Swatinem:
Still needs a bit of work, there are a few TODOs scattered around, but it already gets rid of old actix/futures/tokio.

There are a few changes from the current actix implementation:
- [x] Errors are serialized via `ApiResponseError` or however it is called, this might be good to do as a followup refactor.
- [x] The vendored senty-actix integration used to capture server errors
- [ ] The old code imposed some limits on JSON and multipart bodies, the new code does not. I think we can just drop this, as sentry pushes requests to symbolicator, so they should be well formed.
- [x] The vendored sentry actix integration attached some event attributes, like request url, headers, etc, not sure how useful it is?


crates/symbolicator/src/endpoints/mod.rs
```diff
@@ -1,28 +1,91 @@
...
+use std::time::Instant;
 
-use crate::services::symbolication::MaxRequestsError;
-use crate::services::Service;
+use axum::body::Body;
+use axum::handler::{get, post};
+use axum::http::{Response, StatusCode};
+use axum::response::IntoResponse;
+use axum::Router;
+use sentry::integrations::tower::NewSentryLayer;
```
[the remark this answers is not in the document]
Swatinem: This was done recently.
https://github.com/getsentry/sentry-rust/pull/356
It is the reason why this pulls in sentry-rust via git. I think we should be able to cut a release for that any time.

crates/symbolicator/src/test.rs
```diff
@@ -197,6 +160,20 @@ impl Server {
...
     }
 
+    pub fn with_service(service: Service) -> Self {
+        let socket = SocketAddr::from(([127, 0, 0, 1], 0));
+
+        let server =
+            axum::Server::bind(&socket).serve(endpoints::create_app(service).into_make_service());
+
+        let socket = server.local_addr();
+        let handle = tokio::spawn(async {
```
flub, line 170: Does this need a `CancelOnDrop` wrapper?

(btw, I'd prefer `AbortOnDrop` for that name I think.  matches the tokio naming of aborting a future and is less confusing since we also have `CallOnDrop`)

crates/symbolicator/src/endpoints/mod.rs
```diff
@@ -1,28 +1,43 @@
...
-        .configure(proxy::configure)
-        .configure(requests::configure)
-        .configure(symbolicate::configure)
+pub use error::ResponseError;
+
+use applecrashreport::handle_apple_crash_report_request as applecrashreport;
+use minidump::handle_minidump_request as minidump;
+use proxy::proxy_symstore_request as proxy;
+use requests::poll_request as requests;
+use symbolicate::symbolicate_frames as symbolicate;
```
flub, line 21: nitpicking: Isn't this section above the mod section normally?  Also the `pub use` one should be the bottom of this section afaik https://www.notion.so/sentry/HOWTO-Code-Rust-at-Sentry-7b35f165b10b4492bb95ebe1471a9ada#57304f11240944338fe3828059bcf463

crates/symbolicator/src/utils/sentry.rs
```diff
@@ -1,320 +1,10 @@
...
-
-    fn sentry_hub_new_from_current(self) -> SentryFuture<Self> {
-        self.bind_hub(Hub::new_from_top(Hub::current()))
-    }
-}
-
-impl<F> SentryFutureExt for F where F: futures01::future::Future {}
-
 /// Write own data to [`sentry::Scope`], only the subset that is considered useful for debugging.
 pub trait ConfigureScope {
```
flub, line 2: lol, I've never liked this trait and wouldn't mind if this goes away too.  but doesn't matter for this PR

flub:
I'm kind of in favour of still resolving the two unticked issues in the description.

crates/symbolicator/src/utils/sentry.rs
```diff
@@ -8,3 +15,65 @@ pub trait ConfigureScope {
...
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = S::Future;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        self.service.poll_ready(cx)
+    }
+
+    fn call(&mut self, request: Request<Body>) -> Self::Future {
+        sentry::configure_scope(|scope| {
```
flub, line 48: Shouldn't we turn the future returned from `self.service.call()` into a `SentryFuture` and use `with_scope` instead so that this is all popped back off the scope when this is done?

crates/symbolicator/src/utils/sentry.rs
```diff
@@ -1,320 +1,79 @@
...
-        ty: parse_type_from_debug(&dbg).to_owned(),
-        value: Some(f.to_string()),
-        stacktrace: bt
-            // format the stack trace with alternate debug to get addresses
-            .map(|bt| format!("{:#?}", bt))
-            .and_then(|x| parse_stacktrace(&x)),
+fn extract_request(req: &Request<Body>) -> (Option<String>, sentry::protocol::Request) {
+    let transaction = Some(req.uri().path().to_string());
+    let sentry_req = sentry::protocol::Request {
+        url: req.uri().to_string().parse().ok(),
```
flub, line 68: Does this extract e.g. request.scope nicely?

crates/symbolicator/src/utils/sentry.rs
Swatinem: The `sentry-tower` layer makes sure that every request gets its own fresh hub and scope, and if I understand the tower concepts correctly, all child layers/routes get the correct hub.

crates/symbolicator/src/test.rs
Swatinem: `test::Server` itself does exactly that on `Drop`.

crates/symbolicator/src/utils/sentry.rs
Swatinem: Turned out `request.scope` came from somewhere else entirely :-D

crates/symbolicator/src/utils/sentry.rs
```diff
@@ -46,34 +46,26 @@ where
 
     fn call(&mut self, request: Request<Body>) -> Self::Future {
         sentry::configure_scope(|scope| {
-            let (transaction, req) = extract_request(&request);
+            let headers = request.headers();
+            fn get_str(val: &HeaderValue) -> Option<&str> {
+                val.to_str().ok()
+            }
```
flub: heh, i would never have thought of writing this as an fn rather than a closure.  but that makes total sense.

loewenheim (approved):

flub (approved):
Please make a followup task for the 5Mb limit of the symbolicate request body before merging.

Swatinem:
I filed that as https://getsentry.atlassian.net/browse/NATIVE-207 as a separate backlog task so it does not block us from marking the parent task as "done".

PR #545, opened 2021-08-31
build(ci): Fix sentry integration test due to refactored actions

billyvg:
We refactored [`setup-sentry`](https://github.com/getsentry/sentry/pull/28281) to be able to compose actions better. This workflow was only copying `setup-sentry` which now include `setup-python`.

#skip-changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,25 +146,12 @@ jobs:
           repository: getsentry/sentry
           path: sentry
 
-      - uses: actions/setup-python@v2
-        with:
-          python-version: 3.6
-
       - name: Setup steps
         id: setup
         run: |
-          pip install --upgrade pip wheel
-          echo "::set-output name=pip-cache-dir::$(pip cache dir)"
           # We cannot execute actions that are not placed under .github of the main repo
-          mkdir -p .github/actions/setup-sentry/
-          cp sentry/.github/actions/setup-sentry/action.yml .github/actions/setup-sentry/action.yml
-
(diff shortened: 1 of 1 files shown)


billyvg:
Requires https://github.com/getsentry/sentry/pull/28329 to work

.github/workflows/ci.yml
```diff
@@ -146,32 +146,21 @@ jobs:
...
-          cp sentry/.github/actions/setup-sentry/action.yml .github/actions/setup-sentry/action.yml
-
-      - name: Sentry's pip cache
-        uses: actions/cache@v2
-        with:
-          path: ${{ steps.setup.outputs.pip-cache-dir }}
-          key: sentry-deps-${{ hashFiles('sentry/requirements**.txt') }}
-          restore-keys: sentry-deps-
+          mkdir -p .github/actions/
+          cp -r sentry/.github/actions/* .github/actions/
```
relaxolotl, line 154: nit: are we using other actions that need to be copied over besides `setup-sentry`? 

could we just continue to directly copy over `setup-sentry/action.yml` (ie deleted L160 in this diff) if that's all that's used here? 

if we need more files in `setup-sentry/` besides `action.yml`, could we recursively copy `actions/setup-sentry/` instead of everything in `actions/`?

relaxolotl (approved):

.github/workflows/ci.yml
billyvg: It has another dependency, I've updated to explicitly copy them over.

Swatinem (approved):

PR #546, opened 2021-09-01
ref: Avoid spawning separate download tasks (NATIVE-191)

Swatinem:
Now that the main symbolicator runs on the IO pool, it is not necessary to spawn download tasks on that same pool and await those tasks. Just await those futures directly.
This also simplifies things around taking ownership of self and arguments.

This also introduces a `CancelOnDrop` wrapper which is wrapped around spawns that are otherwise never canceled when the outer future is being timed out.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/bitcode.rs
+++ b/crates/symbolicator/src/services/bitcode.rs
@@ -303,13 +303,13 @@ impl BitcodeService {
         hub.configure_scope(|scope| scope.set_extra("auxdif.source", source.type_name().into()));
 
         let file_type = match dif_kind {
-            AuxDifKind::BcSymbolMap => vec![FileType::BcSymbolMap],
-            AuxDifKind::UuidMap => vec![FileType::UuidMap],
+            AuxDifKind::BcSymbolMap => &[FileType::BcSymbolMap],
+            AuxDifKind::UuidMap => &[FileType::UuidMap],
         };
         let file_sources = self
             .download_svc
-            .clone()
-            .list_files(source, file_type, uuid.into(), hub.clone())
+            .list_files(source, file_type, uuid.into())
+            .bind_hub(hub.clone())
             .await?;
 
         let mut fetch_jobs = Vec::with_capacity(file_sources.len());
--- a/crates/symbolicator/src/services/download/mod.rs
+++ b/crates/symbolicator/src/services/download/mod.rs
@@ -149,12 +149,12 @@ impl DownloadService {
     /// Dispatches downloading of the given file to the appropriate source.
     async fn dispatch_download(
         &self,
-        source: RemoteDif,
+        source: &RemoteDif,
         destination: PathBuf,
     ) -> Result<DownloadStatus, DownloadError> {
         let result = future_utils::retry(|| async {
             let destination = destination.clone();
-            match &source {
+            match source {
                 RemoteDif::Sentry(inner) => {
                     self.sentry
                         .download_source(inner.clone(), destination)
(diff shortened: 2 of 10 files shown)


crates/symbolicator/src/services/download/mod.rs
```diff
@@ -238,42 +224,28 @@ impl DownloadService {
...
-                    slf.sentry
-                        .list_files(cfg, object_id, &filetypes, config)
-                        .bind_hub(hub)
-                        .await
-                };
 
-                // NB: Enter the tokio 1 runtime, which is required to create the timeout.
-                // See: https://docs.rs/tokio/1.0.1/tokio/runtime/struct.Runtime.html#method.enter
-                let _guard = self.worker.enter();
+                let job = self.sentry.list_files(cfg, object_id, filetypes, config);
```
jan-auer, line 236: Can you even get rid of the config clone and pass in a reference instead?

crates/symbolicator/src/services/download/mod.rs
```diff
@@ -238,42 +224,28 @@ impl DownloadService {
...
                 let job = tokio::time::timeout(Duration::from_secs(30), job);
                 let job = measure("service.download.list_files", m::timed_result, None, job);
 
-                // Map all SpawnError variants into DownloadError::Canceled.
-                match self.worker.spawn(job).await {
-                    Ok(Ok(result)) => result,
-                    Ok(Err(_)) | Err(_) => Err(DownloadError::Canceled),
+                match job.await {
+                    Ok(result) => result,
+                    Err(_) => Err(DownloadError::Canceled),
```
jan-auer, line 242: _nit_: Use `map_err`

crates/symbolicator/src/services/download/mod.rs
```diff
@@ -542,14 +514,9 @@ mod tests {
...
-        let svc = DownloadService::new(config, tokio::runtime::Handle::current());
+        let svc = DownloadService::new(config);
         let file_list = svc
-            .list_files(
-                source.clone(),
-                FileType::all().to_vec(),
-                objid,
-                Hub::current(),
-            )
+            .list_files(source.clone(), FileType::all(), objid)
```
jan-auer, line 519: Can you pass a reference to `source` here, instead?

crates/symbolicator/src/utils/futures.rs
```diff
@@ -29,6 +31,44 @@ impl Drop for CallOnDrop {
...
+/// Cancels the [`JoinHandle`] on drop.
+///
+/// Spawning a task on a runtime means it will run independently from the code that calls `spawn`,
+/// even if the code stops polling the [`JoinHandle`]. We have various timeouts configured throughout
+/// the codebase, and some of those are attached to [`JoinHandle`]s.
+/// Which means we stop polling the handle, but the task that was spawned will continue on.
+///
+/// This type makes sure that the spawned task is being canceled/aborted in case we lose interest
+/// in it.
+pub struct CancelOnDrop<T> {
```
jan-auer, line 43: Would annotate this with `must_use`.

flub (approved):

crates/symbolicator/src/services/bitcode.rs
```diff
@@ -303,13 +303,13 @@ impl BitcodeService {
...
-            AuxDifKind::UuidMap => vec![FileType::UuidMap],
+            AuxDifKind::BcSymbolMap => &[FileType::BcSymbolMap],
+            AuxDifKind::UuidMap => &[FileType::UuidMap],
         };
         let file_sources = self
             .download_svc
-            .clone()
-            .list_files(source, file_type, uuid.into(), hub.clone())
+            .list_files(source, file_type, uuid.into())
+            .bind_hub(hub.clone())
```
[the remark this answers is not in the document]
Swatinem: Its a bit weird, I think we were creating a new hub_from_current here in order to correctly pop the configure_scope.

PR #548, opened 2021-09-02
Multipart streaming files and size limits

flub:


crates/symbolicator/src/endpoints/applecrashreport.rs
```diff
@@ -11,6 +15,31 @@ use crate::utils::sentry::ConfigureScope;
...
+
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        let this = &mut self;
+        let inner = std::pin::Pin::new(&mut this.inner);
+        inner
+            .poll_next(cx)
+            .map_err(|_err| std::io::Error::new(std::io::ErrorKind::Other, "oops"))
```
Swatinem, line 39: I think this here is the critical part, right? tokio expects everthing to be `io::Error`, and while the `Field` itself implements `Stream`, it has a different error type? Why doesn’t `StreamReader` offer a `map_err` combinator, that would be so convenient.

crates/symbolicator/src/endpoints/applecrashreport.rs
```diff
@@ -11,6 +15,31 @@ use crate::utils::sentry::ConfigureScope;
...
+
+impl<'a> Stream for MultipartField<'a> {
+    type Item = Result<bytes::Bytes, std::io::Error>;
+
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        let this = &mut self;
+        let inner = std::pin::Pin::new(&mut this.inner);
```
Swatinem, line 36: I’m a bit surprised. I was using an unsafe block everywhere to wrap futures, but it can be totally safe as well?

Swatinem (approved):
maybe `use` more of the fully qualified types, but otherwise looks good. I wonder why CI does not like it?

crates/symbolicator/src/endpoints/applecrashreport.rs
```diff
@@ -11,6 +15,31 @@ use crate::utils::sentry::ConfigureScope;
...
+}
+
+impl<'a> MultipartField<'a> {
+    fn new(inner: axum::extract::multipart::Field<'a>) -> Self {
+        Self { inner }
+    }
+}
+
+impl<'a> Stream for MultipartField<'a> {
+    type Item = Result<bytes::Bytes, std::io::Error>;
```
flub, line 29: Let me explain why the Newtype around `Field`.  The problem is that `StreamReader` needs a `Stream<Item = Result<T, std::io::Error>` and the native `Field` doesn't do that.  The error itself should be something better obviously, but otherwise I think we can create a new `read_multipart_file` out of this.

I prefer to convert to `AsyncReader` as it allows using `tokio::io::copy` which handles buffering more efficient than building a dumb `if let Some(chunk) = stream.next().await` loop.

For a new `read_multipart_data` function which has filesize limits I think we'll have to use the `if let` loop as I'm not sure what we can do better.

crates/symbolicator/src/endpoints/applecrashreport.rs
flub: yes, i just added some comments to clarify this.  And yeah, having to newtype this just to change the error type is a pain :(

crates/symbolicator/src/endpoints/applecrashreport.rs
flub: "safe" in as far pin projection is safe.  but yes, no unsafe block needed, though every time i need this i need to go look up examples of what the trick is.  And lol, just noticed I missed a deref, it should be `let this = &mut *self` so that `this` is no longer pinned.  But according to rust analyser it might not matter because the type of inner doesn't change...

flub:
One implication from this is that we start using `tokio::fs` which spawns all the IO operation on the "blocking threads" from the runtime.  Since we don't configure these it is set to 512, so we'll be spawning up to that many new threads.  It remains to be seen what the overhead of this spawning actually is.

flub:
@Swatinem PTAL, I've left the 5Mb symblicate request size unsolved for now.  I suspect you can do this cleverly with tower but i got lost there.  The other, probably simpler alternative I think is to extract a raw `Body` and then check the content type and stream to a vec using `if let ... .next().await` breaking if it is to large, finally convert to JSON using serde.  Again I kind of got lost but this certainly seems doable.

Also, now that `stream_multipart_file` is just 3 lines and only used in 2 places I wonder if it is even worth having at all as a function.

crates/symbolicator/src/endpoints/multipart.rs
```diff
@@ -0,0 +1,47 @@
...
+use tokio_util::io::StreamReader;
+
+use super::ResponseError;
+
+/// Stream a multipart body to a file.
+///
+/// The file's cursor will be set to the beginning of the file again.
+pub async fn stream_multipart_file(field: Field<'_>, file: &mut File) -> Result<(), ResponseError> {
+    let mut field_reader =
+        StreamReader::new(field.map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err)));
```
Swatinem, line 19: aha! So there was a `map_err` after all ;-)

Swatinem (approved):
This looks awesome! I think the CI failures are because of the updated integration test in master, I merged that into the `axum` branch, so you might rebase, or just merge, since its onto the feature branch anyway.

crates/symbolicator/src/endpoints/multipart.rs
flub: `futures::...::TryStreamExt`!  i suddenly recalled that should exist

PR #550, opened 2021-09-03
ref: Move iterative stackwalking into its own method

Swatinem:
#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/symbolication.rs
+++ b/crates/symbolicator/src/services/symbolication.rs
@@ -1977,6 +1977,76 @@ impl SymbolicationActor {
             .context("Minidump stackwalk future cancelled")?)
     }
 
+    /// Iteratively stackwalks/processes the given `minidump_file` using breakpad.
+    async fn stackwalk_minidump_iteratively_with_breakpad(
+        &self,
+        scope: Scope,
+        minidump_file: TempPath,
+        sources: Arc<[SourceConfig]>,
+        cfi_caches: &mut CfiCacheModules,
+    ) -> anyhow::Result<StackWalkMinidumpResult> {
+        let mut iterations = 0;
+
+        let result = loop {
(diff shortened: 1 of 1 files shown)


crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1977,6 +1977,76 @@ impl SymbolicationActor {
...
+                .await
+            {
+                Ok(res) => res,
+                Err(err) => {
+                    if let Some(dir) = self.diagnostics_cache.cache_dir() {
+                        if let Some(file_name) = minidump_file.file_name() {
+                            let path = dir.join(file_name);
+                            match minidump_file.persist(&path) {
+                                Ok(_) => {
+                                    sentry::configure_scope(|scope| {
```
flub, line 2004: should this be a `with_scope`?  otherwise this crashed minidump stays there for a long time on the scope?

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1977,6 +1977,76 @@ impl SymbolicationActor {
...
+
+        let result = loop {
+            iterations += 1;
+
+            let result = match self
+                .stackwalk_minidump_with_cfi(&minidump_file, cfi_caches)
+                .await
+            {
+                Ok(res) => res,
+                Err(err) => {
```
flub, line 1998: nit: for readability i think i'd prefer if this entire block was moved to a dedicated method.  it does a bunch of stuff that's not really core to the logic of this function.

flub (approved):

crates/symbolicator/src/services/symbolication.rs
Swatinem: At first I thought yes, but looking at the code again, its impossible since this function-scope is completely different from whereever the error itself is being triggered :-(

PR #551, opened 2021-09-03
feat: Prefer client-side stack traces over breakpad ones (NATIVE-196)

Swatinem:
Overrides the breakpad-processed stack traces by client-side stack traces extracted from our minidump extension if those are available.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/symbolication.rs
+++ b/crates/symbolicator/src/services/symbolication.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap};
 use std::convert::TryInto;
 use std::fs::File;
 use std::future::Future;
@@ -33,6 +33,7 @@ use thiserror::Error;
 use crate::cache::CacheStatus;
 use crate::logging::LogError;
 use crate::services::cficaches::{CfiCacheActor, CfiCacheError, CfiCacheFile, FetchCfiCache};
+use crate::services::minidump::parse_stacktraces_from_minidump;
 use crate::services::objects::{FindObject, ObjectError, ObjectPurpose, ObjectsActor};
 use crate::services::symcaches::{FetchSymCache, SymCacheActor, SymCacheError, SymCacheFile};
 use crate::sources::{FileType, SourceConfig};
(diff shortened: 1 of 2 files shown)


loewenheim (approved):

flub (commented):
the diff here looks weird and huge?  i'm sure i've seen some of this already in other places?  do we need to merge some other PRs or target the right places before reviewing this?

Swatinem:
Yes, I will rebase/merge this on master first thing tomorrow.

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -2381,6 +2396,36 @@ impl From<&SymCacheError> for ObjectFileStatus {
     }
 }
 
+/// Merges the Stack Traces processed via Breakpad with the ones captured on the Client.
+///
+/// For now, this means we will prefer the client-side stack trace over the processed one, but in
+/// the future we could be a bit smarter about what to do.
+fn merge_clientside_with_processed_stacktraces(
+    processed_stacktraces: &mut [RawStacktrace],
+    clientside_stacktraces: Vec<RawStacktrace>,
```
flub, line 2405: I wonder if a move here is more expensive than a reference?  Or would the compiler figure that out itself and do something smart?

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -2381,6 +2396,36 @@ impl From<&SymCacheError> for ObjectFileStatus {
...
+
+    for thread in processed_stacktraces {
+        if let Some(thread_id) = thread.thread_id {
+            if let Some(client_thread) = client_traces_by_id.remove(&thread_id) {
+                // NOTE: we could gather all kinds of metrics here, as in:
+                // - are we finding more or less frames via CFI?
+                // - how many frames are the same
+                // - etc.
+                // We could also be a lot smarter about which threads/frames we chose. For now we
+                // will just always prefer client-side stack traces
```
flub, line 2420: Can we at least count the number of events in which we choose client-side frames for a thread?  Like have *any* idea that this code did something, currently there's totally nothing, no log, no metric.  We just have to guess at what happened.

I guess the frame trust is set to `prewalked` so that's at least something.

crates/symbolicator/src/services/symbolication.rs
flub: Hum yes.  So if this is huge we should allocate it on the heap?  I'm exploring the theoretical ideas behind this btw, not saying that this is a concern for us in this situation but rather seeing how well I understand rust ;)

Swatinem: `Vec`s (and `HashMap`s) are always on the heap (well, the actual data is). Whats on the stack are just 3 usize: `ptr, len, capacity`.

flub: doh, yes of course.  bad brain.

PR #553, opened 2021-09-08
ref(symbolication): Re-apply 5MB limit to request bodies

relaxolotl:
Does what it says on the tin. This applies a 5MB limit to `/symbolicate`'s request body. If this check fails, an axum `Rejection` is returned which will be automatically converted and returned as an error response.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/endpoints/symbolicate.rs
+++ b/crates/symbolicator/src/endpoints/symbolicate.rs
@@ -50,8 +50,10 @@ pub struct SymbolicationRequestBody {
 pub async fn symbolicate_frames(
     extract::Extension(state): extract::Extension<Service>,
     extract::Query(params): extract::Query<SymbolicationRequestQueryParams>,
-    // TODO: limit body to 5M
-    extract::Json(body): extract::Json<SymbolicationRequestBody>,
+    extract::ContentLengthLimit(extract::Json(body)): extract::ContentLengthLimit<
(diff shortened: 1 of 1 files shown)


crates/symbolicator/src/endpoints/symbolicate.rs
```diff
@@ -50,8 +50,10 @@ pub struct SymbolicationRequestBody {
 pub async fn symbolicate_frames(
     extract::Extension(state): extract::Extension<Service>,
     extract::Query(params): extract::Query<SymbolicationRequestQueryParams>,
-    // TODO: limit body to 5M
-    extract::Json(body): extract::Json<SymbolicationRequestBody>,
+    extract::ContentLengthLimit(extract::Json(body)): extract::ContentLengthLimit<
+        extract::Json<SymbolicationRequestBody>,
+        { 5_000 * 1024 }, // ~5MB
```
Swatinem, line 55:
```suggestion
        { 5 * 1024 * 1024 },
```

Swatinem (approved):
nice find! I must have been blind, lol

flub:
> nice find! I must have been blind, lol

duh, same 🤦 

Not that this will reject requests without content-length header, that's probably fine and probably was the case before as well but something to keep in mind when deploying this.

flub (approved):

PR #555, opened 2021-09-14
fix: Fetch ELF debug files as well just to be safe if it is a Windows minidump

relaxolotl:
~~A first shot at fixing a regression where we're no longer fetching
ELF DIFs for minidumps generated on Windows, likely caused by
#414.~~

Symbolicator now requests all known debug file types when encountering a minidump generated by a native executable. This means that minidumps produced by PEs, Machos and ELFs will trigger a search for PDBs, dSYMs, and ELFs if possible.

### DIFF
--- a/crates/symbolicator/src/services/download/sentry.rs
+++ b/crates/symbolicator/src/services/download/sentry.rs
@@ -21,7 +21,7 @@ use super::{
 };
 use crate::config::Config;
 use crate::sources::SentrySourceConfig;
-use crate::types::{ObjectId, ObjectType};
+use crate::types::ObjectId;
 use crate::utils::futures::{self as future_utils, m, measure};
 
 /// The Sentry-specific [`RemoteDif`].
@@ -205,19 +205,10 @@ impl SentryDownloader {
                 .append_pair("debug_id", &debug_id.to_string());
         }
 
-        // Minidumps on Windows are assigned an object type of ObjectType::Pe, but it's possible
-        // that they're actually Elfs instead
-        let additional_filetype = if object_id.object_type == ObjectType::Pe {
-            Some(FileType::ElfDebug)
-        } else {
-            None
-        };
-
         // See <sentry-repo>/src/sentry/constants.py KNOWN_DIF_FORMATS for these query strings.
         index_url.query_pairs_mut().extend_pairs(
             file_types
                 .iter()
-                .chain(additional_filetype.iter())
                 .map(|file_type| match file_type {
                     FileType::UuidMap => "uuidmap",
                     FileType::BcSymbolMap => "bcsymbolmap",
--- a/crates/symbolicator/src/services/download/sentry.rs
+++ b/crates/symbolicator/src/services/download/sentry.rs
@@ -21,7 +21,7 @@ use super::{
 };
 use crate::config::Config;
 use crate::sources::SentrySourceConfig;
-use crate::types::ObjectId;
+use crate::types::{ObjectId, ObjectType};
 use crate::utils::futures::{self as future_utils, m, measure};
 
 /// The Sentry-specific [`RemoteDif`].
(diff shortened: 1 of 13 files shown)


crates/symbolicator/src/services/download/sentry.rs
```diff
@@ -205,10 +205,19 @@ impl SentryDownloader {
...
         }
 
+        // Minidumps on Windows are assigned an object type of ObjectType::Pe, but it's possible
+        // that they're actually Elfs instead
+        let additional_filetype = if object_id.object_type == ObjectType::Pe {
+            Some(FileType::ElfDebug)
+        } else {
+            None
+        };
+
```
flub, line 215: i fear this is the wrong place to fix this up, i expect it should be fixed in whoever calls this function to pass the right slice of `FileType`s.  but still looking for all the places where this could be caused

flub: https://github.com/getsentry/symbolicator/blob/master/crates/symbolicator/src/sources.rs#L462 might be closer to where this should be done

relaxolotl: In that case I think https://github.com/getsentry/symbolicator/blob/master/crates/symbolicator/src/services/symbolication.rs#L1716 might also be an option. This is assuming we're only seeing this issue with minidumps, though.

flub: I don't think so, `MinidumpState::object_type` is only used in https://github.com/getsentry/symbolicator/blob/87e846b006971167ff1d5af574272cfa0163158d/crates/symbolicator/src/services/symbolication.rs#L1878 which is not used in any code paths ending up calling `.list_files()`.  The other location is the one which ends up being used in `.list_files()`.

relaxolotl:
there are a few integration tests left that i need to update and the changes are enough of a rat's nest that i'm going to put them off until tomorrow

crates/symbolicator/src/sources.rs
```diff
@@ -458,9 +458,33 @@ impl FileType {
     #[inline]
     pub fn from_object_type(ty: ObjectType) -> &'static [Self] {
         match ty {
-            ObjectType::Macho => &[FileType::MachDebug, FileType::MachCode, FileType::Breakpad],
-            ObjectType::Pe => &[FileType::Pdb, FileType::Pe, FileType::Breakpad],
-            ObjectType::Elf => &[FileType::ElfDebug, FileType::ElfCode, FileType::Breakpad],
+            // There are instances where applications have been cross-compiled for some target
```
flub, line 461:
```suggestion
            // There are instances where applications have been compiled for some target
```
I don't think we can claim this is cross-compiling

flub (commented):
oh no, forgot that this would affect snapshots...  but lgtm

crates/symbolicator/src/sources.rs
relaxolotl: fixed 👍

tests/integration/test_basic.py
```diff
@@ -195,19 +218,12 @@ def _make_error_result(download_error, source="microsoft", bucket_type="http"):
                 "source": source,
             },
         ]
-        if bucket_type != "http":
```
relaxolotl, line 198: removed this because i fixed an error where i'd incorrectly left out the layout field on the non-http sources in the permissions tests, causing them to include dSYMs in their output.

tests/integration/test_basic.py
```diff
@@ -27,65 +27,78 @@
...
-            "arch": "x86",
-            "image_addr": "0x749d0000",
-            "image_size": 851_968,
-            "candidates": [
-                {
-                    "download": {"status": "notfound"},
-                    "location": "http://127.0.0.1:1234/msdl/wkernel32.pdb/FF9F9F7841DB88F0CDEDA9E1E9BFF3B51/wkernel32.pd_",
-                    "source": "microsoft",
+
+def _make_successful_result(filtered=False):
```
relaxolotl, line 31: would recommend hiding whitespace changes for this block

relaxolotl:
i fixed the integration tests but my attempts to update snapshots in sentry has left them in pretty bad shape. those will have to be double checked before merging this PR in, but i think the changes here should be good.

flub (approved):
lgtm i think but this will probably need snapshot updates in the sentry repo as well.  We used to commit the two changes close together but are no longer allowed to break master and i don't know what we're doing about this currently.  i'll try and find out

relaxolotl:
based on the [CI results](https://github.com/getsentry/symbolicator/pull/555/checks?check_run_id=3638592400) it looks like sentry doesn't need updating. i think we can infer based off of that that we can leave sentry as is and this should be safe to merge in without changes to that repo. i (and floris) did run some tests locally to be safe, and the results of those tests support the conclusion that no snapshot updates are needed in sentry.

PR #558, opened 2021-09-18
ref: Minor integration test changes

relaxolotl:
This makes a few quality of (my) life changes to the integration tests so it's a little easier to update and customize them as needed. I've had to muck around in this code a few times now and I'm hoping these changes make it easier to understand what's going on, and how to change the test code in response to a change in symbolicator. 

A major motivation behind the changes is to try to structure things in a way so that the tests don't need to mutate objects too much in order to get the correct expected data to assert against.

- Consolidate common parts of the stack trace and the modules into shared variables
- Merge redundant helpers that construct results for failed symbolication attempts
- Prefer passing in specific configs for error results as opposed to referring to named constants that pre-fill parameters

#skip-changelog

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -271,7 +271,7 @@ def hitcounter(request):
         yield app
 
 
-def assert_symbolication(output, expected):
+def assert_symbolication(output, expected, assertion_info=None):
     """Compares symbolication results, with redactions.
 
     Redactions are necessary to remove random port numbers.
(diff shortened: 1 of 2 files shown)


flub (approved):
This code has indeed always been pretty horrible.  Long term we'd like to migrate these slowly into tests written in rust.  And mostly move them to much more targeted/unit tests with only a few remaining as full integration tests.

PR #559, opened 2021-09-20
ref: Expose the sentry HTTP headers in the context, not tags

flub:
Tags are indexed and we don't need to be able to search on these.
Plus adding a `sentry.project_id` in the tags is highly confusing wrt
request.scope.  Originally this stuff was exposed in the context as
part of the http headers, so lets at least do this partially again.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/utils/sentry.rs
+++ b/crates/symbolicator/src/utils/sentry.rs
@@ -1,3 +1,4 @@
+use std::collections::BTreeMap;
 use std::task::{Context, Poll};
 
 use axum::body::Body;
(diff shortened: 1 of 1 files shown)


loewenheim (approved):

Swatinem:
> Tags are indexed and we don't need to be able to search on these.

I argue that we actually do want to!

We want to be able to correllate events in discover with project ids, and I would love to be able to look up internal errors via a specific event ID.

What prompted this change?

flub:
the prompt of this change was the `sentry.project_id` tag which was entirely confusing and made no sense to anyone looking at it.  It conflicted with `request.scope` which is supposed to be the project for which this symbolication request is made.

Removing all headers from tags was then done because the old code didn't have this as tags either and i had no idea what we gain from having it in tags.  If I understand you correctly you only really care about the event_id though?  This is event ID for which we are symbolicating is it?

Swatinem:
It does look like a uuid, so I guess it is.

So if there is a discrepancy between `project_id` and `scope`, might be a good idea to figure out what the difference is, because I too don’t know!

`event_id` would be especially nice because you can search for it, to get *all* the errors for a particular request/minidump. For the `worker_id`, I really don’t care. The `trace_id` might be interesting once we start to actively working on that.

flub:
I don't think we should need to manually plug `trace_id` in, the performance produce of sentry should just take care of this, it already shows related traces to issues and issues related to traces.

Agree that `project_id` could be interesting to know where that actually comes from, I assumed it was the project that the tracing is using, but maybe that's not really true.

And finally I guess i'd be fine with adding `event_id` back as a tag :)

PR #560, opened 2021-09-20
doc: Update and expand instructions on how to update test snapshots on sentry

relaxolotl:
Just a small update on the snapshot instructions to take into account the new integration tests that have been already added to CI. Also includes an additional expansion of some of the existing instructions to provide context on what is being done as well as some gotchas that can only be found if you rummage around in code in other repositories.

#skip-changelog

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -60,11 +60,16 @@ Sentry runs a symbolicator integration test suite located at [\`tests/symbolicat
 Follow these steps to update snapshots in Sentry:
 
 1. Check out latest Sentry \`master\` and enable the virtualenv.
-2. Stop the symbolicator devservice using \`sentry devservices down symbolicator\`.
-3. Run your development symbolicator on port \`3021\`.
-4. Export \`SENTRY_SNAPSHOTS_WRITEBACK=1\` and run symbolicator tests with pytest.
-5. Review snapshot changes locally, then create a PR to Sentry.
-6. Merge the Symbolicator PR, then merge the Sentry PR.
+2. Make sure your other devservices are running via \`sentry devservices up\`, and that symbolicator
+is enabled (\`symbolicator: true\`) in sentry via \`~/.sentry/config.yml\`.
+4. Stop the symbolicator devservice using \`sentry devservices down symbolicator\`. You want to use your
+own development symbolicator to update the snapshots.
+5. Run your development symbolicator on port \`3021\`, or whatever port symbolicator is configured to use
+in \`~/.sentry/config.yml\`.
+6. Export \`SENTRY_SNAPSHOTS_WRITEBACK=1\` to automatically update the existing snapshots with your new
+results and run symbolicator tests with pytest (\`pytest tests/symbolicator\`).
+7. Review snapshot changes locally, then create a PR to Sentry.
+8. Merge the Symbolicator PR, then merge the Sentry PR.
 
 </details>
   `;
@@ -90,8 +95,8 @@ async function checkSnapshots() {
 
   if (changesSnapshots) {
     warn(
-      "Snapshot changes likely affect Sentry tests. Please check the symbolicator test suite in " +
-        "Sentry and update snapshots as needed."
+      "Snapshot changes likely affect Sentry tests. If the Sentry-Symbolicator Integration Tests in CI are " +
+      "failing for your PR, please check the symbolicator test suite in Sentry and update snapshots as needed."
     );
     markdown(getSnapshotDetails());
   }
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -60,16 +60,16 @@ Sentry runs a symbolicator integration test suite located at [\`tests/symbolicat
 Follow these steps to update snapshots in Sentry:
 
 1. Check out latest Sentry \`master\` and enable the virtualenv.
-2. Make sure your other devservices are running via \`sentry devservices up\`, and that symbolicator
-is enabled (\`symbolicator: true\`) in sentry via \`~/.sentry/config.yml\`.
-4. Stop the symbolicator devservice using \`sentry devservices down symbolicator\`. You want to use your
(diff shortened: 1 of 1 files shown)


dangerfile.js
```diff
@@ -60,11 +60,16 @@ Sentry runs a symbolicator integration test suite located at [\`tests/symbolicat
...
+2. Make sure your other devservices are running via \`sentry devservices up\`, and that symbolicator
+is enabled (\`symbolicator: true\`) in sentry via \`~/.sentry/config.yml\`.
+4. Stop the symbolicator devservice using \`sentry devservices down symbolicator\`. You want to use your
+own development symbolicator to update the snapshots.
+5. Run your development symbolicator on port \`3021\`, or whatever port symbolicator is configured to use
+in \`~/.sentry/config.yml\`.
+6. Export \`SENTRY_SNAPSHOTS_WRITEBACK=1\` to automatically update the existing snapshots with your new
+results and run symbolicator tests with pytest (\`pytest tests/symbolicator\`).
+7. Review snapshot changes locally, then create a PR to Sentry.
+8. Merge the Symbolicator PR, then merge the Sentry PR.
```
flub, line 72: this i'm still not sure about how this was meant to be these days.  I recall some discussions on how we should never make sentry master red but the final solution was tricky.  Maybe @jan-auer or @armenzg remember what our interim solution is?

flub (approved):

dangerfile.js
armenzg: We added Sentry integration tests [1]. 

[1] https://github.com/getsentry/symbolicator/commit/2da1f94aa9ff13f073b74b8d9434703c06eb4ee4#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f

relaxolotl: @armenzg would you happen to know what the workflow is supposed to be to merge symbolicator changes that would cause the sentry integration tests to fail? 

the symbolicator PR would presumably be red because it isn't passing against sentry on master, and the sentry PR to update the integration tests would be failing because they're not passing against symbolicator on master. 

is there a way to merge this theoretical breaking change and its integration test updates without turning sentry on master red for a brief period of time, or is it expected and acceptable that such a change would turn it red?

dangerfile.js
```diff
@@ -60,11 +60,16 @@ Sentry runs a symbolicator integration test suite located at [\`tests/symbolicat
 Follow these steps to update snapshots in Sentry:
 
 1. Check out latest Sentry \`master\` and enable the virtualenv.
-2. Stop the symbolicator devservice using \`sentry devservices down symbolicator\`.
-3. Run your development symbolicator on port \`3021\`.
-4. Export \`SENTRY_SNAPSHOTS_WRITEBACK=1\` and run symbolicator tests with pytest.
-5. Review snapshot changes locally, then create a PR to Sentry.
-6. Merge the Symbolicator PR, then merge the Sentry PR.
+2. Make sure your other devservices are running via \`sentry devservices up\`, and that symbolicator
+is enabled (\`symbolicator: true\`) in sentry via \`~/.sentry/config.yml\`.
```
billyvg, line 64: The config seems important enough (esp if it's not default) to warrant its own line item.

dangerfile.js
```diff
@@ -60,11 +60,16 @@ Sentry runs a symbolicator integration test suite located at [\`tests/symbolicat
...
+4. Stop the symbolicator devservice using \`sentry devservices down symbolicator\`. You want to use your
```
billyvg, line 65: I think [you can use `--exclude`](https://github.com/getsentry/sentry/blob/d15e14f01c597353bc40972043a83df11fa99999/src/sentry/runner/commands/devservices.py#L153) to start devservices w/o symbolicator

billyvg (commented):
drive-by

dangerfile.js
armenzg: As far as I understand, the current situation would look like this:
* The engineers open a symbolicator PR that makes breaking changes (symbolicator PR is :red_circle: )
* PR opened on Sentry with new snapshots (Sentry-new-snapshots integration tests PR are :red_circle: )
* symbolicator gets merged (symbolicator master is :red_circle: )
* Wait for symbolicator image to be pushed
  * From here on, Sentry master and all but one PR would show :red_circle: integration tests if the Sentry-new-snapshots PR does not get merged
* sentry PR can be re-run (sentry PR turns :green_circle: )
* sentry PR gets merged (sentry turns 🟢   and symbolicator integreation test runs will now be :green_circle: )

relaxolotl: For a starting set of participants, both myself and @flub would be good people to include. @flub is currently out on vacation until end of the 27th (next Tues, Sep 28 is the first day he'll be back).

I see you've already started a conversation on slack about this so I'll continue discussion about the logistics of this call in there. 👍

dangerfile.js
relaxolotl: good point, i've split this out 👍

billyvg (approved):

Swatinem:
I tried following these instructions, but I run into this problem here:
```
 ERROR symbolicator::services::objects > Failed to fetch file list from sentry: failed to fetch data from Sentry
  caused by: failed sending request to Sentry
  caused by: error sending request for url (http://host.docker.internal:56997/api/0/projects/baz/bar/files/dsyms/?debug_id=ab880c84-1493-f0f4-250a-cf143001b64c-1&file_formats=sourcebundle&code_id=731653ad322000): error trying to connect: dns error: no record found for name: host.docker.internal.home. type: AAAA class: IN
  caused by: error trying to connect: dns error: no record found for name: host.docker.internal.home. type: AAAA class: IN
  caused by: dns error: no record found for name: host.docker.internal.home. type: AAAA class: IN
  caused by: no record found for name: host.docker.internal.home. type: AAAA class: IN
```

The tests seem to be running from a docker container, and that uses its own URL as the `sources` config when sending requests to my local symbolicator, which then can’t access stuff inside the docker container, hence I can’t regenerate snapshots locally.

PR #563, opened 2021-10-05
fix: Output correct function for client-side frames (NATIVE-150)

Swatinem:
This uses the `function` property instead of `symbol`, which is ultimately what sentry uses.
It also avoids flagging modules as `unwind: missing` which were included in client-walked stacks, but which were *really* missing in the previous breakpad-based walk.

Additionally, it moves the error-reporting of invalid minidumps *after* the breakpad-based walk, to not log errors in case breakpad isnt able to process the minidump either.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/symbolication.rs
+++ b/crates/symbolicator/src/services/symbolication.rs
@@ -338,7 +338,8 @@ impl ModuleListBuilder {
         for trace in stacktraces {
             for frame in &trace.frames {
                 let addr = frame.instruction_addr.0;
-                self.mark_referenced(addr);
+                let is_prewalked = frame.trust == FrameTrust::Prewalked;
+                self.mark_referenced(addr, is_prewalked);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


loewenheim (approved):

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -2074,8 +2068,15 @@ impl SymbolicationActor {
...
-                merge_clientside_with_processed_stacktraces(&mut stacktraces, client_stacktraces);
+            match client_stacktraces {
+                Ok(Some(client_stacktraces)) => merge_clientside_with_processed_stacktraces(
+                    &mut stacktraces,
+                    client_stacktraces,
+                ),
+                Err(e) => {
+                    log::error!("invalid minidump extension: {}", e);
+                }
+                _ => {}
```
relaxolotl, line 2079: do we still have a way to track stack traces that breakpad isn't able to parse with this change? is that useful or meaningful info for us?

Issue #564, opened 2021-10-07
wasm_split release for MacOS


bkotsopoulossc:
The repo currently releases builds of the `wasm_split` tool on linux, however it seems feasible to build and release it for MacOS as well. Is there interest in supporting this, and what would it take to do so? Thanks in advance!

jan-auer:
Thanks, we'd definitely be open to build and release for macOS, too! Would you be up for a hacktoberfest contribution? :) 

The build workflow is located here: https://github.com/getsentry/symbolicator/blob/654086c9247f8915ca0b987645c618e7ac5eb598/.github/workflows/build.yml#L9-L48

Building and releasing macOS requires a second job running on `macos-latest` with the post-build steps updated:
- Remove objcopy, `dsymutil` is run automatically by cargo
- Change paths to publish dSYMs too

This would end up looking similar to [what we do in Relay](https://github.com/getsentry/relay/blob/c2daf430b8efa1d48ad7b2a7c666712ec51a3c13/.github/workflows/build_binary.yml#L35-L67).

bkotsopoulossc:
@jan-auer sure thing, can you give me permissions to run github workflows so I can iterate a bit?
![image](https://user-images.githubusercontent.com/25588019/136586000-bb60e3b6-ef9e-4745-9ace-09ee3eb4b01a.png)

PR #565, opened 2021-10-08
build(ci): Public MacOS x86_64 binaries to github releases

bkotsopoulossc:
closes https://github.com/getsentry/symbolicator/issues/564

PTAL @jan-auer 

I have mostly tested this locally on mac but I don't have permissions to run the github workflows yet so I haven't tested end to end in CI. I imagine a nice way to test this would be to push a commit where we edit the build.YML to run on this branch, but maybe comment out the publishing step and instead just `ls` the directory. And then revert that if all good.


bkotsopoulossc:
Good callout. I can look into that in a follow up

PR #566, opened 2021-10-11
ref: Update deps and remove dead-stores

Swatinem:
Does a `cargo update` in addition to bumping `lru`.

Also satisfied the nightly compiler, which got a bit smarter flagging dead-stores.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/cficaches.rs
+++ b/crates/symbolicator/src/services/cficaches.rs
@@ -215,7 +215,6 @@ impl CacheItemRequest for FetchCfiCacheInternal {
         );
 
         CfiCacheFile {
-            _data,
             features: self.meta_handle.features(),
             status,
             path,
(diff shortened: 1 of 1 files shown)


crates/symbolicator/src/services/cficaches.rs
```diff
@@ -87,10 +87,9 @@ impl CfiCacheActor {
 
 #[derive(Debug)]
 pub struct CfiCacheFile {
-    object_type: ObjectType,
-    identifier: ObjectId,
-    scope: Scope,
-    data: ByteView<'static>,
+    // NOTE: while this really is a dead store, we keep it since the `ByteView` should keep the
+    // underlying file alive via its mmap mapping.
+    _data: ByteView<'static>,
```
Swatinem, line 92: Is this assumption right? Even if so, linux still allows `unlink`, so even if the file itself is still there (in limbo), the procspawn-breakpad sub-process could still fail to open that file, right?

loewenheim (approved):

PR #573, opened 2021-11-03
feat: Add a shared second-layer Cache to symbolicator (NATIVE-248)

Swatinem:


crates/symbolicator/src/services/cacher.rs
```diff
@@ -446,7 +515,7 @@ impl<T: CacheItemRequest> Cacher<T> {
         // cache_path is None when caching is disabled.
         let cache_path = self.get_cache_path(&key, T::VERSIONS.current);
         if let Some(ref path) = cache_path {
-            if let Some(item) = self.lookup_cache(&request, &key, path)? {
+            if let Some(item) = self.lookup_local_cache(&request, &key, path)? {
```
relaxolotl, line 518: if i understand this correctly, this means that we're actually doing the local lookup first, and then falling back to the shared cache if there isn't a hit there. 

is that what the code is doing? if so, aren't the fallbacks reversed in this case, since we want to fall back to local if shared doesn't have a hit?

relaxolotl: an additional tiny nit is that perhaps we would want to revisit the name of `compute_memoized` and `compute` if one of the major differences is the cache that's being used.

crates/symbolicator/src/cache.rs
```diff
@@ -117,6 +121,56 @@ impl CacheStatus {
...
 }
 
+/// A remote cache that can be shared between symbolicator instances.
+#[derive(Debug, Clone)]
+pub struct SharedCache {
+    config: SourceConfig,
+    downloader: Arc<DownloadService>,
+}
+
+impl SharedCache {
```
relaxolotl, line 131: very high level mulling; one of the things i found really difficult to keep straight in my brain when i was working on caches was that there was an implicit categorization of caches: the "data" caches and the derived caches. 

if i understand correctly, this basically means that after this PR now have the following:

### data cache
- object cache
- bitcode?
- this new shared cache

### derived cache
- symcache
- cficache

while there is some code shared between every item listed above irrespective of their "category" by virtue of them all being caches, i feel that their category also has a meaningful effect on how and where they are used. 

i kind of wish the code itself had a way to explicitly declare this categorization, and the addition of a new type of cache i feel makes it more important to do so. i don't think it really fits within the scope of this PR, or possibly even this project; just hoping that this sparks some discussion. food for thought, let's say.

crates/symbolicator/src/services/download/gcs.rs
```diff
@@ -300,6 +300,84 @@ impl GcsDownloader {
         }
     }
 
+    /// Uploads a file to GCS.
+    pub async fn upload(
+        &self,
+        destination: GcsRemoteDif,
+        contents: &[u8],
+    ) -> Result<DownloadStatus, DownloadError> {
```
relaxolotl, line 308: hah, i assume the naming on this is one of the awkward quirks that comes from reusing what's already present to add in upload functionality.

as a small nit, i would like to ask that before this is merged that a second sweep over the logs to is performed to at least ensure that those match the actual action here: the logs right now say things like "insufficient permissions to download x" when it should be "insufficient permissions to upload x".

crates/symbolicator/src/services/cacher.rs
Swatinem: > we want to fall back to local if shared doesn't have a hit

we always want to look into the local cache first, since that is the quickest. It just goes straight to sync fs APIs, and they are supposed to be very quick.
the *second layer* is going via network to a shared cache, which might take a longer time since it needs to download stuff.
only after that, we try recomputing from the source.

crates/symbolicator/src/cache.rs
Swatinem: the distinction between data and derived cache is only in the way they are handling object expiry. the shared cache is completely agnostic about that and does not fit into any of those categories.

but on the topic of expiration: the idea was that we only cache "positive" downloads/computations, and have a TTL that is equal to our normal data retention periods. That way, we won’t have to worry about expiration at all.

Currently the different forms of caches have different settings there, and also different settings for positive vs negative (not found, error, etc). we focus only on one case here.

crates/symbolicator/src/services/download/gcs.rs
Swatinem: oh absolutely, the logs here need updating for sure! And yes, reusing all the types that have "download" in their name for uploads is probably the biggest gripe I have with the solution so far. Though I also don’t have any better ideas at the moment.

crates/symbolicator/src/services/cacher.rs
relaxolotl: looks like i had it backwards, thanks for the explanation!

Swatinem:
this was superceded by a different approach

PR #574, opened 2021-11-03
ref: Avoid using owned PathBuf if a &Path is enough

Swatinem:
While seeing this, I became curious how much stuff we needlessly clone all the time.

#skip-changelog


flub (approved):
As you remark yourself in a comment, this comes from an architecture where we used to spawn much more across different executors.  As we evolve the architecture to involve fewer and better defined executors (that's still not finished ;)) there will still be such cleanups around.

Swatinem:
Absolutely, I bet there is a lot more such opportunities.

PR #578, opened 2021-11-10
Use rust:bullseye-slim as base image

aminvakil:
I have tested and this gets build fine, but I don't know how to test the functionality of docker images.

This was first brought up in getsentry/onpremise#1128.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 0.5.0
+## Unreleased
 
 ### Features
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.5.0
+
+### Features
+
+- Use `rust:bullseye-slim` as base image for docker container builder. ([#578](https://github.com/getsentry/symbolicator/pull/578))
+- Use `debian:bullseye-slim` as base image for docker container runner. ([#578](https://github.com/getsentry/symbolicator/pull/578))
+
 ## 0.4.0
 
 ### Features
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 FROM getsentry/sentry-cli:1 AS sentry-cli
-FROM debian:stretch-slim AS symbolicator-build
+FROM debian:buster-slim AS symbolicator-build
 
 WORKDIR /work
 
@@ -53,7 +53,7 @@ RUN sentry-cli --version \
 # Copy the compiled binary to a clean image #
 #############################################
 
-FROM debian:stretch-slim
+FROM debian:buster-slim
 RUN apt-get update \
     && apt-get install -y --no-install-recommends openssl ca-certificates gosu cabextract \
     && rm -rf /var/lib/apt/lists/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ COPY Cargo.toml Cargo.lock ./
 COPY crates/symbolicator/build.rs crates/symbolicator/Cargo.toml crates/symbolicator/
 
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends libssl-dev pkg-config \
+    && apt-get install -y --no-install-recommends libssl-dev pkg-config g++ \
     && rm -rf /var/lib/apt/lists/*
 # Build without --locked.
 #
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ COPY Cargo.toml Cargo.lock ./
 COPY crates/symbolicator/build.rs crates/symbolicator/Cargo.toml crates/symbolicator/
 
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends libssl-dev pkg-config g++ git \
+    && apt-get install -y --no-install-recommends libssl-dev pkg-config g++ git zip \
     && rm -rf /var/lib/apt/lists/*
 # Build without --locked.
 #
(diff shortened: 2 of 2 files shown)


aminvakil:
Sorry, I was renaming branch name in https://github.com/getsentry/symbolicator/pull/577 and it got closed :)

aminvakil:
I guess this has been fixed now: https://github.com/getsentry/symbolicator/pull/140

flub:
IIRC we had memory regressions running this.  We've recently been thinking of directly going to bullseye really, any reason you want buster?

flub:
Another thing we would like to get out of upgrading the debian image is that we can directly depend on the upstream rust container instead of installing rust ourself.  The reason for this is that we currently have the rustc version number baked into the dockerfile but we'd prefer if our docker files automatically pull the latest stable rustc version as this is what we use on CI and in development.

aminvakil:
> IIRC we had memory regressions running this. We've recently been thinking of directly going to bullseye really, any reason you want buster?

@flub I was wondering if you are fine switching directly to bullseye from stretch and might prefer having buster for a while and then switch to bullseye, but if you think switching directly to bullseye is fine, I would change this PR.

> Another thing we would like to get out of upgrading the debian image is that we can directly depend on the upstream rust container instead of installing rust ourself. The reason for this is that we currently have the rustc version number baked into the dockerfile but we'd prefer if our docker files automatically pull the latest stable rustc version as this is what we use on CI and in development.

Exactly, I saw the comment in Dockerfile and I'm trying to build it on my local machine using `rust:slim-buster` (will change this to `rust:slim-bullseye`), but have some problem because of `openssl` missing, I'll push it as soon as it's fixed.

flub:
> Exactly, I saw the comment in Dockerfile and I'm trying to build it on my local machine using `rust:slim-buster` (will change this to `rust:slim-bullseye`), but have some problem because of `openssl` missing, I'll push it as soon as it's fixed.

`rust:slim-bullseye` sounds like the right approach, thanks!

aminvakil:
I would appreciate if a maintainer runs the workflow...

aminvakil:
`ca-certificates` and `curl` are not necessary anymore, as this is not going to request a web server on https anymore.
Also `g++` is enough in this docker image and installing all packages which `build-essential` is depending on is unnecessary (although this is a builder image and will not affect the result image, but still).

flub:
> I would appreciate if a maintainer runs the workflow...

i can't believe github still hasn't fixed this madness :(

aminvakil:
I'm not sure about the next release version number, so I put 0.5.0 as it has a new feature.

And please run the workflow.
(When you've approved it once in this PR, it should not need approval again, right? What's the reason for asking it every time?)

flub:
> I'm not sure about the next release version number, so I put 0.5.0 as it has a new feature.

Could you change this to "Unreleased" instead?  Our release-automation will replace that when we ask it to make a release.

> And please run the workflow. (When you've approved it once in this PR, it should not need approval again, right? What's the reason for asking it every time?)

who knows

You may have noticed I changed base branch to `release/deploy-docker-bullseye`.  This makes our infrastructure build the images for our prod.  we'll have to try out running an experimental instance of this for a while serving some traffic to see if it behaves correctly before merging to master.  But thanks for this work!  wouldn't have come round to doing this yet otherwise ;)

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.5.0
```
flub, line 3:
```suggestion
## Unreleased
```

flub (approved):

aminvakil:
> > I'm not sure about the next release version number, so I put 0.5.0 as it has a new feature.
> 
> Could you change this to "Unreleased" instead? Our release-automation will replace that when we ask it to make a release.

Well, you've already done that, anything else? :)
> 
> 
> You may have noticed I changed base branch to `release/deploy-docker-bullseye`. This makes our infrastructure build the images for our prod. we'll have to try out running an experimental instance of this for a while serving some traffic to see if it behaves correctly before merging to master. But thanks for this work! wouldn't have come round to doing this yet otherwise ;)

Great, I hope this works and we can get rid of jessie.

aminvakil:
Is this OK?
https://github.com/getsentry/symbolicator/runs/4176331756?check_suite_focus=true#step:10:195

flub:
> Is this OK? https://github.com/getsentry/symbolicator/runs/4176331756?check_suite_focus=true#step:10:195

probably, sentry tests have noise like this sadly :( i don't think this is symbolicator-related (the way the output is structured is super weird too, not sure what the sentry CI does to pytest) the entire suite still passes so probably fine

flub:
I'm afraid this isn't looking so well, memory usage seems to again keep going up here:
image

I'm afraid we'd have to understand what is going on here, it seems to be triggered by simply upgrading the base image.  I'm sure this must be solvable though, people sure have been running things on debian stable.  Somehow our code must do something weird somewhere.

aminvakil:
I was thinking from yesterday about how can I help investigating the issue and couldn't come to anything :(

Are there scripts which can be run to make symbolicator docker container gets to work without sentry? I don't have that much resources to handle all sentry.

How do you think I can help?

flub:
@aminvakil hey, i've also been wondering the same question.  And finally came up with some ideas that could help this move forward: It would be interesting if this is reproducible outside of our prod infra.  e.g. would it be possible to set up something very simple like set up a script which spins up the container and sends some of our minidumps in the test suite to it with curl, in a loop.  Even though that is the same minidump over and over it would still be interesting to see if that keeps increasing the memory or not.  Once that exist it might be interesting to see if symbolicator can then run under something like `cargo-valgrind` inside of the container and maybe it finds something interesting.

aminvakil:
Do you want me to open a new issue to keep track of this?

flub:
> Do you want me to open a new issue to keep track of this?

yeah, that's probably a better idea instead of keeping commenting on a closed PR 😄 
thanks!

PR #583, opened 2021-11-25
fix: Clean up some Hub/Scope manipulation

Swatinem:
This fixes one bug with an ineffective async with_scope,
otherwise it removes some explicit `bind_hub` calls for async functions which are being
`await`ed immediately, cleaning up some Vec allocations while doing so.

This keeps all the `bind_hub` calls that involve spawning futures onto the runtime, or functions that return boxed futures (are not async themselves).
If I understand things correctly, if we `await` futured directly (even with the `join_all` combinator), all of that polling is done in the current thread, so the current thread hub is used for those anyway. No need to explicitly bind another hub. Only spawning things onto the executor would "leave" the current execution context/thread.

#skip-changelog ffs

### DIFF
--- a/crates/symbolicator/src/services/bitcode.rs
+++ b/crates/symbolicator/src/services/bitcode.rs
@@ -264,7 +264,7 @@ impl BitcodeService {
     ) -> Result<Option<Arc<CacheHandle>>, Error> {
         let jobs = sources.iter().map(|source| {
             self.fetch_file_from_source(uuid, dif_kind, scope.clone(), source.clone())
-            .bind_hub(Hub::new_from_top(Hub::current()))
+                .bind_hub(Hub::new_from_top(Hub::current()))
         });
         let results = future::join_all(jobs).await;
         let mut ret = None;
@@ -319,7 +319,9 @@ impl BitcodeService {
                 download_svc: self.download_svc.clone(),
                 cache: self.cache.clone(),
             };
-            self.cache.compute_memoized(request).bind_hub(Hub::new_from_top(Hub::current()))
+            self.cache
+                .compute_memoized(request)
+                .bind_hub(Hub::new_from_top(Hub::current()))
         });
 
         let all_results = future::join_all(fetch_jobs).await;
--- a/crates/symbolicator/src/services/bitcode.rs
+++ b/crates/symbolicator/src/services/bitcode.rs
@@ -264,6 +264,7 @@ impl BitcodeService {
     ) -> Result<Option<Arc<CacheHandle>>, Error> {
         let jobs = sources.iter().map(|source| {
             self.fetch_file_from_source(uuid, dif_kind, scope.clone(), source.clone())
+            .bind_hub(Hub::new_from_top(Hub::current()))
         });
         let results = future::join_all(jobs).await;
         let mut ret = None;
@@ -318,7 +319,7 @@ impl BitcodeService {
                 download_svc: self.download_svc.clone(),
                 cache: self.cache.clone(),
             };
-            self.cache.compute_memoized(request)
+            self.cache.compute_memoized(request).bind_hub(Hub::new_from_top(Hub::current()))
         });
 
         let all_results = future::join_all(fetch_jobs).await;
--- a/crates/symbolicator/src/services/bitcode.rs
+++ b/crates/symbolicator/src/services/bitcode.rs
@@ -262,31 +262,19 @@ impl BitcodeService {
         scope: Scope,
         sources: Arc<[SourceConfig]>,
     ) -> Result<Option<Arc<CacheHandle>>, Error> {
-        sentry::with_scope(
-            |scope| {
-                scope.set_tag("auxdif.debugid", uuid);
-                scope.set_extra("auxdif.kind", dif_kind.to_string().into());
-            },
-            || async {
-                let mut jobs = Vec::with_capacity(sources.len());
-                for source in sources.iter() {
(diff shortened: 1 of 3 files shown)


crates/symbolicator/src/services/bitcode.rs
```diff
@@ -262,31 +262,19 @@ impl BitcodeService {
...
-                    }
-                }
-                Ok(ret)
-            },
-        )
-        .await
+        let jobs = sources.iter().map(|source| {
+            self.fetch_file_from_source(uuid, dif_kind, scope.clone(), source.clone())
+        });
+        let results = future::join_all(jobs).await;
```
flub, line 268: So I think the premise of this PR is a little wrong and whenever you do go into a number of futures which are running concurrently, like with `future::join_all`, you need to give them each a new hub.  The reason is that they call can be pushing and popping scopes, change tags and context on scopes etc.  Since all the futures run concurrently they would be influencing each other a little too much.

So while not a regression here, when creating the jobs here they should all have received a new hub.

crates/symbolicator/src/services/bitcode.rs
```diff
@@ -328,9 +318,8 @@ impl BitcodeService {
                 download_svc: self.download_svc.clone(),
                 cache: self.cache.clone(),
             };
-            let job = self.cache.compute_memoized(request).bind_hub(hub.clone());
-            fetch_jobs.push(job);
-        }
+            self.cache.compute_memoized(request)
```
flub, line 321: I think this should still be binding a cloned hub

crates/symbolicator/src/services/objects/mod.rs
```diff
@@ -263,10 +260,7 @@ impl ObjectsActor {
                         Vec::new()
                     })
             }
-            .bind_hub(Hub::new_from_top(Hub::current()));
-
-            queries.push(query);
-        }
+        });
```
flub, line 263: i think this bind hub is also needed as this goes into `future::join_all`

crates/symbolicator/src/services/objects/mod.rs
```diff
@@ -311,12 +303,10 @@ impl ObjectsActor {
...
                 meta_cache
                     .compute_memoized(request)
-                    .bind_hub(sentry::Hub::new_from_top(sentry::Hub::current()))
                     .await
                     .map_err(|error| CacheLookupError { file_source, error })
-            };
-            queries.push(query);
-        }
+            }
+        });
```
flub, line 309: i think the hub binding was kind of done in the wrong place here, should have been on the entire `async move { }` block

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1818,11 +1816,8 @@ impl SymbolicationActor {
                     })
                     .await;
                 ((*module_id).to_owned(), result)
-            };
-
-            // Clone hub because of join_all concurrency.
-            futures.push(fut.bind_hub(Hub::new_from_top(Hub::current())));
-        }
+            }
+        });
```
flub, line 1820: this should have stayed i think

flub (changes requested):

crates/symbolicator/src/services/bitcode.rs
```diff
@@ -300,8 +288,12 @@ impl BitcodeService {
         scope: Scope,
         source: SourceConfig,
     ) -> Result<Option<Arc<CacheHandle>>, Error> {
-        let hub = Arc::new(Hub::new_from_top(Hub::current()));
-        hub.configure_scope(|scope| scope.set_extra("auxdif.source", source.type_name().into()));
+        let _guard = Hub::current().push_scope();
```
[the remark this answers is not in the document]
Swatinem: agreed, though not sure how doable that would be considering the internals :-D

Swatinem:
I re-added hub binding for all the `join_all` that we do, as well as another place that I found was missing it, and optimized some more manual vec collecting.

flub (approved):

flub:
>  and optimized some more manual vec collecting.

You seem to suggest that the `.iter().map()` versions you write are more efficient than the existing `Vec::with_capacity()` things.  Is this true?  To me it seems (almost) entirely a style choice.

crates/symbolicator/src/services/bitcode.rs
```diff
@@ -262,31 +262,20 @@ impl BitcodeService {
         scope: Scope,
         sources: Arc<[SourceConfig]>,
     ) -> Result<Option<Arc<CacheHandle>>, Error> {
-        sentry::with_scope(
```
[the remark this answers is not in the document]
Swatinem: > I don't fully understand yet what the problem with that was.

The `with_scope` only keeps the scope alive for the duration of the closure. The closure was returning a future, which was polled somewhere else, so the scope would not be alive for that anymore.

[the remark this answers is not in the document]
flub: 👍  🤦

Swatinem:
> You seem to suggest that the `.iter().map()` versions you write are more efficient than the existing `Vec::with_capacity()` things. Is this true?

My assumption is based around the fact `join_all` takes an `IntoIterator` and uses that internally: https://docs.rs/futures-util/0.3.18/src/futures_util/future/join_all.rs.html#102-105

Instead of iterating through and collecting to a vec, which is then being consumed via `into_iter`, the new variant will avoid that intermediate allocation/collection.

PR #581, opened 2021-11-15
Feat/shared cache mk2

flub:
Start of shared cache implementation.

TODO:
- [x] Cache service
- [x] Non-blocking APIs
- [x] Non-copying APIs
- [x] Handle GCP not found correctly
- [x] Tests

NATIVE-248

PR #586, opened 2021-11-30
fix(cache): Add back in the marker at the beginning of malformed cache entries

relaxolotl:
Primarily fixes a regression from the shared cache PR (https://github.com/getsentry/symbolicator/pull/581) which was no longer writing the sentinel/"header" for malformed entries in caches.

Also includes a minor fix that ensures that other non-positive entries in the cache are truncated in case they are overwriting some other content that is slightly longer than their new value. This is highly unlikely to occur, but this is added in as a safeguard if this _does_ ever occur.

This also renames the helper because there's no need for the user to know about the existence of markers at the abstraction level where that would be invoked.

Unit tests have been written because they would have caught this error assuming that they weren't deleted or commented out during the original function's refactor. Proper integration tests would most likely test `compute` itself, which is a pretty complicated function to test.

### DIFF
--- a/crates/symbolicator/src/cache.rs
+++ b/crates/symbolicator/src/cache.rs
@@ -1199,4 +1199,170 @@ mod tests {
         assert!(!cficaches_entry.is_file());
         assert!(!diagnostics_entry.is_file());
     }
+
+    #[tokio::test]
+    async fn test_cache_status_write_positive() -> Result<()> {
+        let dir = tempdir()?;
+        create_dir_all(&dir).unwrap();
+        let path = dir.path().join("honk");
+
+        // copying what compute does here instead of just using
+        // tokio::fs::File::create() directly
+        let sync_file = File::create(&path)?;
+        let mut async_file = tokio::fs::File::from_std(sync_file);
+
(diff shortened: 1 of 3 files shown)


crates/symbolicator/src/cache.rs
```diff
@@ -1198,4 +1207,256 @@ mod tests {
...
+        status.write(&mut async_file).await?;
+
+        // make sure write leaves the cursor at the end
+        let current_pos = async_file.stream_position().await?;
+        assert_eq!(current_pos, 4);
+
+        // make sure write doesn't change the pre-existing contents of the file
+        // rewinding and then reading to the end of async_file throws bad file
+        // descriptor errors, so just reopen the file
+        let mut final_file = File::open(&path)?;
```
Swatinem, line 1237: I’m wondering why though? This reads like a bug in itself.

Swatinem (approved):

flub (approved):
Looks good, especially the tests.  i'm also shocked this wasn't caught by integration tests, but that doesn't have to be a blocker for this fix.

PR #589, opened 2021-11-30
ref(shared-cache): Counts bytes to/from shared cache

flub:
This adds some counters for bytes, on the metrics-side of things we
can turn that into a rough transfer rate to give us some idea of how
much we're transferring.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/shared_cache.rs
+++ b/crates/symbolicator/src/services/shared_cache.rs
@@ -5,7 +5,9 @@
 //! lives closer to symbolicator. Expensive computations related to the computation of derived
 //! caches may also be saved via this shared cache.
 
+use std::convert::TryInto;
 use std::fmt;
+use std::io::SeekFrom;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
@@ -72,7 +74,12 @@ impl GcsState {
         Ok(token)
     }
 
-    async fn fetch<W>(&self, key: &SharedCacheKey, writer: &mut W) -> Result<Option<()>>
+    /// Fetches item from shared cache if available and copies them to the writer.
+    ///
+    /// # Returns
+    ///
+    /// If successful the number of bytes written to the writer are returned.
+    async fn fetch<W>(&self, key: &SharedCacheKey, writer: &mut W) -> Result<Option<u64>>
     where
         W: tokio::io::AsyncWrite + Unpin,
     {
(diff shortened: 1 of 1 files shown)


crates/symbolicator/src/services/shared_cache.rs
```diff
@@ -135,6 +141,7 @@ impl GcsState {
     }
 
     async fn store(&self, key: SharedCacheKey, mut src: File) -> Result<SharedCacheStoreResult> {
+        let total_bytes = src.seek(SeekFrom::End(0)).await?;
```
Swatinem, line 144: you have a `rewind()` right below which already existed before. Given that, what do we expect the current seek position is when calling this function? Is it at the beginning? (which means the previously `rewind()` was useless) Or is it undefined (which means we need to add another explicit `rewind()` here.

crates/symbolicator/src/services/shared_cache.rs
```diff
@@ -473,17 +492,22 @@ impl SharedCacheService {
...
+            Ok(Some(bytes)) => {
                 metric!(
-                    counter("services.shared_cache.store") += 1,
+                    counter("services.shared_cache.fetch") += 1,
                     "cache" => key.name.as_ref(),
                     "hit" => "true",
                 );
+                let bytes: i64 = bytes.try_into().unwrap_or(i64::MAX);
+                metric!(
+                    counter("services.shared_cache.fetch.bytes") += bytes,
```
Swatinem, line 503: don’t we use a different metric type for these histograms?

Swatinem (approved):

crates/symbolicator/src/services/shared_cache.rs
flub: I was treating the position before calling this as undefined, which i think is the more defensive approach.

I don't think we need a second rewind: here we explicitly go to the end to find out the size.  then explicitly go to the beginning so we can use `io::copy()`.

Swatinem: Oh, I maybe misunderstood the function signature. The function will `then yield the new position in the object`, not sure why I was thinking about it returning the number of bytes that were seeked over. So I guess this was just my misunderstanding.

crates/symbolicator/src/services/shared_cache.rs
flub: (turns out i had a half-written response here... i'm good at this)

The aim here is not to get an idea of the object sizes uploaded with percentiles of their sizes etc, object sizes already have a metric.  I'm only after the rate, bytes/second, of uploading and downloading to the shared cache so we know how much bandwidth we're spending on this.  This is to verify that we're not using too much of the available machine's bandwidth and that the throttling we have in place is sufficient to not disrupt symbolicator's normal work.

PR #590, opened 2021-11-30
fix(gcs): Only request read-only access

flub:
We accidentally switched to requesting a read-write scope for all
buckets symbolicator tries to access and customers might not have
given our tokens those permissions, resulting in us not being able to
use their symbol source.

Tests probably keep passing since the GCP tests are not configured to
run correctly in CI yet.  Something else to fix, but let's stop this
regression first.

#skip-changelog


relaxolotl:
is the shared cache disabled on prod right now?

relaxolotl:
pinged @flub on slack and he has confirmed that the shared cache is indeed disabled

PR #591, opened 2021-11-30
feat(shared-cache): Switch to gcp_auth crate

flub:
This handles service accounts while running in GCP much better than
our statically configured tokens would do otherwise.

NATIVE-248


crates/symbolicator/src/services/shared_cache.rs
```diff
@@ -88,7 +72,7 @@ impl GcsState {
         let request = self
             .client
             .get(url)
-            .header("authorization", token.bearer_token())
+            .header("authorization", format!("Bearer {}", token.as_str()))
```
Swatinem, line 75: There is btw https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.bearer_auth
It does its own `format` internally though.

crates/symbolicator/src/services/shared_cache.rs
```diff
@@ -33,13 +33,22 @@ use super::cacher::CacheKey;
...
     client: Client,
     auth_manager: gcp_auth::AuthenticationManager,
 }
 
+impl fmt::Debug for GcsState {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("GcsState")
+            .field("config", &self.config)
+            .field("client", &self.client)
+            .field("auth_manager", &"<AuthenticationManager")
```
Swatinem, line 47: this looks weird, is that intentional?

Swatinem (approved):
related to the "what color is your function" problem, some of the constructors are now async, and return `Result`, which are being unwrapped right now. I’m okay with that, since all the cases I see are in the startup path. But having `unwrap` during runtime is absolutely not something we want to have ;-)

flub:
> related to the "what color is your function" problem, some of the constructors are now async, and return `Result`, which are being unwrapped right now.

huh?  we're not unwrapping any of that in the symbolicator code, only in the tests.  did i miss something?

> But having `unwrap` during runtime is absolutely not something we want to have ;-)

totally agree here!

Swatinem:
Oh, maybe I had an outdated commit in mind. Double-checking, you changed all of the constructors to `try_new`, which is great!

PR #592, opened 2021-12-01
ref: Update dependencies (NATIVE-397)

Swatinem:
This needed some minor changes because of the new `axum` version, otherwise it was just
a normal `cargo update`.

https://getsentry.atlassian.net/browse/NATIVE-397

#skip-changelog

### DIFF
--- a/crates/symbolicator/Cargo.toml
+++ b/crates/symbolicator/Cargo.toml
@@ -6,14 +6,14 @@ edition = "2018"
 license = "MIT"
 
 [dependencies]
-axum = { version = "0.2.1", features = ["multipart"] }
+axum = { version = "0.3.4", features = ["multipart"] }
 anyhow = "1.0.38"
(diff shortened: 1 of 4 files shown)


crates/symbolicator/Cargo.toml
```diff
@@ -27,7 +27,7 @@ lazy_static = "1.4.0"
 log = { version = "0.4.13", features = ["serde"] }
 lru = "0.7.0"
 num_cpus = "1.13.0"
-minidump = "0.3"
+minidump = "0.9.4"
```
flub, line 30: this is a big jump and important change.  might want to watch this

flub (approved):

crates/symbolicator/Cargo.toml
Swatinem: The version jump was purely cosmetic I think, as it went straight from 0.3 to 0.9.

PR #594, opened 2021-12-02
ref(s3): Improve error reporting a little

flub:
This should also improve our sentry error grouping for these.

#skip-changelog


Swatinem (approved):
omg thats horrible, lol

but if it gets the job done 🤷🏻‍♂️ are you sure we can’t do that with custom grouping rules?

PR #600, opened 2021-12-03
fix: Fix lazy computations persisting in `current_computations`

Swatinem:
I think there was a bug where the lazy computation request was being written into `current_computations` and never cleared from there.
Apart from being a memory leak, that could have potentially returned the error to a non-lazy computation requests based on the `key`.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/cacher.rs
+++ b/crates/symbolicator/src/services/cacher.rs
@@ -409,23 +409,8 @@ impl<T: CacheItemRequest> Cacher<T> {
         F: std::future::Future<Output = Result<T::Item, T::Error>> + Send + 'static,
     {
         let (sender, receiver) = oneshot::channel();
+
         let max_lazy_refreshes = self.config.max_lazy_refreshes();
-
-        // we count down towards zero, and if we reach or surpass it, we will short circuit here.
-        if is_refresh && max_lazy_refreshes.fetch_sub(1, Ordering::Relaxed) <= 0 {
-            max_lazy_refreshes.fetch_add(1, Ordering::Relaxed);
-
-            let result = Err(Arc::new(
-                Error::new(
-                    ErrorKind::Other,
-                    "maximum number of lazy recomputations reached; aborting cache computation",
-                )
-                .into(),
-            ));
-            sender.send(result).ok();
-            return receiver.shared();
-        }
-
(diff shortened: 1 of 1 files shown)


loewenheim (approved):

crates/symbolicator/src/services/cacher.rs
```diff
@@ -477,6 +462,29 @@ impl<T: CacheItemRequest> Cacher<T> {
             } else {
                 // A concurrent cache lookup is considered new. This does not imply a cache miss.
                 metric!(counter(&format!("caches.{}.channel.miss", name)) += 1);
+
+                // We count down towards zero, and if we reach or surpass it, we will short circuit here.
+                // Doing the short-circuiting here means we don't create a channel at all, and don't
+                // put it into `current_computations`.
+                // Previously, that could have potentially returned the error to a non-lazy
+                // computation requests based on the `key`.
```
flub, line 470: not much point leaving something like this in a comment.  better to write it in the commit message

crates/symbolicator/src/services/cacher.rs
```diff
@@ -477,6 +462,29 @@ impl<T: CacheItemRequest> Cacher<T> {
...
+                let max_lazy_refreshes = self.config.max_lazy_refreshes();
+                if is_refresh && max_lazy_refreshes.fetch_sub(1, Ordering::Relaxed) <= 0 {
+                    max_lazy_refreshes.fetch_add(1, Ordering::Relaxed);
+
+                    metric!(counter(&format!("caches.{}.lazy_limit_hit", name)) += 1);
```
flub, line 475: can we put the cache name in a tag instead of interpolating the metric name?  I know there's other metrics that do this, but let's not do this for new metrics at least.

flub (approved):
the logic of this is getting ever more complicated and intermingled.  but i think this correct after all.  still would like to improve this all sometime though.

crates/symbolicator/src/services/cacher.rs
Swatinem: fair enough, I think changing all the others might involve more work also in DD, etc.

flub: yes, no need to change existing metric names

PR #607, opened 2021-12-09
fix(shared-cache): Ensure startup with broken shared cache

flub:
If the shared cache is not working for some reason we still want to
ensure symbolicator starts up so that it does not affect the normal
operation.

NATIVE-424

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/mod.rs
+++ b/crates/symbolicator/src/services/mod.rs
@@ -77,8 +77,7 @@ impl Service {
             .context("failed to create process pool")?;
 
         let downloader = DownloadService::new(config.clone());
-        let shared_cache =
-            Arc::new(SharedCacheService::try_new(config.shared_cache.clone()).await?);
+        let shared_cache = Arc::new(SharedCacheService::new(config.shared_cache.clone()).await);
         let caches = Caches::from_config(&config).context("failed to create local caches")?;
         caches
             .clear_tmp(&config)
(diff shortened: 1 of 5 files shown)


crates/symbolicator/src/services/mod.rs
```diff
@@ -77,8 +77,7 @@ impl Service {
             .context("failed to create process pool")?;
 
         let downloader = DownloadService::new(config.clone());
-        let shared_cache =
-            Arc::new(SharedCacheService::try_new(config.shared_cache.clone()).await?);
+        let shared_cache = Arc::new(SharedCacheService::new(config.shared_cache.clone()).await);
         let caches = Caches::from_config(&config).context("failed to create local caches")?;
```
Swatinem, line 81: This code has been here for a far longer time, I wonder if we should have a followup discussion about "hard fail on mis-configuration", as this was such a case essentially?

Swatinem (approved):

loewenheim (approved):

crates/symbolicator/src/services/mod.rs
flub: question is whether you can still do meaningful work when misconfigured.  if you have no working cache... while in theory you can still fail in practice my guess is symbolicator in prod is better off crashlooping compared to trying to work without cache.

but generally reviewing cases could be an interesting exercise

PR #613, opened 2021-12-14
build: Switch back to released gcp_auth version

flub:
The required fixes and features have been merged and released.

#skip-changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -951,8 +951,9 @@ dependencies = [
 
 [[package]]
 name = "gcp_auth"
-version = "0.6.0"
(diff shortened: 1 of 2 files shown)


loewenheim (approved):

relaxolotl (approved):

relaxolotl:
out of curiosity, what'll happen to the fork that this was originally pointing to? will we be taking that down, or?

flub:
That's a good question, i was planning on removing it to avoid confusion.  OTOH it means if you check out an old version you might not be able to build :/

PR #619, opened 2021-12-22
feat(wasm-split): --external-dwarf-url option

bkotsopoulossc:
**Background**

Closes https://github.com/getsentry/symbolicator/issues/614

Note that I brought in a new library to implement the variable-int encoding of the size prefix for the custom section payload. I noticed that the wasmbin library already has an implementation of this [here](https://github.com/GoogleChromeLabs/wasmbin/blob/77f8d736219b5129188c7a6a355c74151bfdc67d/src/builtins/integers.rs#L89). I've filed a ticket on that repo [here](https://github.com/GoogleChromeLabs/wasmbin/issues/6) asking for help.

In the meantime though, this implementation seems to work from my testing. I'm new to rust so please let me know what can improve.

PTAL @Swatinem 

**Testing**

Using a sample wasm binary with debug symbols
```
% wasm-objdump -h main.wasm          

main.wasm:      file format wasm 0x1

Sections:

...
   Custom start=0x009579fb end=0x0af8e3f3 (size=0x0a6369f8) "name"
   Custom start=0x0af8e3f8 end=0x0f78df7a (size=0x047ffb82) ".debug_info"
   Custom start=0x0f78df7f end=0x0fff1866 (size=0x008638e7) ".debug_loc"
   Custom start=0x0fff186b end=0x1034d7e9 (size=0x0035bf7e) ".debug_ranges"
   Custom start=0x1034d7ed end=0x10502f61 (size=0x001b5774) ".debug_abbrev"
   Custom start=0x10502f66 end=0x1202e813 (size=0x01b2b8ad) ".debug_line"
   Custom start=0x1202e818 end=0x18701a0b (size=0x066d31f3) ".debug_str"
```

Without a debug file, and no option specified, no change:
```
% ./target/release/wasm-split --strip-names --strip main.wasm -o stripped.wasm
e966d2ec508f45038a05c396aecc2e29
% wasm-objdump -j external_debug_info -s stripped.wasm   
stripped.wasm:  file format wasm 0x1
Section not found: external_debug_info
```

Without a debug file, option specified, url is present:
```
% ./target/release/wasm-split --strip-names --strip main.wasm -o stripped.wasm --external-dwarf-url https://my-site.com/foo.debug.wasm
649fad5124584411912e58c9ee8d7dd6
% wasm-objdump -j external_debug_info -s stripped.wasm

stripped.wasm:  file format wasm 0x1

Contents of section Custom:
[...]

PR #623, opened 2022-01-07
fix incorrect changelog entry from #619

bkotsopoulossc:
Explanation here https://github.com/getsentry/symbolicator/pull/619/files#r779745210.

PTAL @Swatinem and if possible can you cut a 0.4.2 release afterwards?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+### Features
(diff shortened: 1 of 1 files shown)


mitsuhiko (approved):

bkotsopoulossc:
thanks for the quick review, we've found that the unit tests fail because a PR from a fork doesn't have access to the secrets in CI to access GCS. And the danger changelog check seems to be failing for a similar reason - it can't access the PR

PR #628, opened 2022-01-10
feat: Hook up the new sentry-tower Http Layer (NATIVE-327)

Swatinem:
This hooks up the new sentry-tower Http layer to capture request properties, and also start a new transaction.

It also pulls in `tracing` and sprinkles a few `instrument` attrs around, and does explicit trace/transaction forwarding in a couple of places. (so this supercedes #515)

The result is this:

Bildschirmfoto 2022-01-13 um 15 18 14

#skip-changelog ffs

### DIFF
--- a/crates/symbolicator/Cargo.toml
+++ b/crates/symbolicator/Cargo.toml
@@ -37,7 +37,8 @@ reqwest = { git = "https://github.com/jan-auer/reqwest", tag = "v0.11.0", featur
 rusoto_core = "0.47.0"
 rusoto_credential = "0.47.0"
 rusoto_s3 = "0.47.0"
-sentry = { git = "https://github.com/getsentry/sentry-rust", version = "0.23.0", features = ["anyhow", "debug-images", "log", "tower"] }
+sentry = { git = "https://github.com/getsentry/sentry-rust", version = "0.23.0", features = ["anyhow", "debug-images", "log", "tracing"] }
+sentry-tower = { git = "https://github.com/getsentry/sentry-rust", version = "0.23.0", features = ["http"] }
 serde = { version = "1.0.119", features = ["derive", "rc"] }
 serde_json = "1.0.61"
 serde_yaml = "0.8.15"
@@ -47,6 +48,8 @@ symbolic = { git = "https://github.com/getsentry/symbolic", branch = "fix/demang
 # symbolic = { path = "../../../symbolic/symbolic", features = ["common-serde", "debuginfo", "demangle", "minidump-serde", "symcache"] }
 tempfile = "3.2.0"
 thiserror = "1.0.26"
+tracing = "0.1.29"
+tracing-subscriber = { version = "0.3.5", default-features = false, features = ["std", "registry"] }
 tokio = { version = "1.0.2", features = ["rt", "macros", "fs"] }
 tokio-util = "0.6"
 tower = "0.4"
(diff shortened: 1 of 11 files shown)


Swatinem:
I filed https://github.com/tokio-rs/tracing/issues/1831 about the bogus lint issue.

Swatinem:
Lets do this. We are currently using a git version of `sentry-rust` anyway, this just bumps the version up. Releasing is currently blocked on https://github.com/getsentry/publish/issues/740, but that shouldn’t block this here.

crates/symbolicator/Cargo.toml
```diff
@@ -37,7 +37,8 @@ reqwest = { git = "https://github.com/jan-auer/reqwest", tag = "v0.11.0", featur
 rusoto_core = "0.47.0"
 rusoto_credential = "0.47.0"
 rusoto_s3 = "0.47.0"
-sentry = { git = "https://github.com/getsentry/sentry-rust", version = "0.23.0", features = ["anyhow", "debug-images", "log", "tower"] }
+sentry = { git = "https://github.com/getsentry/sentry-rust", version = "0.23.0", features = ["anyhow", "debug-images", "log", "tracing"] }
+sentry-tower = { git = "https://github.com/getsentry/sentry-rust", version = "0.23.0", features = ["http"] }
```
flub, line 41: does it really make sense to still have `version = ` in there if there's a `git = `?

crates/symbolicator/src/services/cacher.rs
```diff
@@ -421,13 +421,31 @@ impl<T: CacheItemRequest> Cacher<T> {
...
+            // otherwise it will not "outlive" its parent span, so attach it to the parent transaction.
+            let transaction = if is_refresh {
+                let span = sentry::configure_scope(|scope| scope.get_span());
+                let ctx = sentry::TransactionContext::continue_from_span(
+                    "Lazy Cache Computation",
+                    "spawn_computation",
+                    span,
+                );
+                let transaction = sentry::start_transaction(ctx);
+                sentry::configure_scope(|scope| scope.set_span(Some(transaction.clone().into())));
```
flub, line 434: not for here, but what's the logic for `.set_span` to take an `Option` rather than the span/transaction directly?  Seems weird

crates/symbolicator/src/services/cficaches.rs
```diff
@@ -137,6 +137,11 @@ struct FetchCfiCacheInternal {
     threadpool: tokio::runtime::Handle,
 }
 
+/// Extracts the Call Frame Information (CFI) from an object file.
+///
+/// The extracted CFI is written to `path` in symbolic's
+/// [`CfiCache`](symbolic::minidump::cfi::CfiCache) format.
```
flub, line 143: ❤️

flub (approved):

crates/symbolicator/src/services/cacher.rs
Swatinem: Well, you are supposed to unset the "active" span somehow. The unified API unfortunately does not really make this very clear, meh.

crates/symbolicator/Cargo.toml
Swatinem: not necessarily, but that way you can just remove the `git` ref and you are left with the version ;-)

crates/symbolicator/src/services/cacher.rs
flub: I we have some creative license wrt to the unified API, for rust I expect more `fn set_span(&self, span: impl TracingSpan)` and `fn detach_span(&self)` or something

PR #631, opened 2022-01-13
ref: Forward CacheItemRequest::compute to async fns

Swatinem:
This refactors the code, pulling as much of the actual computation out into an async fn.

Moves some refactors out of #628 and this should in the end help with https://github.com/tokio-rs/tracing/issues/1831

#skip-changelog ffs

### DIFF
--- a/crates/symbolicator/src/services/objects/meta_cache.rs
+++ b/crates/symbolicator/src/services/objects/meta_cache.rs
@@ -8,7 +8,7 @@
 //! consistency.
 
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
 use anyhow::Context;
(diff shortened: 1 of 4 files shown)


crates/symbolicator/src/services/objects/meta_cache.rs
```diff
@@ -103,47 +96,57 @@ impl CacheItemRequest for FetchFileMetaRequest {
     /// This returns [`CacheStatus::CacheSpecificError`] if the download failed.  If the
     /// data cache is [`CacheStatus::Negative`] or [`CacheStatus::Malformed`] then the same
     /// status is returned.
-    fn compute(&self, path: &Path) -> BoxFuture<'static, Result<CacheStatus, Self::Error>> {
+    ///
+    /// This is the actual implementation of [`CacheItemRequest::compute`] for
+    /// [`FetchFileMetaRequest`] but outside of the trait so it can be written as async/await
+    /// code.
+    async fn compute_file_meta(self, path: PathBuf) -> Result<CacheStatus, ObjectError> {
```
flub, line 103: this version is much nicer than some of the others because of the `self` arg which means it doesn't need a really long arg list.  I guess doing this for the others is too much refactoring (probably not all `self`s are clonable)?

PR #645, opened 2022-01-25
wip: Draft for replacing breakpad with rust-minidump

jan-auer:
This is a PoC for how to integrate rust-minidump into Symbolicator. It doesn't allow for side-by-side comparison, so it should not be merged as-is.


crates/symbolicator/src/services/symbolication.rs
```diff
@@ -2032,6 +2126,7 @@ impl SymbolicationActor {
                 break result;
             }
 
+            // TODO(ja): Doesn't check that this actually resolves modules
```
jan-auer, line 2129: I think this is a misconception, please ignore.

PR #662, opened 2022-02-08
ref: Try to optimize stackwalking procspawn usage

Swatinem:
This does a couple of things:
- Switches from `BTreeMap` back to `Vec`, keeping the types that are being serialized over procspawn simpler.
  - This is: all modules, missing modules, and cfi caches
- Converting modules to `HashMap`, as lookups should ideally be `O(1)` vs `O(log n)`, though not sure that matters too much.
- Avoid serializing/sending Object Infos for missing modules, the mapping happens *after* procspawn now.
- Only pass the complete module list on the first run, since that should be idempotent.
- Comparison mode will only compare `stacktraces`, as Vec/HashMap have unpredictable sorting order, and the previous optimization makes sure we only get the complete module list once anyway.

The effect is probably hard to quantify. It went from `Requests/sec:    189.40` to `Requests/sec:    216.52`, though I have a bunch of other stuff running in the background now so numbers are not quite stable anyway :-D

Another data point: profiling with Instruments showed 20% time spent in ipc-channel before vs 6% now. So a lot better than before. And an interesting data point in itself, as this quantifies the potential savings by cutting out procspawn completely.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/symbolication.rs
+++ b/crates/symbolicator/src/services/symbolication.rs
@@ -219,7 +219,7 @@ impl CfiCacheModules {
     }
 
     /// Returns a mapping of module IDs to paths that can then be loaded inside a procspawn closure.
-    fn for_processing(&self) -> BTreeMap<DebugId, PathBuf> {
+    fn for_processing(&self) -> Vec<(DebugId, PathBuf)> {
         self.inner
             .iter()
             .filter_map(|(id, module)| Some((*id, module.cfi_path.clone()?)))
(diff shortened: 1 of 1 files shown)


wrk_minidump.lua
```diff
@@ -0,0 +1,29 @@
+-- This is a `wrk` script that posts a multipart minidump request to symbolicator.
```
flub, line 1: could we not have this in the root of the project somehow?  maybe a "scripts" or "contrib" directory?

flub (approved):

PR #663, opened 2022-02-08
ref: Optimize Symbolication a bit

Swatinem:
* Avoids some cloning, in particular the whole module list
* Avoids an intermediate `Vec` allocation
* Avoids creating a bunch of short-circuiting Futures, including all the Arc-ing they require

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/symbolication.rs
+++ b/crates/symbolicator/src/services/symbolication.rs
@@ -1,4 +1,4 @@
-use std::collections::{BTreeMap, BTreeSet, HashMap};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::convert::TryInto;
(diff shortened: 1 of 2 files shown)


loewenheim (approved):

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -824,22 +826,26 @@ impl SymCacheLookup {
...
-            sources,
-            scope,
-            ..
-        } = request;
-
+        scope: Scope,
+        sources: Arc<[SourceConfig]>,
+        stacktraces: &[RawStacktrace],
+    ) {
+        let mut referenced_objects = HashSet::new();
```
flub, line 846: same here, i think this is storing an index?  i assume the same logic applies but i'm curious what it is

PR #673, opened 2022-02-11
ref: Target upstream rust-minidump

flub:
This adapts to the async API of rust-minidump but also does a few other things:

* Changes the FrameTrust mapping to our own struct so we can impl Default etc we need for it.
* Completely ignores registers for now, this is a regression

Apart from the obvious there are still a few test failures that I haven't gotten to the bottom off.

#skip-changelog

### DIFF
--- a/crates/symbolicator/Cargo.toml
+++ b/crates/symbolicator/Cargo.toml
@@ -43,7 +43,7 @@ serde_json = "1.0.61"
 serde_yaml = "0.8.15"
 similar = "2.1.0"
 structopt = "0.3.21"
-symbolic = { git = "https://github.com/getsentry/symbolic", branch = "fix/demangle-fixes", version = "8.5.0", features = ["common-serde", "debuginfo", "demangle", "minidump-serde", "symcache"] }
+symbolic = { git = "https://github.com/getsentry/symbolic", branch = "fix/demangle-fixes", features = ["common-serde", "debuginfo", "demangle", "minidump-serde", "symcache"] }
 tempfile = "3.2.0"
 thiserror = "1.0.26"
 tokio = { version = "1.0.2", features = ["rt", "macros", "fs"] }
--- a/crates/symbolicator/Cargo.toml
+++ b/crates/symbolicator/Cargo.toml
@@ -6,9 +6,10 @@ edition = "2021"
 license = "MIT"
 
 [dependencies]
-axum = { version = "0.4.3", features = ["multipart"] }
 anyhow = "1.0.38"
 apple-crash-report-parser = "0.4.2"
+async-trait = "0.1.52"
+axum = { version = "0.4.3", features = ["multipart"] }
 backtrace = "0.3.55"
 base64 = "0.13.0"
 cadence = "0.27.0"
@@ -25,9 +26,9 @@ ipnetwork = "0.18.0"
 jsonwebtoken = "7.2.0"
 lazy_static = "1.4.0"
 lru = "0.7.0"
+minidump = { git = "https://github.com/luser/rust-minidump", branch = "master" }
+minidump-processor = { git = "https://github.com/luser/rust-minidump", branch = "master", features = ["symbolic-syms"] }
 num_cpus = "1.13.0"
-minidump = { git = "https://github.com/getsentry/rust-minidump", branch = "feat/symbolicator-integration" }
-minidump-processor = { git = "https://github.com/getsentry/rust-minidump", branch = "feat/symbolicator-integration" }
 parking_lot = "0.11.1"
 procspawn = { version = "0.10.0", features = ["backtrace", "json"] }
 regex = "1.4.3"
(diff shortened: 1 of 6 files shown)


crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1500,18 +1501,18 @@ impl TempSymbolProvider {
...
-        module: &dyn Module,
-        _frame: &mut dyn minidump_processor::FrameSymbolizer,
-    ) -> Result<(), minidump_processor::FillSymbolError> {
-        // TODO(ja): Deduplicate this. Probably should use a different map key, ...
+        module: &(dyn Module + Sync),
+        _frame: &mut (dyn FrameSymbolizer + Send),
+    ) -> Result<(), FillSymbolError> {
         let debug_id = module
             .debug_identifier()
             .and_then(|id| DebugId::from_str(&id).ok())
```
Swatinem, line 1518: just a NOTE, and maybe an obsolete one given that we are changing these IDs upstream:
I have found through profiling that debug-id parsing is surprisingly expensive, so avoiding that might be a good idea.

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1938,13 +1941,20 @@ impl SymbolicationActor {
...
+                        rt.block_on(async move {
+                            stackwalk_with_rust_minidump(
+                                cfi_caches,
+                                minidump_path,
+                                spawn_time,
+                                return_modules,
+                            )
+                            .await
+                            .map(procspawn::serde::Json)
+                        })
```
Swatinem, line 1957:
```suggestion
                        procspawn::serde::Json(rt.block_on(
                            stackwalk_with_rust_minidump(
                                cfi_caches,
                                minidump_path,
                                spawn_time,
                                return_modules,
                            )
                        ))
```
(+ reformatting ;-)

Might be a bit simpler (and avoid un unneeded async move block)

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1500,18 +1501,18 @@ impl TempSymbolProvider {
             .ok()
     }
 
-    fn into_missing_ids(self) -> BTreeSet<DebugId> {
-        self.missing_ids.into_inner()
+    fn missing_ids(&self) -> Vec<DebugId> {
+        self.missing_ids.read().iter().copied().collect()
```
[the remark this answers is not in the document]
flub: TIL RwLock has `.into_inner()`

crates/symbolicator/Cargo.toml
```diff
@@ -40,28 +41,26 @@ sentry-tower = { version = "0.24.2", features = ["http"] }
 serde = { version = "1.0.119", features = ["derive", "rc"] }
 serde_json = "1.0.61"
 serde_yaml = "0.8.15"
+similar = "2.1.0"
 structopt = "0.3.21"
 symbolic = { git = "https://github.com/getsentry/symbolic", branch = "fix/demangle-fixes", version = "8.5.0", features = ["common-serde", "debuginfo", "demangle", "minidump-serde", "symcache"] }
-# Uncomment the line below for local development
-# symbolic = { path = "../../../symbolic/symbolic", features = ["common-serde", "debuginfo", "demangle", "minidump-serde", "symcache"] }
```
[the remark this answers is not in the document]
flub: shouldn't this be done in a `patch.crates-io` section?  Having that here makes sorting a lot harder :)

https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#working-with-an-unpublished-minor-version

loewenheim (approved):

Swatinem (approved):

crates/symbolicator/src/services/snapshots/symbolicator__services__symbolication__tests__minidump_linux_rust_minidump.snap
```diff
@@ -1,15 +1,15 @@
...
 
 ---
 status: completed
 timestamp: 1522061032
 system_info:
   os_name: Linux
-  os_version: 0.0.0
-  os_build: "Linux 4.9.60-linuxkit-aufs #1 SMP Mon Nov 6 16:00:12 UTC 2017 x86_64"
+  os_version: 4.9.60-linuxkit-aufs
+  os_build: "#1 SMP Mon Nov 6 16:00:12 UTC 2017"
```
relaxolotl, line 12: thanks for fixing this up!

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -571,17 +571,13 @@ fn object_info_from_minidump_module_rust_minidump(
...
     module: &MinidumpModule,
 ) -> RawObjectInfo {
-    let code_id = module.code_identifier().to_string();
+    let code_id = module.code_identifier();
 
     RawObjectInfo {
         ty,
-        // TODO: shouldn't this be None if code_id is empty?
-        code_id: Some(code_id),
+        code_id: Some(code_id.to_string()),
```
relaxolotl, line 578: nit: it doesn't look like `code_id` gets reused anywhere, so this could be a one-liner like the other fields,

`code_id: Some(code_id.code_identifier().to_string()`

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -571,17 +571,13 @@ fn object_info_from_minidump_module_rust_minidump(
     ty: ObjectType,
     module: &MinidumpModule,
 ) -> RawObjectInfo {
-    let code_id = module.code_identifier().to_string();
+    let code_id = module.code_identifier();
 
     RawObjectInfo {
         ty,
-        // TODO: shouldn't this be None if code_id is empty?
```
relaxolotl, line 578: this is for my own personal curiosity, but what was the answer to this question?

crates/symbolicator/src/types/mod.rs
```diff
@@ -236,48 +235,38 @@ pub struct RawFrame {
...
-    use super::*;
-    use serde::{Deserializer, Serializer};
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum FrameTrust {
+    None,
+    Scan,
+    CfiScan,
+    Fp,
+    Cfi,
```
relaxolotl, line 249: Nitpicky, but I think the original names (`FramePointer`, `CallFrameInfo`) are better than their new ones here, despite their verbosity. 

Figuring out what these expand to on your own via the codebase is already difficult enough: there're only two places that mention what CFI expands out to, and one place that does the same for FP. The `From` implementation below serves as one of those spots because `rust-minidump` uses the full name. Looking these up via a search engine does not provide useful results, even if you add "frametrust" to the search query. This additional lookup to see what the acronym expands out to can be easily resolved by just using the original values. 

Using the acronyms in these variants makes it less unwelcoming for newcomers, and creates another place in the code where the easiest way to find an answer to one's question is to speak to a member of the project which we should really cut down on.

crates/symbolicator/src/services/symbolication.rs
```diff
@@ -1495,48 +1491,48 @@ impl TempSymbolProvider {
...
         // minidump processing and the types, however.
         // TODO(ja): Check if this is OK. Shouldn't trigger skip heuristics
-        if self.files.contains_key(&debug_id) {
-            Ok(())
-        } else {
-            Err(minidump_processor::FillSymbolError {})
+        match module
+            .debug_identifier()
+            .map(|debug_id| self.files.contains_key(&debug_id))
+            .unwrap_or_default()
```
relaxolotl, line 1513: Mixed feelings seeing this change because I think it makes it more difficult to see the main operation being acted upon here (`self.files.contains_key`), but I don't think it's a dealbreaker. Looks like logically this is the same, so this is more of a readability nit than anything else.

relaxolotl (approved):
Thanks for including the changes coming from my recent PR in rust-minidump! Left some nitpicky thoughts but overall I don't think any of them should block this, which removes one of the big barriers when it comes to testing rust-minidump changes.

crates/symbolicator/src/services/symbolication.rs
flub: code_id can't be empty because of the new API which guarantees to give you one.  But, hum, good question.  Maybe this isn't true because you can still create those to be empty.  I'll do this as a followup up though because this could be messy.

Issue #654, opened 2022-02-03
Symstore proxy does not actually work


vaartis:
### Environment

Symbolicator built from github tags.

### Steps to Reproduce

1. Configured an S3 bucket source and tried to open the url `http://localhost:3021/proxy/(name of the pdb)/(hash of the pdb)/(the same name of the pdb)`. 
2. The URL only matches when there is a single component after `/proxy/`, otherwise it does not. This is probably because [here](https://github.com/getsentry/symbolicator/blob/master/crates/symbolicator/src/endpoints/mod.rs#L38) there is `:path` at the end, which _seems_ to only match when there is a single URL component afterwards and fail otherwise. [This](https://github.com/getsentry/symbolicator/blob/93caba50446ffaf1965f3532560207bf3d9bf7f7/crates/symbolicator/src/utils/paths.rs#L444-L448), however, implies that there should be at least 3 components in the URL after `/symbols`, which means this is probably a bug.
3.  This seems to affect all releases after 0.3.4 (which actually has this working, but the URL is `/symbols/` instead of `/proxy` and it uses different routing which is not affected by this problem). In 0.3.4 everything works as expected.
 
### Expected Result

`symbolicator` should have proxied the request and sent the file.

### Actual Result

Symbolicator sends a 404 and writes nothing at all into the logs. Perhaps there should be at least some indication in the logs if the URL doesn't match anything.

jan-auer:
We'll have to update the path matcher to [use a wildcard](https://docs.rs/axum/latest/axum/struct.Router.html#wildcards).

vaartis:
Still doesn't work, and at this point since I can actually test it, I'll fix and test it myself, and send a PR.

akimyonoglu:
for anyone scratching their head to get proxy working and made it to here. I was able to get a download started using following url pattern.

```<SYMBOLICATOR_IP>:3021/proxy/symbols/<DEBUG FILENAME>/<HASH>/<DEBUG FILENAME>```

it was a wonderful exploration, documentation says `/symbols`, code looks like to be expecting `/proxy` but actual was `/proxy/symbols` :)

PR #675, opened 2022-02-12
fix: Skip the leading slash for the symstore path

vaartis:
For #654. Now it does actually work, at least from my testing on an actual S3 bucket.
Would be nice to have a new release published after this, so people who use tagged versions can actually benefit from this.

#skip-changelog


flub (approved):

flub:
Test failures are due to the GCS/GCP secrets not being available to forks.  Changelog entry already exists for that issue, so fine to override the checks afaik.

PR #687, opened 2022-02-22
ref: Give http transactions better names

Swatinem:
This solves the problem of having a ton of unique `GET /requests/UUID` performance monitoring transactions.

Also updates the Sentry SDK for newly introduced API.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/cficaches.rs
+++ b/crates/symbolicator/src/services/cficaches.rs
@@ -299,7 +299,6 @@ impl CfiCacheActor {
 #[tracing::instrument(skip_all)]
 fn write_cficache(path: &Path, object_handle: &ObjectHandle) -> Result<(), CfiCacheError> {
     configure_scope(|scope| {
-        scope.set_transaction(Some("compute_cficache"));
         object_handle.to_scope(scope);
     });
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2726,9 +2726,9 @@ checksum = "568a8e6258aa33c13358f81fd834adb854c6f7c9468520910a9b1e8fac068012"
 
 [[package]]
 name = "sentry"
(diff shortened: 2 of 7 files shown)


jan-auer:
@Swatinem where was the original transaction name assigned, is that done by an integration?

Swatinem:
Yes, the tower-http layer does that. It only has access to the URL, not the matcher itself.

jan-auer:
> It only has access to the URL, not the matcher itself.

Thanks, that's interesting. We're seeing similar issues in our React integration, so maybe this will require a broader solution after all in the long-term. cc @jjbayer 

@Swatinem In the interim, it would be valuable to document this workaround. Effectively everyone using the tower integration has to check and work around this to get proper transactions.

crates/symbolicator/src/services/cficaches.rs
```diff
@@ -299,7 +299,6 @@ impl CfiCacheActor {
 #[tracing::instrument(skip_all)]
 fn write_cficache(path: &Path, object_handle: &ObjectHandle) -> Result<(), CfiCacheError> {
     configure_scope(|scope| {
-        scope.set_transaction(Some("compute_cficache"));
```
flub, line 302: in theory it seems to me this should have been some kind of nested transaction inside the main request one?  does the product handle nested transactions?

flub (approved):

crates/symbolicator/src/services/cficaches.rs
flub: so i'm curious then why you decided to remove these transactions?

Swatinem: because they would overwrite the `symbolicate_stacktraces` name.

PR #697, opened 2022-02-25
ref: Pacify clippy

Swatinem:
#skip-changelog


crates/symbolicator/src/services/symbolication.rs
```diff
@@ -790,7 +790,7 @@ fn is_likely_base_frame(frame: &SymbolicatedFrame) -> bool {
         .raw
         .function
         .as_deref()
-        .or_else(|| frame.raw.symbol.as_deref())
+        .or(frame.raw.symbol.as_deref())
```
flub, line 793: oh, this finally is stable?  this was super annoying as there was no syntax that beta and stable clippy where both happy with 😠

flub (approved):

PR #703, opened 2022-03-02
ref: Metrics around symbolication task spawning

flub:
This experiments with metrics targetted around the symbolication task
spawning.  This matters because these spawning times starting to
increase is a precursor to things going south and our backpressure not
working correctly.  Until recently we had our healthchecks failing in
these cases but now they should be more isolated and not provide that
incredibly crude form of backpressure (kill symbolicator).

It does a manual histogram-timer first spawn duration as we have been
using in various places before.  It also experiments with adding
tokio-metrics to measure the same thing in more detail.  The reason to
do both is to be able to compare and evaluate tokio-metrics a bit.

#skip-changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3264,6 +3264,7 @@ dependencies = [
  "test-assembler",
  "thiserror",
  "tokio",
+ "tokio-metrics",
  "tokio-util",
  "tower",
  "tower-layer",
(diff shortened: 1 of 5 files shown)


Swatinem (approved):
didn’t we have "future poll"/"task spawn" metrics already somewhere?

flub:
> didn’t we have "future poll"/"task spawn" metrics already somewhere?

Good point, we have a `measure` function that does something similar, yes:
https://github.com/brancz/symbolicator/blob/f85460aae8dd35c177350dd26e3a3c4c5c3edb79/crates/symbolicator/src/utils/futures.rs#L210-L219

It throws in the task status instead of composing that.  It also emits on every use, statsd style, which is more accurate.

I have mixed feelings about this all and using tokio-metrics here is definitely experimental: it's the very first 0.1 version plus it is modelled more after the more traditional metrics use of e.g. prometheus, hence the need to spawn a task.  OTOH it might or should have less overheads, though it needs to emit more metrics.

PR #708, opened 2022-03-03
ref: Update dependencies

Swatinem:
#skip-changelog

### DIFF
--- a/crates/symbolicator/Cargo.toml
+++ b/crates/symbolicator/Cargo.toml
@@ -11,7 +11,7 @@ anyhow = "1.0.38"
 apple-crash-report-parser = "0.4.2"
 backtrace = "0.3.55"
 base64 = "0.13.0"
-cadence = "0.28"
+cadence = "0.28.0"
 chrono = { version = "0.4.19", features = ["serde"] }
 console = "0.15.0"
 filetime = "0.2.14"
@@ -22,12 +22,12 @@ glob = "0.3.0"
 hostname = "0.3.1"
 humantime-serde = "1.0.1"
 ipnetwork = "0.18.0"
-jsonwebtoken = "8"
+jsonwebtoken = "8.0.1"
 lazy_static = "1.4.0"
 lru = "0.7.0"
 num_cpus = "1.13.0"
 minidump = "0.9.4"
-parking_lot = "0.12"
+parking_lot = "0.12.0"
 procspawn = { version = "0.10.0", features = ["backtrace", "json"] }
 regex = "1.4.3"
 reqwest = { git = "https://github.com/jan-auer/reqwest", tag = "v0.11.0", features = ["gzip", "json", "stream", "trust-dns"] }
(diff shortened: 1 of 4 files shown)


crates/symbolicator/src/services/symcaches.rs
```diff
@@ -390,6 +390,7 @@ fn write_symcache(
                 handle.uuid,
                 object_handle
             );
+            #[allow(deprecated)]
```
flub, line 393: we deprecated an api we ourselves keep using? 😂

flub (approved):

crates/symbolicator/src/services/symcaches.rs
Swatinem: :-D it was a bit overeager, I agree. We should migrate to symcache transformers. But that requires unsafe SelfCell code for now. So creating an `OwnedBcSymbolMap` type or something inside symbolic that is `'static` will solve that.

crates/symbolicator/Cargo.toml
```diff
@@ -22,12 +22,12 @@ glob = "0.3.0"
 hostname = "0.3.1"
 humantime-serde = "1.0.1"
 ipnetwork = "0.18.0"
-jsonwebtoken = "7.2.0"
+jsonwebtoken = "8"
```
jan-auer, line 25:
```suggestion
jsonwebtoken = "8.0.1"
```

crates/symbolicator/Cargo.toml
```diff
@@ -11,7 +11,7 @@ anyhow = "1.0.38"
 apple-crash-report-parser = "0.4.2"
 backtrace = "0.3.55"
 base64 = "0.13.0"
-cadence = "0.27.0"
+cadence = "0.28"
```
jan-auer, line 14:
```suggestion
cadence = "0.28.0"
```

crates/symbolicator/Cargo.toml
```diff
@@ -22,12 +22,12 @@ glob = "0.3.0"
...
 humantime-serde = "1.0.1"
 ipnetwork = "0.18.0"
-jsonwebtoken = "7.2.0"
+jsonwebtoken = "8"
 lazy_static = "1.4.0"
 lru = "0.7.0"
 num_cpus = "1.13.0"
 minidump = "0.9.4"
-parking_lot = "0.11.1"
+parking_lot = "0.12"
```
jan-auer, line 30:
```suggestion
parking_lot = "0.12.0"
```

crates/symbolicator/Cargo.toml
```diff
@@ -50,7 +50,7 @@ tracing = "0.1.29"
 tracing-subscriber = { version = "0.3.6", features = ["tracing-log", "local-time", "env-filter", "json"] }
 tokio = { version = "1.0.2", features = ["rt", "macros", "fs"] }
 tokio-metrics = "0.1.0"
-tokio-util = "0.6"
+tokio-util = { version = "0.7", features = ["io"] }
```
jan-auer, line 53:
```suggestion
tokio-util = { version = "0.7.0", features = ["io"] }
```

jan-auer (commented):
Note that Dependabot will start taking care of this after #709, we could actually skip this manual update.

If you merge this, I'd prefer if we stick to full version numbers for consistency, which matches most documentation examples and also the behavior of `cargo-edit`.

PR #709, opened 2022-03-08
ci: Enable dependabot

flub:
Try out dependabot again.  Once per week should be plenty.

#skip-changelog

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+- package-ecosystem: "cargo"
+  directory: "/"
+  schedule:
+    interval: "weekly"
+    day: "monday"
+    time: "05:00"
+    timezone: "UTC"
+  # Because we only do this weekly, allow for many to be created.
+  open-pull-requests-limit: 20


Swatinem (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+updates:
+- package-ecosystem: "cargo"
+  directory: "/"
+  schedule:
+    interval: "weekly"
+    day: "monday"
+    time: "05:00"
+    timezone: "UTC"
+  # Because we only do this weekly, allow for many to be created.
+  open-pull-requests-limit: 20
```
jan-auer, line 11:
```yaml
versioning-strategy: increase
```

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#versioning-strategy

I think we want `increase` in this case since it's an app and we want the version to be updated in both Cargo.lock and Cargo.toml

jan-auer, line 11:
```yaml
    commit-message:
      prefix: "build:"
```

This would add our required prefix to every commit, and hopefully also the PR.

jan-auer (commented):
Unfortunately there is no option to batch passing updates together. 

Added suggestions for additional configuration in separate comments below.

.github/dependabot.yml
flub: oops, github ui didn't show me you left comments!

anyway, `increase` is not supported by cargo according to the docs.

PR #714, opened 2022-03-09
ci(danger): Do not require changelogs from dependabot

flub:
For plain dependency upgrades that require not changes we normally do
not need any changelog entries.  But not having them fails CI which
renders every dependabot PR a huge hassle.

Sadly we can not customise the dependabot PRs to add a #skip-changelog
marker (see e.g.
https://github.com/dependabot/dependabot-core/issues/2091), so lets
resort to changing the check instead.

There is a risk that someone pushes some changes to such a PR that
does require a changelog.  Hopefully we'll recognise this and do the
right thing anyway.

#skip-changelog

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -3,6 +3,9 @@ on:
   pull_request:
     types: [opened, synchronize, reopened, edited, ready_for_review]
 
+permissions:
+  pull-requests: write
+
 jobs:
   build:
     name: Changelogs
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -42,7 +42,9 @@ function skipChangelog() {
     return true;
   }
   for (let review of danger.github.reviews) {
-    if (review.body && review.body.includes("#skip-changelog")) {
+      if (review.body
+          && review.body.includes("#skip-changelog")
+          && review.body.includes("@dependabot")) {
       return true;
     }
   }
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -34,13 +34,23 @@ async function containsChangelog(path) {
   return contents.includes(PR_LINK);
 }
 
-async function checkChangelog() {
-  const skipChangelog =
-    danger.github
-      && ((danger.github.pr.body + "").includes("#skip-changelog")
-          || danger.github.pr.user.href === "https://github.com/apps/dependabot");
+function skipChangelog() {
+  if (!danger.github) {
+    return false;
+  }
+  if ((danger.github.pr.body + "").includes("#skip-changelog")) {
+    return true;
+  }
+  for (let review of danger.github.reviews) {
+    if (review.body && review.body.includes("#skip-changelog")) {
+      return true;
+    }
+  }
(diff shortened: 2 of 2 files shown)


Swatinem (approved):

jan-auer:
I would like not to have this. We need to review & approve every dependency bump anyway, and as part of that can add a #skip-changelog to the PR description. If dependabot overwrites the description, we should rather update our changelog bot to also check in comments. 

Now the actual reason why I would like to have this: In _some_ cases we do want a changelog entry and it's too easy to forget. Especially for dependency upgrades that fix CVEs, come with user-visible changes, or in the case of `symbolic` (we're setting a precedent here) require downstream changes.

flub:
So I describe the tradeoff made here pretty clearly: if you don't need to push any changes to the PR you won't need a changelog.  But if you do make changes it won't prompt you.

I think I could like the suggestion of looking in comments as well though, it would mean you can write:
```
@dependabot merge
#skip-changelog
```
and it'll just work.  let me change it to that to try it out for a while

jan-auer:
The comment compromise sounds pretty sweet, if dependabot is graceful with the delay. Otherwise, we just comment #skip-changelog, and then merge with the large green button at the bottom.

Counter example for checking for changes: https://github.com/getsentry/relay/pull/1207
In this case we just bumped the dependency, but it also fixed a CVE so it's significant enough to communicate.

flub:
PTAL

jan-auer:
We may have a problem now. Danger checks for a comment with `#skip-changelog`, but it's own comment that it leaves on the PR contains this string...

relaxolotl:
> We may have a problem now. Danger checks for a comment with #skip-changelog, but it's own comment that it leaves on the PR contains this string...
>
> (NB I'm not sure if [reviews](https://danger.systems/js/reference.html#GitHubReview) actually includes all comments or just "comment" reviews)

FWIW the code in the PR currently checks top-level comments included in a review. Let's use #673: the review loop iterates over the entries returned by https://api.github.com/repos/getsentry/symbolicator/pulls/673/reviews. 

If you want to look at all comments, the endpoint (using the above PR) is actually https://api.github.com/repos/getsentry/symbolicator/issues/673/comments. If the intent is to scan through all comments, you probably want to invoke [`danger.github.api.getPullRequestComments`](https://github.com/danger/danger-js/blob/a6b40e953ff318c6230bf0f09152e4ba58cd567f/source/platforms/github/GitHubAPI.ts#L308-L312). The API also has [`danger.github.api.getDangerCommentIDs`](https://github.com/danger/danger-js/blob/a6b40e953ff318c6230bf0f09152e4ba58cd567f/source/platforms/github/GitHubAPI.ts#L85-L96) which you can use to ignore danger's comments and therefore dodge this issue, assuming that's publicly exposed.

flub:
PTAL, this now only accepts `#skip-changelog` in comments that also include a dependabot command.  This should avoid our loop.

Swatinem (approved):

PR #715, opened 2022-03-09
CI: Skip tests when credentials are not available.

flub:
This makes sure that tests from forks which do not have access to the secrets will work seamlessly, just like those tests work seamlessly when you run in local development without the secrets.

The risk of regressing on the GCS functionality is mitigated by the fact CI also runs on master after merging.

#skip-changelog


flub:
Despite the initial title this should now be reviewed and merged 😄

PR #716, opened 2022-03-11
ref: Update rust-minidump, and merge master

Swatinem:
Version `0.10.0` was recently tagged with all our latest fixes.
The only blocker as far as I know of is the broken debug-ids that are missing os-specific logic.

#skip-changelog


Swatinem:
NOTE: make sure to "create a merge commit" when merging ;-)

relaxolotl (approved):

PR #722, opened 2022-03-18
fix: Measure the correct minidump processing time

Swatinem:
The duration was only measuring the stackwalking itself, not the minidump and symbolfile parsing.

#skip-changelog


loewenheim (approved):
That's what I get for being overly literal and only measuring the `process_minidump` function.

PR #727, opened 2022-03-22
ref: Normalize Registers earlier than in Comparison

Swatinem:
The previous code was making changes to the registers that were only used for the comparison, and the diff itself still had those changes.
This also pulls in more of these normalizations, as there are also other differences here.

#skip-changelog


crates/symbolicator/src/services/symbolication/mod.rs
```diff
@@ -2487,11 +2488,30 @@ impl SymbolicationActor {
...
+    let canonical = match name {
+        "r11" if family == CpuFamily::Arm32 => "fp",
+        "r13" if family == CpuFamily::Arm32 => "sp",
+        "r14" if family == CpuFamily::Arm32 => "lr",
+        "r15" if family == CpuFamily::Arm32 => "pc",
+        "eflags" => "efl",
+        "x29" => "fp",
+        "x30" => "lr",
+        name => name,
+    };
```
relaxolotl, line 2502: working off of the comment for the function, are these mappings pulled from some specific place in `rust-minidump` somewhere? can you include some sort of reference to where this is being pulled from?

Swatinem: Its just conventions which might be different in some cases.

Speaking of conventions, it looks like `eflags` is the more widely used variant though.
Interesting also that rust-minidump does not expose it in the x64 context at all.

See https://en.wikipedia.org/wiki/FLAGS_register

Swatinem: symbolic and breakpad also used to use the `eflags` name.
https://github.com/getsentry/symbolic/blob/7294a852dec94e70918ff60b1cf9804dcaa6df96/symbolic-minidump/cpp/data_structures.cpp#L239
https://github.com/getsentry/breakpad/blob/f12b6897897fb5b7e74437371fdf2b36a36f5661/src/processor/dump_context.cc#L490

Swatinem: However, the `stackwalk_minidump` CLI uses `efl`:

https://github.com/getsentry/breakpad/blob/09b056975dacd1f0f815ad820b6dc9913b0118a3/src/processor/stackwalk_common.cc#L311

loewenheim (approved):

PR #749, opened 2022-04-05
ci: auto-approve dependabot PRs

flub:
To smoothen the dependabot workflow a bit let's try out auto-approving
dependabot PRs.  This means we should be able to merge these PRs
without having to do the, checks notes, 4 clicks this takes.  Already
a massive improvement.

#skip-changelog

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,21 @@
+name: Dependabot auto-approve
+on: pull_request
+
+permissions:
+  pull-requests: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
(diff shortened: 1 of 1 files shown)


loewenheim (approved):
Looks sensible to me, FWIW. I don't have a lot of experience with these CI files.

relaxolotl:
To be clear, what's the workflow after this is merged in? Does it still require some kind of manual action from a developer to finalize the change, or does this automatically approve and merge in dependabot PRs?

flub:
Yep, this does not mean PRs will be auto-merged!  We'll still have to merge this with an explicit action, but we won't have to go do the approval stuff.  I hope, that together with #714 - which still needs updating - I can get merging of dependabot PRs down to creating one comment containing
```
@dependabot merge
#skip-changelog
```
At least in the best case when they can be merged without any further work.

Issue #728, opened 2022-03-22
Error while downloading file: failed to fetch data from S3


mk4437:
### Environment

Self-hosted symbolicator 0.4.2 on amazonlinux2

### Steps to Reproduce

Send /applecrashreport requests

### Expected Result

No ERROR messages logged

### Actual Result

Similar to https://github.com/getsentry/symbolicator/issues/640, we intermittently see error messages like the following:

```
ERROR symbolicator::services::objects::data_cache > Error while downloading file: failed to fetch data from S3
caused by: expected value at line 1 column 1
caused by: expected value at line 1 column 1
```
We use two S3 buckets, one for system symbol files and another for application symbol files.  All files end without a suffix.  So four locations get searched for each symbol file (two S3 buckets, with and without .app).  The errors when they occur are for the non-existent locations.  

In spot-checking I do usually see the downloaded file in symcache, though in at least one case one of our instances showed size 0 in symcache.

Our cache settings are as follows:
```
caches:
  downloaded:
    retry_misses_after: 0s
  derived:
    retry_misses_after: 0s
```
and in each `sources` entry:
```
    filters:
      filetypes:
        - "mach_debug"
        - "mach_code"
    layout:
      type: "native"
      casing: "lowercase"
```

mk4437:
Additionally we are seeing a lot of WARN messages like these in the production logs.
```
WARN goblin::mach::segment > section #1 size 23184212 out of bounds
WARN goblin::mach::segment > section #4 size 321015 out of bounds
WARN goblin::mach::segment > section #5 size 376576 out of bounds
WARN goblin::mach::segment > section #6 size 647699 out of bounds
WARN goblin::mach::segment > section #7 size 86438 out of bounds
WARN goblin::mach::segment > section #8 size 898758 out of bounds
WARN goblin::mach::segment > section #9 size 47120 out of bounds
```

flub:
The WARN messages are something else, they come from parsing a successfully downloaded DIF.  It probably means your compiler generates debugging info which we can't parse, and thus can't use, and is maybe more likely to be a bug on our end than the compiler.  But please separate this out into a different issue (maybe even on the symbolic repo as that's probably where the bug is, but symbolicator is probably fine too).

[...]

PR #754, opened 2022-04-10
Automatically refresh expired container credentials

josb:
So far, testing by @mk4437 suggests this patch addresses the S3 errors [#728] we are seeing. Happy to tweak this patch further as needed.


crates/symbolicator/src/services/download/s3.rs
```diff
@@ -116,7 +116,11 @@ impl S3Downloader {
             );
             let s3 = Arc::new(match key.aws_credentials_provider {
                 AwsCredentialsProvider::Container => {
-                    let provider = rusoto_credential::ContainerProvider::new();
+                    let container_provider = rusoto_credential::ContainerProvider::new();
+                    let provider = match rusoto_credential::AutoRefreshingProvider::new(container_provider) {
+                        Ok(provider) => provider,
+                        Err(err) => panic!("Unable to instantiate rusoto_credential::AutoRefreshingProvider: {:?}", err),
```
Swatinem, line 122: As source configs are being fed into symbolicator with each request, we should avoid panic-ing as much as possible. I haven’t looked at the docs to figure out the exact failure conditions for this `Result`, it might as well be an extreme edge-case condition in which case it might as well be okay to `unwrap`.

Swatinem (commented):
Might be good to clear up that panic/unwrap situation. Otherwise this looks good, apart from CI complaining.

The Changelog check should have actually suggested a code change, but that seems to have broken recently; I will look into that.

crates/symbolicator/src/services/download/s3.rs
josb: Looking at the [source](https://rusoto.github.io/rusoto/src/rusoto_credential/lib.rs.html#243-251) of `rusoto_credential::AutoRefreshingProvider::new`, I don't even understand how it can return `CredentialsError`?

PR #755, opened 2022-04-10
Don't emit ANSI escape codes with auto=false or simplified logging formats

josb:
Fixes [#753].


Swatinem (approved):

josb:
Would you like me to add a changelog entry for this?

relaxolotl:
@josb if you don't mind doing so, that'd be great. It looks like dangerjs is struggling a little right now, but the check should pass once you add in a changelog entry. Feel free to add this into `CHANGELOG.md` as a fix under the unreleased section - there should be other existing entries to reference if you need any guidance on what one should look like.

If you could do the same for your other PR (#754) that'd also be much appreciated.

josb:
[Added](https://github.com/josb/symbolicator/commit/d2a0a38fcb90de830ac47d8b64250d7f000744d3) CHANGELOG.md entry.

josb:
Thanks, @relaxolotl !

relaxolotl:
I'll leave #754 up to @Swatinem since he's already taking care of it 👍

josb:
Sounds good!

PR #783, opened 2022-04-25
deps: upgrade rusoto

flub:
Upgrade rusoto crate, replaces the 3 PRs from dependabot which doesn't
do this as a single thing.

#skip-changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2484,9 +2484,9 @@ dependencies = [
 
 [[package]]
 name = "rusoto_core"
-version = "0.47.0"
+version = "0.48.0"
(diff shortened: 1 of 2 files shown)


Swatinem (approved):
Oh, I thought there were breaking changes and that was why this failed.
I think you can configure dependabot to update certain packages as a unit. That makes sense for this, the sentry sdk, symbolic, and rust-minidump.

flub:
> I think you can configure dependabot to update certain packages as a unit. That makes sense for this, the sentry sdk, symbolic, and rust-minidump.

Yes, though there's a contributor working on replacing rusoto with the aws sdk plus the release notes claim this is the last rusoto release... so i decided not to bother configuring that.

PR #784, opened 2022-04-26
deps: Cleanup cargo.lock file

flub:
It seems that dependabot has left it in an inconsistent state?  Kind
of surprised CI didn't balk at that.

#skip-changelog


loewenheim (approved):

Swatinem:
> Kind of surprised CI didn't balk at that.

I think we would need to run with `--locked` for that?

PR #786, opened 2022-04-27
meta(changelog): Summarise breakpad to rust-minidump migration

flub:
There's no need to have these as a narrative riddle in the changelog,
collapse it into one digestable entry.


Swatinem (approved):
oh, the irony of this needing a `#skip-changelog` :-D

PR #787, opened 2022-04-27
ref: Thread `CfiCacheModules` through iterative stackwalking

Swatinem:
This was an experiment trying to thread `CfiCacheModules` through iterative stackwalking, with the aim of eventually merging it with `ModuleListBuilder` and `SymbolicatorSymbolProvider` to avoiding re-parsing of already parsed cfi caches (which is expensive).

Not sure this experiment was worth it though. There is a ton of work ahead to make this code better.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/services/symbolication/process_minidump.rs
+++ b/crates/symbolicator/src/services/symbolication/process_minidump.rs
@@ -43,8 +43,9 @@ use super::{
 type CfiCacheResult = (DebugId, Result<Arc<CfiCacheFile>, Arc<CfiCacheError>>);
 type Minidump = minidump::Minidump<'static, ByteView<'static>>;
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug)]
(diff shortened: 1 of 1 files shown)


loewenheim (commented):
The one thing I don't like about this is the addition of `cfi_caches` to `StackwalkMinidumpResult`. AFAICT the result is never interesting, it's always exactly what was passed into `stackwalk`. Does `stackwalk` need to take `cfi_caches` by value?

Swatinem:
> Does `stackwalk` need to take `cfi_caches` by value?

It does, because of the task being spawned on another runtime. Essentially what I complained about during the meetings yesterday :-(

Swatinem:
That said, we have a jira ticket tracking the iterative vs async stack walking, though the thing that causes problems here specifically is crossing the async runtime boundaries :-(

Swatinem:
#838 makes this fully async so this is not needed anymore

PR #829, opened 2022-06-06
Add support for serving web requests using HTTPS

josb:
The config file format has been extended with a bind_https keyword,
as well as

    server_config:
      https:
        certificate_path: cert.pem
        key_path: key.pem

### DIFF
--- a/crates/symbolicator/Cargo.toml
+++ b/crates/symbolicator/Cargo.toml
@@ -10,7 +10,7 @@ anyhow = "1.0.57"
 apple-crash-report-parser = "0.4.2"
 async-trait = "0.1.53"
 axum = { version = "0.5.4", features = ["multipart"] }
-axum-server = { version = "0.4.0", features = [ "tls-rustls"] }
+axum-server = { version = "0.4.0" }
 backtrace = "0.3.65"
 base64 = "0.13.0"
 cadence = "0.29.0"
--- a/crates/symbolicator/Cargo.toml
+++ b/crates/symbolicator/Cargo.toml
@@ -10,6 +10,7 @@ anyhow = "1.0.57"
 apple-crash-report-parser = "0.4.2"
 async-trait = "0.1.53"
 axum = { version = "0.5.4", features = ["multipart"] }
+axum-server = { version = "0.4.0", features = [ "tls-rustls"] }
 backtrace = "0.3.65"
 base64 = "0.13.0"
 cadence = "0.29.0"
--- a/crates/symbolicator/Cargo.toml
+++ b/crates/symbolicator/Cargo.toml
@@ -65,3 +65,6 @@ reqwest = { git = "https://github.com/jan-auer/reqwest", tag = "v0.11.0", featur
 sha-1 = "0.10.0"
 test-assembler = "0.1.5"
 warp = "0.3.0"
+
+[features]
+https = ["axum-server/tls-rustls"]
--- a/crates/symbolicator/Cargo.toml
+++ b/crates/symbolicator/Cargo.toml
@@ -67,5 +67,4 @@ test-assembler = "0.1.5"
 warp = "0.3.0"
 
 [features]
-default = ["https"]
 https = ["axum-server/tls-rustls"]
--- a/crates/symbolicator/Cargo.toml
+++ b/crates/symbolicator/Cargo.toml
@@ -67,4 +67,5 @@ test-assembler = "0.1.5"
 warp = "0.3.0"
 
 [features]
+default = ["https"]
 https = ["axum-server/tls-rustls"]
--- a/crates/symbolicator/src/config.rs
+++ b/crates/symbolicator/src/config.rs
@@ -28,6 +28,7 @@ pub enum LogFormat {
     Json,
 }
 
+#[cfg(feature = "https")]
 #[derive(Clone, Debug, Eq, PartialEq, Deserialize)]
 #[serde(rename_all = "lowercase")]
 pub struct HTTPSConfig {
@@ -35,6 +36,7 @@ pub struct HTTPSConfig {
     pub key_path: PathBuf,
 }
 
+#[cfg(feature = "https")]
 impl Default for HTTPSConfig {
     fn default() -> Self {
         HTTPSConfig {
@@ -44,7 +46,8 @@ impl Default for HTTPSConfig {
     }
 }
 
-/// Controls the HTTP server setup
+/// Controls the HTTPS server setup
+#[cfg(feature = "https")]
 #[derive(Clone, Debug, Default, Deserialize)]
 #[serde(default)]
 pub struct ServerConfig {
@@ -267,6 +270,7 @@ pub struct Config {
     pub bind: String,
 
     /// Host and port to bind the HTTPS webserver to.
+    #[cfg(feature = "https")]
     pub bind_https: Option<String>,
 
     /// Configuration for internal logging.
@@ -275,6 +279,7 @@ pub struct Config {
     /// Configuration for reporting metrics to a statsd instance.
     pub metrics: Metrics,
 
+    #[cfg(feature = "https")]
     pub server_config: ServerConfig,
 
     /// DSN to report internal errors to
@@ -394,8 +399,10 @@ impl Default for Config {
         Config {
             cache_dir: default_cache_dir(),
             bind: default_bind(),
+            #[cfg(feature = "https")]
             bind_https: None,
             logging: Logging::default(),
+            #[cfg(feature = "https")]
             server_config: ServerConfig::default(),
             metrics: Metrics::default(),
             sentry_dsn: None,
--- a/crates/symbolicator/src/config.rs
+++ b/crates/symbolicator/src/config.rs
@@ -28,6 +28,36 @@ pub enum LogFormat {
     Json,
 }
 
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub struct HTTPSConfig {
+    pub certificate_path: PathBuf,
+    pub key_path: PathBuf,
+}
+
+impl Default for HTTPSConfig {
+    fn default() -> Self {
+        HTTPSConfig {
+            certificate_path: PathBuf::from("cert.pem"),
+            key_path: PathBuf::from("cert.pem"),
+        }
+    }
+}
+
+/// Controls the HTTP server setup
+#[derive(Clone, Debug, Deserialize)]
+#[serde(default)]
+pub struct ServerConfig {
+    /// The log level for the relay.
+    pub https: Option<HTTPSConfig>,
+}
+
+impl Default for ServerConfig {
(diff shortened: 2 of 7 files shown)


josb:
I saw the integration tests fail but I have no idea what the actual error is. Any ideas?

crates/symbolicator/src/server.rs
```diff
@@ -29,24 +39,46 @@ pub fn run(config: Config) -> Result<()> {
...
-    tracing::info!("Starting server on {}", server.local_addr());
-
-    web_pool.block_on(server)?;
+    if let Some(ref bind_str) = config.bind_https {
+        let https_conf = match config.server_config.https {
+            None => panic!("Need HTTPS config"),
+            Some(ref conf) => conf,
+        };
+        let service_https = web_pool
+            .block_on(Service::create(
```
Swatinem, line 63: Please re-use the existing service, as that is meant to be a global singleton. Having two services would mean that you wouldn’t share caches properly, and some dedupe logic would be missing.

For example, starting a request via http and polling via https would fail. connections to external symbol servers would be duplicated, etc.

The Service is `Clone` so that should not be a problem.

crates/symbolicator/src/server.rs
```diff
@@ -29,24 +39,46 @@ pub fn run(config: Config) -> Result<()> {
...
+                cpu_pool.handle().to_owned(),
+            ))
+            .context("failed to create HTTPS service state")?;
+        let socket_https = bind_str.parse::<SocketAddr>()?;
+        let certificate = read_pem_file(&https_conf.certificate_path)?;
+        let key = read_pem_file(&https_conf.key_path)?;
+        let tls_config =
+            web_pool.block_on(async { RustlsConfig::from_pem(certificate, key).await })?;
+        let server_https = axum_server::bind_rustls(socket_https, tls_config)
+            .serve(endpoints::create_app(service_https).into_make_service());
```
Swatinem, line 75: A bit similarly, I’m a bit unsure if these types are `Clone` or not. It would be nice if we could share the "endpoints" across these two servers. Though I doubt this has any negative consequences apart from maybe memory usage.

crates/symbolicator/Cargo.toml
```diff
@@ -10,6 +10,7 @@ anyhow = "1.0.57"
 apple-crash-report-parser = "0.4.2"
 async-trait = "0.1.53"
 axum = { version = "0.5.4", features = ["multipart"] }
+axum-server = { version = "0.4.0", features = [ "tls-rustls"] }
```
flub, line 13: would it be worth putting this behind a cargo feature?

/cc @Swatinem

Swatinem: That might be a good idea, yes. Outgoing connections use native-tls (aka openssl) for now.

crates/symbolicator/src/server.rs
josb: Thanks @Swatinem, fixed!

crates/symbolicator/src/config.rs
```diff
@@ -28,6 +28,30 @@ pub enum LogFormat {
...
+            key_path: PathBuf::from("cert.pem"),
+        }
+    }
+}
+
+/// Controls the HTTP server setup
+#[derive(Clone, Debug, Default, Deserialize)]
+#[serde(default)]
+pub struct ServerConfig {
+    /// The log level for the relay.
```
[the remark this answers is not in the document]
josb: Yeah :-) Fixed.

crates/symbolicator/src/server.rs
Swatinem: Ah nice, I see you directly deduped the whole `endpoints::create_app(service_https).into_make_service()` expression, which is what I asked about ;-)

crates/symbolicator/Cargo.toml
josb: So you want me to make the https server optional then? I'll have to research how to do that; first I'll try moving its creation into a separate function (and for symmetry, do the same thing for the HTTP server). Any suggestions about how to go about this feature business are appreciated. :-)

Swatinem: There is 3 parts to that:
1. Mark the `axum-server` dependency as `optional: true`
2. Introduce a `[features]` section with at least a `https = ["axum-server"]` entry. If you want to make this opt-out instead of opt-in by default, you can set `default = ["https"]` as well.
3. In code, you can sprinkle `#[cfg(feature = "https")]` all over the place, for example above code blocks, which are then conditionally compiled.

Swatinem:
I’m not entirely sure why the integration tests are hanging, but it seems they are just waiting forever for something, so I think thats related to this change.

crates/symbolicator/Cargo.toml
josb: Thank you for the extra detail, I'll fix this.

josb:
I think the integration test is looking for the actual port number, but instead hangs trying to connect to port 0 because my PR doesn't log the actual port? Instead of `local_addr() `, there's a `Handle` struct which offers a `listening` method which is supposed to return the actual listen port, but since that method is `async` I've been unable to make it work. 😔

crates/symbolicator/Cargo.toml
josb: I just realized that if the axum-server dependency is optional, the HTTP server will necessarily need to continue to use axum. This will lead to two different web server frameworks to be used, with different APIs and abilities. Not sure if that's a good thing...

Swatinem: Ah absolutely, I missed that. In that case, you can have `https = ["axum-server/tls-rustls"]` as the feature, which will activate that specific feature of an existing dependency.

crates/symbolicator/src/server.rs
```diff
@@ -1,12 +1,22 @@
...
+use axum_server::tls_rustls::RustlsConfig;
+use futures::future::try_join_all;
+use futures::future::BoxFuture;
+
 use crate::config::Config;
 use crate::endpoints;
 use crate::services::Service;
 
-/// Starts all actors and HTTP server based on loaded config.
+fn read_pem_file(path: &PathBuf) -> Result<Vec<u8>> {
```
Swatinem, line 15: I see there is `RustlsConfig::from_pem_file`, so you don’t need to read these files manually.

crates/symbolicator/Cargo.toml
josb: Thanks, I'll look into fixing this.

crates/symbolicator/src/server.rs
Swatinem: You are wrapping `RustlsConfig::from_pem` is already async, and you are wrapping it in a `block_on`. So I’m not sure what the benefit would be to read the pem files ahead of time with sync code, just to pass them into an async constructor anyway.

crates/symbolicator/Cargo.toml
josb: Just pushed a change to make the https server an optional feature.

crates/symbolicator/src/server.rs
josb: The main motivation is to be able to provide accurate errors about which file is missing, and to avoid the test-then-open race.

Honestly it's not clear to me why `RustlsConfig::from_pem` is even `async`, it seems to just complicate its use here. But if you would like me to implement this part using `RustlsConfig::from_pem_file`, I'm happy to do so, it's not a big deal.

josb: Fair. With `from_pem_file0,` a missing `key.pem` file yields this error and backtrace:
```
  2022-06-13T23:08:36.173987Z  INFO symbolicator::server: Starting HTTP server on 127.0.0.1:3021
    at crates/symbolicator/src/server.rs:62

  2022-06-13T23:08:36.181959Z ERROR symbolicator::logging: failed to start the server

Caused by:
    No such file or directory (os error 2)

Stack backtrace:
   0: backtrace::backtrace::trace_unsynchronized
   1: backtrace::backtrace::trace
   2: anyhow::backtrace::capture::Backtrace::create
   3: anyhow::backtrace::capture::Backtrace::capture
   4: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
   5: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
   6: symbolicator::server::run
   7: symbolicator::cli::execute
   8: symbolicator::main
   9: core::ops::function::FnOnce::call_once
  10: std::sys_common::backtrace::__rust_begin_short_backtrace
  11: std::rt::lang_start::{{closure}}
  12: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &F>::call_once
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/ops/function.rs:259:13
      std::panicking::try::do_call
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:492:40
      std::panicking::try
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:456:19
      std::panic::catch_unwind
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panic.rs:137:14
      std::rt::lang_start_internal::{{closure}}
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/rt.rs:128:48
      std::panicking::try::do_call
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:492:40
      std::panicking::try
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:456:19
      std::panic::catch_unwind
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panic.rs:137:14
      std::rt::lang_start_internal
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/rt.rs:128:20
  13: std::rt::lang_start
  14: _main
    at crates/symbolicator/src/logging.rs:85
```
It seems clearer to be able to say `unable to read file: key.pem`.

crates/symbolicator/Cargo.toml
Swatinem: In that case you don’t need the `features` here, as that feature is activated by the feature flag down below.

Swatinem (commented):
This looks good now. However fixing the integration tests is a hard blocker to move forward with this.

josb:
Thanks for all the feedback so far! I'll take another stab at resolving the integration test issue in the next few days.

josb:
FWIW, [this comment](https://github.com/programatik29/axum-server/issues/46#issuecomment-1156821583) says my approach to printing the listening ports is fine.

josb:
@Swatinem @flub any other suggestions to improve this PR, please?

Swatinem (approved):

flub (approved):

PR #836, opened 2022-06-16
ci: automatic submodule dependency update

vaind:
See https://github.com/getsentry/sentry-unity/issues/788 for context

sample PR: https://github.com/vaind/sentry-symbolicator/pull/2

#skip-changelog


Swatinem (approved):
IMO doing this once every week to every month is sufficient. Updating our native SDK dependency is super low priority.

PR #844, opened 2022-06-23
ref: Clean up tokio::runtime usage

Swatinem:
Instead of explicit spawning of tasks, this moves the responsibility which runtimes a certain task is being spawned on to the `Cacher`.

The idea is that incoming requests from the `web_pool` get spawned onto the `cpu_pool` for processing.
This processing is synchronous and cpu-bound in the happy path when every needed file is already cached.
In case some needed files are not in the cache already, the cache computation is spawned on the cacher runtime.
That runtime is the `cpu_pool` for cfi/symcaches, and the `io_pool` for the underlying file downloads.

The changes to shared cache are something I’m not entirely happy with though :-( Mostly because we now spawn the shared cache fetch as a task which needs to be `'static`.

This PR is based on top of #838
#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Use Unity il2cpp line mapping files in symcache creation ([#831](https://github.com/getsentry/symbolicator/pull/831))
 - Read thread names from minidumps and Apple crash reports ([#834](https://github.com/getsentry/symbolicator/pull/834))
 
+### Internal
+
+- Fetch CFI on-demand during stackwalking ([#838](https://github.com/getsentry/symbolicator/pull/838))
+
 ## 0.5.1
 
 ### Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 - Use Unity il2cpp line mapping files in symcache creation ([#831](https://github.com/getsentry/symbolicator/pull/831))
 - Read thread names from minidumps and Apple crash reports ([#834](https://github.com/getsentry/symbolicator/pull/834))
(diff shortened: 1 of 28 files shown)


Swatinem:
TODO for self:
`DownloadService::list_files` also does network IO, but *only* in the case of the sentry source, and that function is called in a bunch of places. We would want to spawn those requests (if they are not in the LRU cache) on the `io_pool` as well.

flub (commented):
Ooh, very nice!  I wasn't expecting to fix even the old bad spawning, but this makes sense.  Originally there was the idea that the boundary of a "service" meant that it'd spawn and make sure it's resources where handled correctly internally, but that did get watered down over time.  You're sort of having to bring back that idea in some places here which in interesting to see.

I think the meta_cache is mainly computational, it is the data_cache which does the downloading, so the objects actor probably should get both pools.
Furthermore the data_cache part of the object actor also has a CPU intensive part: it unzips etc.  So maybe it makes more sense to put the entire objects actor on the cpu pool and instead create the download service with the io-pool and make sure calls to it all spawn onto this pool?  (for some reason you already do this for the sentry source but not for the others, i didn't really follow that so might have missed something)

Swatinem:
> the object actor also has a CPU intensive part: it unzips etc. So maybe it makes more sense to put the entire objects actor on the cpu pool and instead create the download service with the io-pool

Indeed, so all of the cachers run on the cpu pool anyways, so there is no reason to carry that around explicitly. So it is indeed only the downloader which needs an explicit spawn here.

loewenheim (approved):

PR #849, opened 2022-07-03
Migrate from rusoto to aws-sdk-rust

### LINKED ISSUE #575: Migrate from rusoto to aws-sdk-rust
https://github.com/awslabs/aws-sdk-rust is shaping up to be a great long-term replacement for https://github.com/rusoto/rusoto; this project should consider migrating to it.

josb:
Fixes #575. Posting this for feedback.


crates/symbolicator/src/services/download/mod.rs
```diff
@@ -53,8 +55,10 @@ pub enum DownloadError {
     Sentry(#[from] sentry::SentryError),
     #[error("failed to fetch data from S3")]
     S3(#[from] s3::S3Error),
-    #[error("S3 error code: {1} (http status: {0})")]
-    S3WithCode(StatusCode, String),
+    #[error("aws-sdk: failed to fetch data from S3")]
+    S3SDK(#[from] aws_smithy_http::byte_stream::Error),
```
flub, line 59: I believe this should be `S3Sdk`?  Doesn't clippy complain about this?

crates/symbolicator/src/services/download/mod.rs
```diff
@@ -53,8 +55,10 @@ pub enum DownloadError {
...
+    // #[error("S3 error code: {1} (http status: {0})")]
+    // S3WithCode(StatusCode, String),
```
flub, line 61: should probably be removed

crates/symbolicator/src/services/download/s3.rs
```diff
@@ -116,37 +120,45 @@ impl S3Downloader {
...
-                            Ok(provider) => provider,
-                            Err(err) => panic!(
-                            "Unable to instantiate rusoto_credential::AutoRefreshingProvider: {:?}",
-                            err
-                        ),
-                        };
-                    self.create_s3_client(provider, region)
+                    let provider = LazyCachingCredentialsProvider::builder()
+                        .load(provide_credentials_fn(|| async {
+                            aws_config::ecs::EcsCredentialsProvider::builder()
```
flub, line 125: we should check with the folks who added the `AutoRefreshingProvider` and ask them to test this to see if it works equivalently.  I recall at some point they had an outage around this.  git blame should point in the right direction.

crates/symbolicator/src/services/download/s3.rs
```diff
@@ -164,94 +176,75 @@ impl S3Downloader {
...
-                // https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html
-                match &err {
-                    RusotoError::Service(_) => return Ok(DownloadStatus::NotFound),
-                    RusotoError::Unknown(response) if response.status == StatusCode::FORBIDDEN => {
-                        return Err(DownloadError::Permissions)
+            Ok(Err(err0)) => {
+                tracing::debug!("Skipping response from s3://{}/{}: {}", bucket, &key, err0);
+                return match &err0 {
+                    ConstructionFailure(err) => {
+                        println!("ERROR: ConstructionFailure: {:?}", err);
```
flub, line 197: looks like you're still debugging all this.  But you could capture all in a single match arm and maybe capture it as a sentry error before throwing cancelled up the chain?

crates/symbolicator/src/sources.rs
```diff
@@ -125,38 +127,59 @@ pub struct FilesystemSourceConfig {
...
+                "us-east-2",
+                "us-gov-east-1",
+                "us-gov-west-1",
+                "us-west-1",
+                "us-west-2",
+            ];
+            let pos = AWS_REGIONS.iter().position(|&x| x == value);
+            match pos {
+                Some(_) => Ok(Region::new(String::from(value))),
+                None => Err(E::custom(format!("unknown region: {}", value))),
```
flub, line 177: IIUC this means we need to updated our code every time they add a region?  That might not be ideal?

crates/symbolicator/src/sources.rs
```diff
@@ -125,38 +127,59 @@ pub struct FilesystemSourceConfig {
...
+    struct SdkRegion;
 
-    impl<'de> serde::de::Visitor<'de> for RusotoRegion {
-        type Value = rusoto_core::Region;
+    impl<'de> serde::de::Visitor<'de> for SdkRegion {
+        type Value = Region;
 
         fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
-            formatter.write_str("string or tuple")
+            formatter.write_str("string")
```
flub, line 143: is this type change going to be a problem for some configurations?

flub (commented):
thanks for picking this up again!

crates/symbolicator/src/sources.rs
josb: I updated this code to use the EC2 `describe_regions` function and cache the result. Things I don't like about my solution:
* I'm calling `tokio::runtime::Runtime::new()` to grab a runtime so I can call `block_on` on it. This is needed because the `describe_regions` call uses an `async` `send` wrapper. Would it not be more appropriate to use one of the existing runtimes created in `server.rs`?
* I'm creating a `shared_config` using `aws_config::from_env().load().await`; not even sure this will work, given that it should really use the `credentials_provider` (`Container` or `Static`) associated with the current source. Thoughts?

Also, does the use of the `#[cached]` macro look OK?

crates/symbolicator/src/services/download/mod.rs
josb: I don't see any complaints from Clippy. It's only used [here](https://github.com/josb/symbolicator/blob/task/aws-sdk-rust-migration/crates/symbolicator/src/services/download/s3.rs#L245). Do you want me to rename it?

crates/symbolicator/src/services/download/mod.rs
josb: Removed; I left it as a reminder that it's possible to report the HTTP status code using the `from` macro, although the mechanics of how aren't clear to me yet.

crates/symbolicator/src/services/download/s3.rs
```diff
@@ -9,21 +9,22 @@ use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
 
+use aws_config::meta::credentials::provide_credentials_fn;
+use aws_config::meta::credentials::LazyCachingCredentialsProvider;
+use aws_sdk_s3::error::GetObjectErrorKind;
+use aws_sdk_s3::types::SdkError::*;
```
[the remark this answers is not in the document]
josb: Done.

crates/symbolicator/src/sources.rs
josb: Yes; custom regions, which use tuples per the [docs](https://github.com/getsentry/symbolicator/blob/6fd6e55fb0fd6188638bcbf6d5697fafe9d1090e/docs/api/index.md), don't currently work, as the AWS Rust SDK doesn't support them.

crates/symbolicator/src/sources.rs
josb: Ugh, this does indeed break the unit tests. Have to rethink this, because we need the region list during `serde` config parsing yet we won't have the credentials until _after_ parsing. And there's no way to get the region list anonymously beforehand so we can pre-populate the list.

crates/symbolicator/src/services/download/mod.rs
flub: yeah, let's rename this

crates/symbolicator/src/sources.rs
```diff
@@ -124,39 +127,56 @@ pub struct FilesystemSourceConfig {
     pub files: CommonSourceConfig,
 }
 
+#[cached]
+fn aws_regions() -> Vec<aws_sdk_ec2::model::Region> {
```
Swatinem, line 131: Maybe using something like `once_cell::Lazy` might be a better idea than a proc macro just for this.
Also this answers my question of why we are pulling in the `ec2` crate. Isn’t there a better way to do this?

crates/symbolicator/src/services/download/mod.rs
josb: Renamed.

crates/symbolicator/src/services/download/s3.rs
```diff
@@ -83,29 +84,33 @@ pub struct S3Downloader {
...
     }
 
-    fn get_s3_client(&self, key: &Arc<S3SourceKey>) -> Arc<rusoto_s3::S3Client> {
-        let mut container = self.client_cache.lock();
-        if let Some(client) = container.get(&*key) {
+    async fn get_s3_client(&self, key: &Arc<S3SourceKey>) -> Arc<Client> {
+        if let Some(client) = {
+            let mut container = self.client_cache.lock();
+            let val = container.get(&*key).cloned();
+            drop(container);
```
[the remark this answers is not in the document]
josb: Dropped the `drop()`!

crates/symbolicator/src/services/download/s3.rs
```diff
@@ -164,94 +177,75 @@ impl S3Downloader {
...
+                // TODO Verify this is still the correct action to take with aws-sdk-rust
                 // Timed out
                 return Err(DownloadError::Canceled);
             }
         };
 
-        let stream = match response.body {
-            Some(body) => body.map_err(DownloadError::Io),
-            None => {
+        let content_length = u32::try_from(response.content_length).ok();
```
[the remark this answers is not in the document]
josb: Shouldn't `content_length_timeout` be updated to accept an `i64`?

crates/symbolicator/src/sources.rs
josb: Yes, I'll do that.

By the way, Rusoto currently also hardcodes the list of [known regions](https://rusoto.github.io/rusoto/rusoto_core/region/enum.Region.html), so if a new region is added, a recompile is needed anyway once Rusoto is updated to include it.

josb:
I'm still slowly chipping away at this, currently focused on fixing the S3 `no_permission` integration test. Addressing this runs into the fact that the API is very different, notably empty user credentials produce a different error than Rusoto. I'm currently trying to make non-empty but fake credentials as used by the test work.

crates/symbolicator/src/services/download/s3.rs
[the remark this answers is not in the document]
josb: @flub @Swatinem any thoughts about this particular issue?

crates/symbolicator/src/services/download/s3.rs
josb: Sounds like a good idea but I am not sure how to go about that, any pointers? Doesn't it require access to a Sentry installation and corresponding API keys?

crates/symbolicator/src/services/download/s3.rs
josb: I'll investigate this. Once I have this branch working well enough, I can also test this myself to some extent.

crates/symbolicator/src/services/download/s3.rs
```diff
@@ -164,94 +177,75 @@ impl S3Downloader {
...
+        let timeout = content_length.map(|cl| content_length_timeout(cl, self.streaming_timeout));
+
+        let stream = match content_length {
+            Some(0) | None => {
```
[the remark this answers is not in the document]
josb: I have changed this code but I am still not sure it's correct. But at least `response.content_length` is an `i64`, should that not solve any file size concerns?

crates/symbolicator/src/services/download/s3.rs
```diff
@@ -116,37 +120,45 @@ impl S3Downloader {
...
+                container.put(key.clone(), s3.clone())
+            };
             s3
         }
     }
 
-    fn create_s3_client<P: ProvideAwsCredentials + Send + Sync + 'static>(
+    async fn create_s3_client<P: ProvideCredentials + Send + Sync + 'static>(
         &self,
         provider: P,
```
josb, line 153: Why not use this simpler syntax instead?
```rust
    async fn create_s3_client(
        &self,
        provider: impl ProvideCredentials + Send + Sync + 'static,
        region: Region,
    ) -> Client {
        let shared_config = aws_config::from_env()
            .credentials_provider(provider)
            .region(region)
            .load()
            .await;
        Client::new(&shared_config)
    }
```

flub: Sure, it might well be that this code was from before `impl trait` was allowed in parameters.  Feel free to change if if you prefer, I don't think there's a particular reason here to prefer the parametrised version.

josb: Thanks, commited in [2bf55d0](https://github.com/getsentry/symbolicator/pull/849/commits/2bf55d070bd2764ae47fa50f263e70ab830b56a3).

crates/symbolicator/src/services/download/s3.rs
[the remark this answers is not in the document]
josb: I [updated](https://github.com/getsentry/symbolicator/pull/849/commits/0d3c4857401dd0730ae9fc0a94ac5701c6b666f9) the code to accept an `i64`; this included some changes in the other source providers.

Swatinem:
@josb thanks for keeping this PR up to date!

Whats the status? I believe there is a bunch of `println` calls and a merge marker still in there.

crates/symbolicator/src/services/download/s3.rs
josb: This is still pending me looking into this.

crates/symbolicator/src/services/download/s3.rs
josb: Looking for some guidance here as I know nothing about Sentry. SOme of these can probably grouped into a single response but the mapping from Rusoto errors to AWS SDK errors isn't clear to me (although granted I have not spent much time investigating this yet).

Also, the Rusoto code does some XML parsing of the AWS responce, I wonder how much of that is still required with the AWS SDK.

crates/symbolicator/src/sources.rs
josb: How about I add a simple test at startup to see if `sts get-caller-identity` works and if it fails, error out with an appropriate error message?

crates/symbolicator/src/sources.rs
josb: I don't think this is in-scope for the AWS SDK, at any rate it doesn't support it today so if we want to offer this functionality we'll have to add it ourselves. :-/

crates/symbolicator/src/services/download/s3.rs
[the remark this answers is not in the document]
josb: Does this solution look OK?

josb:
Still working on this, albeit slowly. I replied to some of the comments; the main issues are the failing integration test/addressing the error handling differences between the Rusoto code and the AWS SDK code, as well as the AWS region/credentials check as outlined in an earlier comment.

josb:
@Swatinem @flub I've been trying to merge the recent changes to `master` in my branch but am stuck on how to call `async` methods from inside the closure passed to `self.client_cache.get_with_by_ref`, any suggestions how to go about this? That closure needs to satisfy `impl FnOnce` but at the same time allow `async` functions to be called and `await`ed on. So far I've had no luck. Please help!

Swatinem:
I yes, we introduced `moka` for some in-memory LRU caches like the `client_cache`. It has a `moka::future` variant that has almost the same API that can accept futures.

crates/symbolicator-service/src/services/download/s3.rs
```diff
@@ -88,56 +90,62 @@ impl S3Downloader {
...
+                );
+                Arc::new(match key.aws_credentials_provider {
+                    AwsCredentialsProvider::Container => {
+                        let provider = LazyCachingCredentialsProvider::builder()
+                            .load(provide_credentials_fn(|| async {
+                                aws_config::ecs::EcsCredentialsProvider::builder()
+                                    .build()
+                                    .credentials()
+                                    .await
+                            }))
```
Swatinem, line 121:
```suggestion
                            .load(aws_config::ecs::EcsCredentialsProvider::builder().build())
```

`aws_config::ecs::EcsCredentialsProvider` implements `ProvideCredentials` directly:
https://docs.rs/aws-config/latest/aws_config/ecs/struct.EcsCredentialsProvider.html#impl-ProvideCredentials-for-EcsCredentialsProvider

crates/symbolicator-service/src/services/download/s3.rs
```diff
@@ -155,93 +163,78 @@ impl S3Downloader {
...
-
-        let source = RemoteDif::from(file_source);
+        let client = self.get_s3_client(source_key).await;
+        let request = client
+            .get_object()
+            .bucket(bucket.clone())
+            .key(key.clone())
+            .send();
+
+        let source = RemoteDif::from(file_source.clone());
```
Swatinem, line 173: I believe if you clean up all the `println!`s below, you can avoid these `clone`s as well.

crates/symbolicator-service/src/services/download/s3.rs
```diff
@@ -264,6 +257,7 @@ impl S3Downloader {
     }
 }
 
+/*
```
Swatinem, line 260: I believe you can leave these tests in, they should be noops in case our credentials are not set.

Swatinem (commented):
This does look very good, thanks for sticking with it for so long!

The only missing thing would be to correctly map the errors. But I wouldn’t spend too much effort on that.
If we correctly flag connection timeout and notfound, we should be good. everything else can be an opaque S3Error.

Swatinem:
If the tests are giving troubles due to credentials, I can also pull this into a local branch manually and merge from there, that way it should have the right test credentials.

crates/symbolicator-sources/src/sources.rs
```diff
@@ -217,7 +248,7 @@ impl std::hash::Hash for S3SourceKey {
     fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
         self.access_key.hash(state);
         self.secret_key.hash(state);
-        self.region.name().hash(state);
+        self.region.to_string().hash(state);
```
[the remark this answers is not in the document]
josb: Fixed, thanks.

crates/symbolicator-sources/Cargo.toml
```diff
@@ -7,9 +7,10 @@ edition = "2021"
 license = "MIT"
 
 [dependencies]
+aws-types = { version = "0.51.0", features = ["hardcoded-credentials"] }
 glob = "0.3.0"
 lazy_static = "1.4.0"
-rusoto_core = "0.48.0"
+# rusoto_core = "0.48.0"
```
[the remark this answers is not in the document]
josb: Fixed!

crates/symbolicator-service/src/services/download/s3.rs
```diff
@@ -23,7 +25,7 @@ use symbolicator_sources::{
 use super::locations::SourceLocation;
 use super::{content_length_timeout, DownloadError, DownloadStatus, RemoteDif, RemoteDifUri};
 
-type ClientCache = moka::sync::Cache<Arc<S3SourceKey>, Arc<rusoto_s3::S3Client>>;
+type ClientCache = moka::sync::Cache<Arc<S3SourceKey>, Arc<Client>>;
```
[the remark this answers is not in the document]
josb: Done, thanks for getting me unstuck on this :-)

crates/symbolicator-service/src/services/download/s3.rs
```diff
@@ -72,14 +73,15 @@ pub struct S3Downloader {
...
-            .field("http_client", &format_args!("rusoto_core::HttpClient"))
             .field("connect_timeout", &self.connect_timeout)
             .field("streaming_timeout", &self.streaming_timeout)
             .finish()
     }
 }
 
-pub type S3Error = RusotoError<GetObjectError>;
+// It's one of these?
+// pub type S3Error = aws_sdk_s3::types::SdkError;
```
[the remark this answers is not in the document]
josb: Removed.

crates/symbolicator-service/src/services/download/s3.rs
```diff
@@ -88,56 +90,62 @@ impl S3Downloader {
...
-                    self.create_s3_client(provider, region)
-                }
-                AwsCredentialsProvider::Static => {
-                    let provider = rusoto_credential::StaticProvider::new_minimal(
-                        key.access_key.clone(),
-                        key.secret_key.clone(),
-                    );
-                    self.create_s3_client(provider, region)
-                }
+            tokio::runtime::Handle::current().block_on(async {
```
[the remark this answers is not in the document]
josb: Yes, thanks again, didn't know how to handle this better. Next time I'll be sure to look at the docs more carefully to see if an `async` variant is available.

crates/symbolicator-service/src/services/download/s3.rs
josb: Nice! Fixed, thanks.

crates/symbolicator-service/src/services/download/s3.rs
josb: Fixed by converting to clones to references.

crates/symbolicator-service/src/services/download/s3.rs
josb: I'll have to spend some time to convert the tests from rusoto to aws-sdk-rust, please stay tuned.

josb: The tests now compile.

crates/symbolicator-sources/src/sources.rs
```diff
@@ -129,39 +129,67 @@ pub struct FilesystemSourceConfig {
...
         {
-            FromStr::from_str(value).map_err(|e| E::custom(format!("region: {:?}", e)))
-        }
-
-        fn visit_seq<S>(self, seq: S) -> Result<rusoto_core::Region, S::Error>
-        where
-            S: serde::de::SeqAccess<'de>,
-        {
-            Deserialize::deserialize(serde::de::value::SeqAccessDeserializer::new(seq))
+            // aws-sdk-rust does not validate region strings
```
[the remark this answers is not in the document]
josb: Dropped in https://github.com/getsentry/symbolicator/pull/849/commits/da3eb8b1d9257d5020bfa3fcba46fb5f01594889.

josb:
At this point, as far as I know the major issues are:
* Custom region support; not sure how to go about this.
* `SdkError<GetObjectError>` handling in `download_source()`; I might be able to clean this up a bit but was hoping for some better guidance from https://github.com/awslabs/aws-sdk-rust/issues/191.
* The failing `no_permission` integration test, related to the previous item.

Anything else you can think of?

Swatinem:
> Custom region support; not sure how to go about this.

I believe its okay to be as lenient as possible. People will get NotFound / NoPermission if they provide an invalid region.

josb:
BTW by "custom region support" I mean this:
> In order to use a custom region for an S3 compatible service such as Ceph or minio, specify a tuple: ["custom-region-name", "http://minio-address/"].

My patch took away support for this because unlike [Rusoto](https://rusoto.github.io/rusoto/rusoto_core/region/enum.Region.html#variant.Custom), the Rust AWS SDK doesn't support a Custom Region variant.

Swatinem:
Hm, I don’t think we really support those. But I could be wrong. First time I read about it.

josb:
This [discussion](https://github.com/awslabs/aws-sdk-rust/discussions/367) seems relevant; the [answer](https://docs.aws.amazon.com/sdk-for-rust/latest/dg/localstack.html) given suggests that a custom endpoint resolver can be used, as I guess the region is ultimately used to synthesize the URI of the wanted service endpoint.

Swatinem:
Ahhh, now I understand what this is about. I should have read the comment more closely (which needs updating btw :-)

> This is a Visitor that forwards string types to rusoto_core::Region's `FromStr` impl and forwards tuples to rusoto_core::Region's `Deserialize` impl.

That enum variant indeed consists of name + endpoint:

https://docs.rs/rusoto_core/latest/rusoto_core/enum.Region.html#variant.Custom

I think this is where my confusion was coming from.

I believe in this case we can mimic that behavior and have a:

```rust
enum Region {
  Builtin(String),
  Custom {
    name: String,
    endpoint: String,
  }
}
```

Then on usage one would have to match to then create the appropriate endpoint.

crates/symbolicator/src/services/download/s3.rs
josb: I am keeping an eye on the seemingly-related https://github.com/awslabs/aws-sdk-rust/issues/629.

Swatinem:
Thanks @josb for the great work here!

I pulled in your changes and created a new PR: https://github.com/getsentry/symbolicator/pull/954

That merges the latest changes, adds back support for custom regions and tries to clean up the error handling a bit, though the error handling from the upstream SDK is truely horrific :-(

PR #853, opened 2022-07-15
deps: Update dependencies, esp minidump-processor

Swatinem:
This does a `cargo update`, and bumps the rust-minidump dependencies to 0.12.0 which should now support MIPS stackwalking

#skip-changelog


kamilogorek:
Is this fine if we skip this `impls`?

loewenheim (approved):

PR #861, opened 2022-08-04
fix: Update symbolic and bump CFI/SymCache

Swatinem:
This improves Win-x64 CFI generation, and fixes even more wrong SymCaches when dealing with inlinees.


Swatinem:
changelog bot, you are drunk!

Swatinem:
The changes to the integration test make it clear that we now clean up C-decorated names which we previously did not do.

Those were however cleaned up afterwards in sentry itself:

https://github.com/getsentry/sentry/blob/c6c07f1544a2581fb0ec8db48dc8bfef7fb8fc7e/tests/sentry/stacktraces/test_functions.py#L77-L78

Therefore, this will not have an effect on / break grouping. CC @mitsuhiko @jan-auer since both of you had concerns here

ashwoods (approved):

PR #863, opened 2022-08-05
feat: Pretty-print process-event output

Swatinem:
This adds a new `-p`/`-pretty` option that will pretty-print the crashing thread of a processed minidump.

#skip-changelog

### DIFF
--- a/crates/process-event/Cargo.toml
+++ b/crates/process-event/Cargo.toml
@@ -11,3 +11,4 @@ reqwest = { version = "0.11.0", features = ["blocking", "json", "multipart"] }
 serde = { version = "1.0.137", features = ["derive"] }
 serde_json = "1.0.81"
 structopt = "0.3.21"
+symbolic-common = { git = "https://github.com/getsentry/symbolic", branch = "fix/demangle-fixes" }
--- a/crates/process-event/src/main.rs
+++ b/crates/process-event/src/main.rs
@@ -1,5 +1,6 @@
 //! Tool to run Minidumps or Sentry Events through a local Symbolicator.
 
+use std::collections::HashMap;
 use std::fs::File;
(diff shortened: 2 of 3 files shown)


crates/process-event/src/main.rs
```diff
@@ -8,6 +9,12 @@ use ::reqwest::blocking::multipart;
 use reqwest::blocking as reqwest;
 use serde_json::{to_string, Map, Value};
 use structopt::StructOpt;
+use symbolic_common::split_path;
+
+#[path = "../../symbolicator/src/utils/hex.rs"]
```
Swatinem, line 14: https://github.com/rust-lang/rust-analyzer/issues/9173 :-(

crates/process-event/Cargo.toml
```diff
@@ -11,3 +11,4 @@ reqwest = { version = "0.11.0", features = ["blocking", "json", "multipart"] }
 serde = { version = "1.0.137", features = ["derive"] }
 serde_json = "1.0.81"
 structopt = "0.3.21"
+symbolic-common = { git = "https://github.com/getsentry/symbolic", branch = "fix/demangle-fixes" }
```
kamilogorek, line 14: All changes from this branch are already merged on master from what I see

kamilogorek (approved):

crates/process-event/Cargo.toml
Swatinem: Not quite. We are carrying that branch around for a way too long time already.

We can’t get rid of it quite yet, as that would regress some of our C++ demangling :-( But maybe we can switch from a symbolic branch to maybe a `Cargo.toml` `[patch]` section so we only pull in a specific commit/fix from `cpp_demangle`, and not a whole `symbolic` branch.

PR #866, opened 2022-08-08
ref: Avoid git dependencies in favor of cargo patch

Swatinem:
This uses a cargo patch section to replace very specific dependencies, instead of whole dependency trees.
The advantage here is that we can get rid of the log running demangle-branch of symbolic, and rather pull in
a targeted `cpp_demangle` dependency which is a transitive dependency of symbolic.

I also took the liberty to do the same for `reqwest`, which also gets rid of a duplication that previously happened
due to reqwest being pulled in as a transitive dependency from cargo through the sentry SDK.

This also means that we now depend on a *published* version of symbolic, which means we need to publish it more often ;-)

#skip-changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -206,7 +206,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "miniz_oxide",
- "object 0.29.0",
+ "object",
  "rustc-demangle",
 ]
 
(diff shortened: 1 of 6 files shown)


Swatinem:
This is blocked by actually publishing a new symbolic: https://github.com/getsentry/publish/issues/1278

loewenheim (approved):
+1000

PR #868, opened 2022-08-10
experiment: Enable profiling with latest Sentry SDK

Swatinem:


crates/symbolicator/Cargo.toml
```diff
@@ -36,7 +36,7 @@ reqwest = { git = "https://github.com/jan-auer/reqwest", tag = "v0.11.0", featur
 rusoto_core = "0.48.0"
 rusoto_credential = "0.48.0"
 rusoto_s3 = "0.48.0"
-sentry = { version = "0.27.0", features = ["anyhow", "debug-images", "log", "tracing"] }
+sentry = { git = "https://github.com/getsentry/sentry-rust.git", branch = "master", features = ["anyhow", "debug-images", "log", "tracing"] }
```
viglia, line 39:
```suggestion
sentry = { git = "https://github.com/getsentry/sentry-rust.git", branch = "master", features = ["anyhow", "debug-images", "log", "tracing", "profiling"] }
```

I think you might need to add `profiling` to _features_

Swatinem: oh jeez, you are totally right ;-)

crates/symbolicator/Cargo.toml
```diff
@@ -70,3 +70,6 @@ warp = "0.3.0"
 
 [features]
 https = ["axum-server/tls-rustls"]
+
+[profile.release]
+debug = true
```
Swatinem, line 75: we should have this already in the global `Cargo.toml`

Dockerfile
```diff
@@ -30,9 +30,12 @@ COPY --from=symbolicator-planner /work/recipe.json recipe.json
 # Build only the dependencies identified in the `symbolicator-planner` image
 RUN cargo chef cook --release --features=${SYMBOLICATOR_FEATURES} --recipe-path recipe.json
 
+RUN rustup toolchain install nightly
+
 COPY . .
 RUN git update-index --skip-worktree $(git status | grep deleted | awk '{print $2}')
-RUN cargo build --release --features=${SYMBOLICATOR_FEATURES}
+RUN cargo +nightly run -Z build-std --target x86_64-unknown-linux-gnu
```
Swatinem, line 37: I believe these rustup & cargo invocations should be above the `RUN cargo chef` command.

The `cargo chef cook` builds all the dependencies independently from the main symbolicator for docker layer caching purposes. Those dependencies should also be built with the proper RUSTFLAGS and std.
Plus, the newly built std should also be part of the docker layer cache.

crates/symbolicator/src/cli.rs
```diff
@@ -78,6 +78,11 @@ pub fn execute() -> Result<()> {
         release: Some(env!("SYMBOLICATOR_RELEASE").into()),
         session_mode: sentry::SessionMode::Request,
         auto_session_tracking: false,
+        enable_profiling: true,
+        // profiles are sub-sampling transactions, which we only sample based on upstream requests.
+        // the sample rate we get from upstream is very low (< 20 transactions per minute, spread
+        // across ~30 somewhat instances), so we might as well go full blast here.
```
Swatinem, line 84: we only have ~11 instances by now, but that is just an implementation detail.

PR #870, opened 2022-08-17
fix: Switch to upstream gcp_auth

Swatinem:
Our patches for gcp_auth have since been merged so we can switch back to a published version.

This also fixes the `process-event` utility to specify the required `trust-dns` `reqwest` feature,
as our patches do not compile without that feature.

#skip-changelog


loewenheim (approved):

flub (approved):
We've had problems with gcp_auth upgrades before though, so probably best to roll out via canary first

PR #874, opened 2022-08-31
ref: Replace custom shared-cache Queue with barbados

Swatinem:
This does simplify the shared-cache queue code a little bit.
Next step would be to figure out how to also replace the lazy cache refresh feature with this code.

#skip-changelog


Swatinem:
Some feedback on `barbados` API design:

- All the types in symbolicator want this to be `Debug`.
- The shared-cache wants to maintain a `gauge` metric for the number of currently running tasks.
- `Cacher::create_channel` has its own custom logic with its own channels, spawning, sentry transactions, etc.
- ^ Threading (pun intended) barbados in there might be a bit of work. Symbolicator is also careful to avoid any kind of extra work when it hit the max number of lazy computations.

PR #884, opened 2022-09-23
feat: Add Portable PDB support to symsorter

Swatinem:
This allows sorting portable PDB files into a unified directory layout. The files are using a `debuginfo` suffix.

Depends on https://github.com/getsentry/symbolic/pull/689


Swatinem:
this is not needed anymore now that `PortablePdb` is a proper `Object`. The `Archive` parsing code in symsorter can handle these completely transparently now.

PR #887, opened 2022-10-04
ref: Clean up empty cache directories

Swatinem:
Thus far, the cleanup process never removed empty cache directories. For one, the worklist based approach cleaning the parent before the children, did not make this possible in the first place, but also because normal operations would have errored out on failing FS operations, for example when a directory or file could not be created in a non-existent directory.

This changes the cleanup process to be depth-first, cleaning up empty directories as it goes. It also makes the writing side of things extra defensive and retries FS operations in the unlikely case that a directory is being removed while operating in it.

### DIFF
--- a/crates/symbolicator/src/cache.rs
+++ b/crates/symbolicator/src/cache.rs
@@ -1,7 +1,7 @@
 //! Core logic for cache files. Used by `crate::services::common::cache`.
 
 use core::fmt;
-use std::fs::{read_dir, remove_file};
+use std::fs::{read_dir, remove_dir, remove_file};
 use std::io::{self, Read, SeekFrom};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::AtomicIsize;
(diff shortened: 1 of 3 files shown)


crates/symbolicator/src/cache.rs
```diff
@@ -421,8 +442,38 @@ impl Cache {
...
+                let mut retries = 0;
+                loop {
+                    retries += 1;
+
+                    if let Err(e) = std::fs::create_dir_all(path) {
+                        sentry::with_scope(
+                            |scope| scope.set_extra("path", path.display().to_string().into()),
+                            || tracing::error!("Failed to create cache directory: {:?}", e),
+                        );
+                        if retries >= MAX_RETRIES {
```
kamilogorek, line 457: This will make this whole function to be called only twice, not thrice. Which is technically incorrect, as it should "retry" N times (2), so it should be called 3 times.

`if retries > MAX_RETRIES` will make it be called 3 times.

crates/symbolicator/src/cache.rs
```diff
@@ -421,8 +442,38 @@ impl Cache {
...
+                    }
+
+                    match tempfile::Builder::new().prefix("tmp").tempfile_in(path) {
+                        Ok(temp_file) => return Ok(temp_file),
+                        Err(e) => {
+                            sentry::with_scope(
+                                |scope| scope.set_extra("path", path.display().to_string().into()),
+                                || tracing::error!("Failed to create cache file: {:?}", e),
+                            );
+                            if retries >= MAX_RETRIES {
```
kamilogorek, line 470: Same here.

crates/symbolicator/src/cache.rs
```diff
@@ -285,48 +285,69 @@ impl Cache {
...
-                    tracing::warn!("Directory not found");
-                    return Ok(());
-                }
-            };
+        Ok(())
+    }
+
+    /// Cleans up the directory recursively, returning `true` if the directory is left empty after cleanup.
+    fn cleanup_directory_recursive(&self, directory: &Path) -> Result<bool> {
+        let entries = match catch_not_found(|| read_dir(directory))? {
```
kamilogorek, line 299: `s/catch/cache`?

crates/symbolicator/src/cache.rs
```diff
@@ -285,48 +285,69 @@ impl Cache {
...
-                    directories.push(path.to_owned());
-                } else if let Err(e) = self.try_cleanup_path(&path) {
-                    sentry::with_scope(
-                        |scope| scope.set_extra("path", path.display().to_string().into()),
-                        || tracing::error!("Failed to clean cache file: {:?}", e),
-                    );
+        let mut is_empty = true;
+        for entry in entries {
+            let entry = entry?;
+            let path = entry.path();
```
kamilogorek, line 310:
```suggestion
            let path = entry?.path();
```

Does this syntax work?

crates/symbolicator/src/services/cacher.rs
```diff
@@ -379,8 +379,42 @@ impl<T: CacheItemRequest> Cacher<T> {
...
+                let mut retries = 0;
+                loop {
+                    retries += 1;
+
+                    if let Err(e) = fs::create_dir_all(parent).await {
+                        sentry::with_scope(
+                            |scope| scope.set_extra("path", parent.display().to_string().into()),
+                            || tracing::error!("Failed to create cache directory: {:?}", e),
+                        );
+                        if retries >= MAX_RETRIES {
```
kamilogorek, line 395: Same as above in both places.

kamilogorek (changes requested):

crates/symbolicator/src/cache.rs
Swatinem: This is correct, it catches not found errors, turning them into `None`.

crates/symbolicator/src/cache.rs
Swatinem: agreed. naming this is hard :-D

kamilogorek (approved):

PR #888, opened 2022-10-19
fix: Properly mask Portable PDB Age for symstore/SSQP lookups

Swatinem:
Portable PDB lookups via the SSQP scheme say that the `Age` portion of the ID needs to be replaced with u32::MAX.

See https://github.com/dotnet/symstore/blob/main/docs/specs/SSQP_Key_Conventions.md#portable-pdb-signature

This allows us to properly look up Portable PDB files on the upstream microsoft symbol server.
For example, http://msdl.microsoft.com/download/symbols/microsoft.aspnetcore.routing.pdb/fbd5dfdfa5a44e09a75862fcbb8c914aFFFFFFFF/microsoft.aspnetcore.routing.pdb might be such a valid file which matches a stack frame from https://sentry.io/organizations/sentry-sdks/issues/2442781715/events/a4f4c59a7d1a4b06bd98c0d1dcfa7e8c/?project=5428537


local.example.yml
```diff
@@ -38,7 +38,7 @@ sources:
     type: http
     url: https://msdl.microsoft.com/download/symbols/
     layout: { type: "symstore" }
```
Swatinem, line 40: tbh, I have no idea why this is `symstore` and not `ssqp`. The only difference seems to be the casing, and the upstream microsoft server seems to be case insensitive anyway as far as I have tried.

loewenheim (approved):

PR #890, opened 2022-10-20
ref: Extract symbolicator-sources as its own workspace crate

Swatinem:
Moves the paths and sourceconfig related types to its own workspace crate.

The split might not be perfect, and I’m open to suggestions.

#skip-changelog


loewenheim (approved):
Very nice!

PR #891, opened 2022-10-21
feat: Create a new Caching infrastructure

Swatinem:


Swatinem:
We had some brainstorming around the types / errors that we want to return *at the very end* when requesting symcaches, cficaches, etc:

```
fetch_symcache(...) -> (CacheEntry<SymCache>, Candidates)

// moka =>
struct InternalCacheEntry<T> {
    entry: CacheEntry<T>,
    ttl: Instant,
}

enum CacheEntry<T> {
    AllGood(T), // => symcache => SymCache, object_meta => ObjectFeatures, cficache => SymbolFile
    NotFound,
    PermissionDenied(String), // => whatever the server returned
    Timeout(Duration), // => should we return the duration after which we timed out?
    DownloadError(String), // => connection lost, dns resolution, or 5xx server errors
    Malformed(String), // => unsupported object file, symcache conversion failure
    InternalError, // => unexpected errors like: io::Error, serde_error, symcache parse error,
}
```

PR #896, opened 2022-10-25
fix: Use cargo-chef for Docker Layer Caching

Swatinem:
Our manual approach to Docker layer caching is annoying and fragile. Cargo chef solves exactly that usecase.

How it works and needs to be used is documented on https://github.com/LukeMathWalker/cargo-chef. We rather `cargo install` it than use the pre-published docker images.

#skip-changelog


kamilogorek (approved):
Looks reasonable

PR #926, opened 2022-12-01
ref: Test cached Error Statuses

Swatinem:
This adds a second iteration to the download-errors test, making sure that round trips through the on-disk cache yield the same statuses.

This also tightens up the retry logic, as we should try things 3 times, not an additional 3 times. It also adds a couple of notes for where the retry logic does not match the expectation. Why are we retrying permission errors that most likely will not change after retry, while we are *not* retrying timeouts which we probably should, as an intermittent timeout can be gone on the next try.

### DIFF
--- a/crates/symbolicator-service/src/services/objects/data_cache.rs
+++ b/crates/symbolicator-service/src/services/objects/data_cache.rs
@@ -420,7 +420,7 @@ mod tests {
                 "failed to download: 500 Internal Server Error"
             ))
         );
-        assert_eq!(server.accesses(), 1 + 3); // 1 initial attempt + 3 retries
+        assert_eq!(server.accesses(), 3); // up to 3 tries on failure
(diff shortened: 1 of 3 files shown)


Swatinem:
the timeout mystery is this: we have ~3 timeouts:

1. connection timeouts, these we do retry as they are happening "inside the download"
2. adaptive download timeouts, these we also do retry as they are happening "inside the download"
3. we have a global download timeout which we do *not* retry as it is wrapping the "inner download"

In our tests we hit the global timeout. We can probably extend the tests to cover a server that never accepts a socket request to cover 1. and maybe cover 2. by adjusting the adaptive timeouts.

Also, I’m not sure if we should be retrying the adaptive timeout? Will the server magically stream files faster if we connect a second time?

Maybe we should also think about flagging *what* timed out. And decide in more detail what exactly we want to retry.

kamilogorek (approved):

PR #928, opened 2022-12-01
Add image workflow

mattgauntseo-sentry:
#skip-changelog

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -2,7 +2,9 @@ name: image
 on:
   pull_request:
   push:
-    branches: [master]
+    branches:
+      - master
+      - release/**
 jobs:
   build-image:
     strategy:
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -11,7 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     env:
       # IMG_CACHE: ghcr.io/getsentry/symbolicator:${{ matrix.arch }}-latest
-      # IMG_VERSIONED: ghcr.io/getsentry/symbolicator:${{ matrix.arch }}-${{ github.sha }}
+      IMG_VERSIONED: ghcr.io/getsentry/symbolicator:${{ matrix.arch }}-${{ github.sha }}
+      NIGHTLY_IMG_CACHE: ghcr.io/getsentry/symbolicator:${{ matrix.arch }}-nightly
       BUILDER_IMG_CACHE: ghcr.io/getsentry/symbolicator:${{ matrix.arch }}-builder
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/image.yml
```diff
@@ -0,0 +1,54 @@
+name: image
+on:
+  pull_request:
+  push:
+    branches: [master]
```
Swatinem, line 5:
```suggestion
    branches:
      - master
      - release/**
```

This should match the branches that are built for deployment

.github/workflows/image.yml
```diff
@@ -0,0 +1,54 @@
...
+    needs: [build-image]
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          set -euxo pipefail
+          docker login --username '${{ github.actor }}' --password '${{ secrets.GITHUB_TOKEN }}' ghcr.io
+          docker manifest create ghcr.io/getsentry/symbolicator:latest \
+            ghcr.io/getsentry/symbolicator:arm64-latest \
+            ghcr.io/getsentry/symbolicator:amd64-latest
+          docker manifest push ghcr.io/getsentry/symbolicator:latest
```
asottile-sentry, line 54: this will fail with only one image -- and we can't use manifest-based images if we only have 1 architecture or it will break M1

.github/workflows/image.yml
```diff
@@ -0,0 +1,75 @@
...
+              --platform linux/${{ matrix.arch }} \
+              --tag "$BUILDER_IMG_CACHE" \
+              --target symbolicator-build \
+              .
+      - name: push builder img
+        run: |
+          set -euxo pipefail
+          docker login --username '${{ github.actor }}' --password '${{ secrets.GITHUB_TOKEN }}' ghcr.io
+          docker push "$BUILDER_IMG_CACHE"
+        if: github.event_name != 'pull_request'
```
asottile-sentry, line 46: it should probably push everything at the end only when both builds have succeeded

Swatinem (approved):

mattgauntseo-sentry:
Looks like /data exists in the cloud build because it's using a cached image since https://github.com/getsentry/symbolicator/commit/26e04c78b72316eab4506655ef6578f176f2fe0d

From: https://console.cloud.google.com/cloud-build/builds%3Bregion=global/f179ec95-7638-462c-a4e1-b725ee4efdf4?project=sentryio
```
Step #3: Step 22/28 : VOLUME ["/data"]
Step #3:  ---> Using cache
Step #3:  ---> 81e5f90c0557
```

Swatinem:
I don’t know enough about docker tbh. I believe this is fine, as you can just mount over that dir when the docker image is actually *used*?

PR #394, opened 2021-02-23
test: Move remaining tests to Rust

jan-auer:
Moves the remaining tests from `test_basic.py` to Rust.

- [ ] test_basic_windows
- [x] test_no_sources
- [x] test_unknown_field
- [ ] test_lookup_deduplication
- [ ] test_sources_filetypes
- [ ] test_unknown_source_config
- [ ] test_timeouts
- [ ] test_unreachable_bucket
- [ ] test_malformed_objects
- [ ] test_path_patterns
- [ ] test_redirects
- [ ] test_reserved_ip_addresses
- [ ] test_no_dif_candidates

PR #959, opened 2023-01-03
ref: Migrate a couple python tests to rust

Swatinem:
Admittedly, the python tests are concise, but also hard to understand.

This does introduce one behavior change: Most tests are run against the `service` now, bypassing the http server layer.
The tests that are actually asserting something about the server protocol are still kept in the `symbolicator` tests.

Supercedes #394

#skip-changelog


crates/symbolicator-service/tests/integration/e2e.rs
```diff
@@ -0,0 +1,328 @@
...
+        "s3://symbolicator-test/502F/C0A5/1EC1/3E47/9998/684FA139DCA7.app"
+    );
+    assert_eq!(
+        candidates[5].download,
+        ObjectDownloadInfo::NoPerm { details: "".into() }
+    );
+}
+
+#[tokio::test]
+async fn test_reserved_ip_addresses() {
```
kamilogorek, line 220: That's a long test 😅

kamilogorek (approved):

PR #982, opened 2023-01-19
fix: Use unique Cache files for symcaches with markers

Swatinem:
This gives symcaches a unique cache key depending on the `secondary_sources` and the same cache file is not being reused with different sets of `secondary_sources`.

It also paves the way to avoid appending markers to files and thus getting rid of the `should_load` check in the future.

This still does not solve the edge-case when any of the `secondary_sources` itself changes. But that is the same edge-case we currently have if files on external symbol sources are overwritten. As long as we have that file cached, we will never try to re-download that file.
Cache files are considered to be immutable.

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/services/symcaches/markers.rs
+++ b/crates/symbolicator-service/src/services/symcaches/markers.rs
@@ -1,5 +1,7 @@
+use std::fmt::Write;
 use std::io::SeekFrom;
 
+use crate::services::cacher::CacheKey;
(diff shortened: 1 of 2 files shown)


loewenheim (approved):

Swatinem:
We should be cautious when we roll this out, as this change here means that all the symcaches that *are* using some kind of secondary source won’t be found in the cache anymore.

Not sure if we should go forward with this right now, or rather make progress on #983, especially related to forward compatibility.

PR #989, opened 2023-01-24
ref: Make `SharedCache` store a `ByteView`

Swatinem:
Instead of using a `tokio::File`, this now uses a `ByteView`, which simplifies things and also makes `lookup_local_cache` fully sync.

#skip-changelog


loewenheim (approved):
Looks much nicer this way!

PR #990, opened 2023-01-25
ref: Remove `SharedCacheKey`

Swatinem:
Removes the `SharedCacheKey` in favor of just using a `String` directly, though I’m not entirely sure that is the right approach yet.

Though the usage of `CacheName` for diagnostics is a bit of a leaky abstraction.


Swatinem:
This was done in #1033 instead

PR #996, opened 2023-01-29
feat: Support PortablePDB embedded sources

vaind:
Followup from https://github.com/getsentry/symbolic/pull/734 now that symbolic got released,
this enables source context support for users uploading Portable PDBs built with `<EmbedAllSources>true</EmbedAllSources>` project property.

- chore: update symbolic to v11.0.0
- test: portable PDB with embedded sources
- feat: support portable PDB embedded sources


Cargo.lock
```diff
@@ -2892,7 +2892,7 @@ dependencies = [
  "serde",
  "serde_json",
  "structopt",
- "symbolic-common",
+ "symbolic-common 10.2.1",
```
Swatinem, line 2895: pls also bump the version in the `process-event` utility :-)

crates/symbolicator-service/tests/integration/snapshots/integration__symbolication__wasm_payload.snap
```diff
@@ -49,4 +49,8 @@ modules:
         location: "http://localhost:<port>/symbols/bd/a18fd85d4a4eb893022d6bfad846b1.src.zip"
         download:
           status: notfound
+      - source: local
+        location: "http://localhost:<port>/symbols/demo.wasm/BDA18FD85D4A4EB893022D6BFAD846B10/demo.wasm"
```
Swatinem, line 53: I’m wondering why a completely unrelated filetype now has more candidates? 🤔

vaind: yeah that was weird but my assumption is that it's related to symbolic update

Cargo.lock
vaind: done

crates/symbolicator-service/tests/integration/snapshots/integration__symbolication__wasm_payload.snap
Swatinem: possible. trying yet another file name pattern does definitely not hurt, though I would still love to find out why this suddenly happens.

Swatinem (approved):

PR #1010, opened 2023-02-02
ref: Use `moka` for `Cacher` implementation, 2nd try

Swatinem:
Reopens https://github.com/getsentry/symbolicator/pull/979:

Replaces the current request coalescing solution with `moka`. This simplifies the channel creation and deduplication logic, as `moka` does that already. This also gives us in-memory caches right now, even though they are not hooked up to configuration yet.

Just opening this for now that its not lost


Swatinem:
Seems to be stable so far, lets try to roll this out tomorrow.

loewenheim (approved):

PR #1016, opened 2023-02-06
build: move DIF uploading to buildtime

joshuarli:
This moves DIF uploading to build time so that gocd doesn't need to `docker pull` the whole image and extract those files to upload. I opted for this instead of trying to make `docker pull` work on gocd - the agents run unprivledged so it was tricky to try and set up rootless dockerd (dockerd needs to be running for pulling to work).

As a result we'll be uploading on every push to master, but given the slow rate of commits I think this is a worthwhile tradeoff to keep gocd leaner and faster.

edit: Just realized there's GHA doing this as well, but for ghcr.io releases - will add a switch to disable DIF uploading if SENTRY_AUTH_TOKEN isn't present.

#skip-changelog


Dockerfile
```diff
@@ -68,7 +66,6 @@ RUN mkdir -p /etc/symbolicator /data && \
 EXPOSE 3021
 
 COPY --from=symbolicator-build /usr/local/bin/symbolicator /bin
-COPY --from=symbolicator-build /opt/symbolicator-debug.zip /opt/symbolicator.src.zip /opt/
```
Swatinem, line 71: this might be relevant for self-hosted users. TBH, I have 0 insight into self-hosted works right now, or if it even sets up something equivalent to S4S. If so, where will self-hosted users ever get the symbolicator debug files?

Swatinem (commented):
someone who knows self-hosted should chime in.
otherwise lgtm as soon as CI agrees

Swatinem:
```
#23 [symbolicator-build 9/9] RUN scripts/build-and-upload-difs.sh
#23 0.175 /bin/sh: 1: scripts/build-and-upload-difs.sh: not found
#23 ERROR: executor failed running [/bin/sh -c scripts/build-and-upload-difs.sh]: exit code: 127
```

Ah yes, I believe you copy that file over, should be easy to fix.

Swatinem:
Another thing we would love to have is separate canary deployments. We often want to run longer running experiments in canary only.

Dockerfile
joshuarli: Maybe Github release artifacts? I see `craft.yml` references stuff like `symbolicator-Linux-x86_64-debug.zip` which is built by `.github/workflows/build.yml`

joshuarli: cc @chadwhitacre

chadwhitacre: @hubertdeng123 Can you take a look here?

hubertdeng123: I'm not too familiar with this either but I think @joshuarli is probably right. Is might be good to double check that self-hosted works with the new workflow

joshuarli:
I'm gonna shelve this for now - realized last night that we need to get `docker pull` working elsewhere

PR #1023, opened 2023-02-07
cd: separate canary deployments

joshuarli:
https://github.com/getsentry/symbolicator/pull/1016#issuecomment-1421272942

### DIFF
--- a/gocd/pipelines/symbolicator-canary.yaml
+++ b/gocd/pipelines/symbolicator-canary.yaml
@@ -3,7 +3,9 @@
 # - https://www.notion.so/sentry/GoCD-New-Service-Quickstart-6d8db7a6964049b3b0e78b8a4b52e25d
 format_version: 10
 pipelines:
-    deploy-symbolicator:
+    # Canary deploys are completely separate from production
+    # because it's desirable to leave longer running experiments in canary.
+    deploy-symbolicator-canary:
         environment_variables:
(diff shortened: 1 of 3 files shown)


gocd/pipelines/symbolicator.yaml
```diff
@@ -42,43 +42,24 @@ pipelines:
                                     ${GO_REVISION_SYMBOLICATOR_REPO} \
                                     sentryio \
                                     "us.gcr.io/sentryio/symbolicator"
-
-            - deploy_canary:
+            # Deploys production and lpq symbolicators.
```
Swatinem, line 45: Does it still deploy canary too?
Right now we have 3 configs in freight:
- `canary` deploys just that
- `lpq` deploys just that
- `production` deploys all of them

That way, when we deploy to production we know everything is running the deployed version.

joshuarli: I'm confused - you said you wanted to leave long running experiments in canary, so I took that as prod should not deploy canary.

Do you want this?
- separate canary deployment pipeline
- separate lpq deployment pipeline
- prod deployment pipeline that covers everything (as it was before)

ashwoods: yes, canary is always an intermediate step, and our goal in this quarter is to adapt our processes to deploy master to canary automatically.

ashwoods (approved):

PR #1029, opened 2023-02-13
refactor: let get_context_lines() take RawFrame 

vaind:
#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/services/symbolication/module_lookup.rs
+++ b/crates/symbolicator-service/src/services/symbolication/module_lookup.rs
@@ -16,7 +16,7 @@ use crate::services::ppdb_caches::{
 use crate::services::symcaches::{FetchSymCache, OwnedSymCache, SymCacheActor};
 use crate::types::{
     AllObjectCandidates, CompleteObjectInfo, CompleteStacktrace, ObjectFeatures, ObjectFileStatus,
-    RawStacktrace, Scope,
+    RawFrame, RawStacktrace, Scope,
 };
 use crate::utils::addr::AddrMode;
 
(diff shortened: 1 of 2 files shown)


crates/symbolicator-service/src/services/symbolication/module_lookup.rs
```diff
@@ -370,18 +370,21 @@ impl ModuleLookup {
...
+        let abs_path = frame.abs_path.as_ref()?;
+        let lineno = frame.lineno? as usize;
+
+        let entry = self.get_module_by_addr(frame.instruction_addr.0, frame.addr_mode)?;
         let session = debug_sessions.get(&entry.module_index)?.as_ref()?;
         let source = session.source_by_path(abs_path).ok()??;
+        let source = match source {
+            SourceCode::Content(text) => text,
+            SourceCode::Url(_url) => todo!("IMPLEMENT ME"),
+        };
```
Swatinem, line 385: We have pretty much the same challenge with sourcemap processing:

We are pre-fetching "bundles" that in the best case include all the sources. However, we still need to fall back to load sources by-file in case one is not bundled.

I was thinking about doing this in two stages:

1. collect all the "modules" into a Set
2. load the source bundles for all the modules in the Set
3. do a first pass applying source context. for each file that is not readily available, collect it into a second Set
4. load all the individual source files for the second Set
5. apply the source context for all the frames that were missing in 3

From an operations pov, this is blocked right now by solving https://github.com/getsentry/symbolicator/issues/1015 first. We can surely build all the logic, but would be hesitant to roll that out without some form of rate limiting / blocking that protects us from bad external servers, as we would do a ton more requests as we did before.

Swatinem (approved):
IMO this is a nice refactor on its own, might be worth landing without the actual code matching on the source code.

crates/symbolicator-service/src/services/symbolication/module_lookup.rs
vaind: pretty much what I'm doing (not pushed at the moment), thanks for confirmation. Also, the big open question for me is caching - but wanted to do it after finishing the PoC.

Swatinem: nice! Yes caching this and sourcemap file downloads is a very similar topic.

Instead of something like the `objects` cache that does some transforms in itself, and the "derived" caches, we would want one just for arbitrary file downloads I believe.

PR #1030, opened 2023-02-13
test(sourcemap): Add testing infra and few initial integration tests

kamilogorek:
There are currently 5 tests. 2 passing, and 3 failings (ignored, so we can merge this PR anyway).
Each failing test has a note regarding why it fails.

#skip-changelog


loewenheim (approved):

crates/symbolicator-service/tests/integration/sourcemap.rs
```diff
@@ -0,0 +1,300 @@
...
+#[tokio::test]
+async fn test_sourcemap_embedded_source_expansion() {
+    let (symbolication, _) = setup_service(|_| ());
+    let srv = symbolicator_test::sourcemap_server("05_sourcemap_embedded_source_expansion");
+
+    let stacktraces: Vec<JsProcessingRawStacktrace> = serde_json::from_str(
+        r#"[{
+            "frames": [
+                {
+                    "function": "function: \"HTMLDocument.<anonymous>\"",
```
Swatinem, line 163: I don’t think thats intentional? :-D

tests/fixtures/sourcemaps/01_sourcemap_expansion/test.min.js
```diff
@@ -0,0 +1,2 @@
+var makeAFailure=function(){function n(n){}function e(n){throw new Error("failed!")}function r(r){var i=null;if(r.failed){i=e}else{i=n}i(r)}function i(){var n={failed:true,value:42};r(n)}return i}();
+//# sourceMappingURL=test.min.js.map
```
Swatinem, line 2: Hm, I’m actually surprised this works. Does it really allow spaces in between `//#sourceMappingURL`?

Swatinem (approved):

crates/symbolicator-service/tests/integration/sourcemap.rs
kamilogorek: It is. It will test the error case later on, that's why I copy the payloads literally from monolith :p

tests/fixtures/sourcemaps/01_sourcemap_expansion/test.min.js
kamilogorek: It's mandatory :) - https://github.com/getsentry/rust-sourcemap/blob/1ba39818c2e28e8c46d284ee110df86680481584/src/detector.rs#L76

PR #1034, opened 2023-02-13
feat: source context links support

vaind:
Part of https://github.com/getsentry/symbolic/issues/735

TODO:
* [ ] implement caching & download properly


Swatinem:
https://github.com/getsentry/symbolicator/pull/1057 now provides a cache for downloaded source files. It might be worth adding a simple wrapper around that as well that accepts an arbitrary `Url` to download from. I might add that as a followup, then this should be unblocked.

vaind:
replaced by #1103

PR #1035, opened 2023-02-13
ci(gocd): deploy to canary before prod

mattgauntseo-sentry:
This change adds a manual step in the prod deployment flow to deploy to canary before deploying to prod. 

This also simplifies the production deployment step to rely on the `deploy_to_production=true` which I believe all deployments have: https://github.com/getsentry/ops/blob/master/k8s/services/symbolicator/_values.yaml

Screenshot 2023-02-13 at 2 23 06 PM

This doesn't remove the canary or lpq workflows as this sounds like they may still be helpful, however once we create a good rollback process (i.e. nicer than the gocd + commit sha) we can revisit if this is better than separate workflows. The main reason this may be better is that if you deploy to canary for a long running test (let's say a week or two), you'd run this pipeline, stop after the canary rollout and anyone wanting to do a new deploy will see the canary deployment in the dashboard as an incomplete/in-progress deployment and will hopefully check before overwriting these changes.

Screenshot 2023-02-13 at 2 27 28 PM

@joshuarli Looking at the ops repo, it looks like deploy_if_lpq may not be defined, do you know if thats due to drift from the ops repo?

#skip-changelog


joshuarli (commented):
re: `deploy_if_lpq`, that's my bad. It doesn't exist and I assumed it did. So I think the symbolicator-lpq pipeline won't work unless it's added to ops. Although, the _only_ lpq deployment 5 months ago seems to have worked - https://freight.getsentry.net/deploys/symbolicator/lpq/1.

Was there a prompt for this overall change or is it just something you thought about @mattgauntseo-sentry? I'm realizing now that I'm still unclear on what symbolicator owners want. Rereading  https://github.com/getsentry/symbolicator/pull/1023#discussion_r1101128108, I think we want to actually autodeploy canary. In which case I think your change is good, and the separate canary-only deployment pipeline is redundant and should be removed. 

Also given that LPQ has only been explicitly deployed once - not sure the utillity here of a separate pipeline for it. May as well remove the (currently broken) lpq pipeline, and just deploy lpq when promoting canary to production. @Swatinem thoughts?

mattgauntseo-sentry:
Prompt is largely one of best practice we are trying to promote with canary in general, deploy new release to a small portion of users before promoting to production.

It seems like symbolicator uses canary for some long term tests BUT it's still deployed when prod is deploy (it has the deploy_if_production=true label), so making it part of the prod deployment is not much of a change beyond encouraging folks to check canary deployed and is working as expected before pushing to all of prod.

I could easily by missing something though, so part of this PR was to start a conversation. If it's easier to chat @Swatinem let mw know and I'll find time for us to discuss.

Swatinem:
I will summarize a conversation that I had with @mattgauntseo-sentry:

- We use canary for two different things:
- First, in the normal deploy to production pipeline. We deploy canary and then let it sit for some time. Depending on the changes we deploy, we let it sit for a few days even. We then decide if we want to roll back, or roll out to production.
- We also use it for "experiments", which means deploying and validating some ideas/code. One long running experiment was to dogfood profiling support in the Rust SDK, sadly with limited success :-( https://getsentry.atlassian.net/browse/OPS-2894 BTW is still open to change the reporting back to S4S :-) With these experiments we want to always roll those back. So this is really not too different from the point above.
- For the PLQ, you are right that we only used that once so far, but it was critical. Here is the whole story:

A very specific customers debug files were triggering OOMs in symbolicator. (These files are so huge they even ran into 32bit overflows in the file format: https://github.com/getsentry/symbolic/issues/745) Because they were hitting our infra very hard, we moved them to the LPQ via a killswitch. Then to validate a hotfix for the issue (https://github.com/getsentry/symbolic/pull/683), we wanted to deploy that to LPQ only (which was *only* serving that single customer which already had degraded symbolication) to validate the hotfix before rolling that out.

This is admittedly a very niche use case, and I hope we won’t be in the situation again to actually need it, though it was useful in that case :-)

joshuarli:
Ok, so:

- need to add `deploy_if_lpq` to ops and restore the deploy lpq only pipeline
- canary deployment stage is not autodeployed for now?

If that's desired @Swatinem then matt's PR is good as-is.

mattgauntseo-sentry:
I believe this PR is the right approach for the steps above.

It'll make deploying to canary a required part of the production deployment and if you want a deployment to canary with zero risk of rolling out to prod, the canary only pipeline is still there.

Swatinem (approved):

PR #1043, opened 2023-02-16
ref: Clean up deprecated Cache features

Swatinem:
Now that caches have been refreshed using the new stable cache keys, we can remove a couple of legacy code:
- Code that deals with legacy cache markers.
- Code related to legacy cache paths.
- Fallback to older cache versions that were still using legacy keys.
- The `should_load` machinery.
- SymCache markers, as they are not needed anymore.

fixes https://github.com/getsentry/symbolicator/issues/983, fixes https://github.com/getsentry/symbolicator/issues/965

Lets wait with landing this until the lazy cache refresh has progressed a bit more, but it should probably be good to land by tomorrow.


loewenheim (approved):
That's some serious cleanup :)

PR #1046, opened 2023-02-17
ref: Add trim_context_line for sourcemap processing

kamilogorek:
Translated from https://github.com/getsentry/sentry/blob/abdfcfdaf0f2e16db06b5348bb80a0f4e268455e/src/sentry/lang/javascript/processor.py#L89-L121
and https://github.com/getsentry/sentry/blob/abdfcfdaf0f2e16db06b5348bb80a0f4e268455e/tests/sentry/lang/javascript/test_processor.py#L1192-L1216

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/services/symbolication/sourcemap.rs
+++ b/crates/symbolicator-service/src/services/symbolication/sourcemap.rs
@@ -172,9 +172,57 @@ fn fold_function_name(function_name: &str) -> String {
     format!("{folded}.{tail}")
 }
 
+/// Trims a line down to a goal of 140 characters, with a little wiggle room to be sensible
+/// and tries to trim around the given `column`. So it tries to extract 60 characters
+/// before and after the provided `column` and yield a better context.
+fn trim_context_line(line: &str, column: Option<usize>) -> String {
+    use std::cmp::{max, min};
+
+    let mut line = line.trim_end().to_string();
+    let len = line.len();
+
+    if len <= 150 {
+        return line;
+    }
+
+    let col = column.map(|c| min(c, len)).unwrap_or_default();
+
+    let mut start = max(col.saturating_sub(60), 0);
+    // Round down if it brings us close to the edge.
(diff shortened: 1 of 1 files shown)


crates/symbolicator-service/src/services/symbolication/sourcemap.rs
```diff
@@ -172,9 +172,57 @@ fn fold_function_name(function_name: &str) -> String {
     format!("{folded}.{tail}")
 }
 
+/// Trims a line down to a goal of 140 characters, with a little wiggle room to be sensible
+/// and tries to trim around the given `column`. So it tries to extract 60 characters
+/// before and after the provided `column` and yield a better context.
```
Swatinem, line 177: According to the code its actually 60 before, and 80 after the provided `column`.
Are we okay with all these values hardcoded?

crates/symbolicator-service/src/services/symbolication/sourcemap.rs
```diff
@@ -172,9 +172,57 @@ fn fold_function_name(function_name: &str) -> String {
...
+
+    let mut end = min(start + 140, len);
+    // Round up to the end if it's close.
+    if end > len.saturating_sub(5) {
+        end = len;
+    }
+
+    // If we are bumped all the way to the end, make sure we still get a full 140 chars in the line.
+    if end == len {
+        start = max(end.saturating_sub(140), 0);
```
Swatinem, line 204:
```suggestion
        start = end.saturating_sub(140);
```
Same here, and it will indeed saturate to `0`.

crates/symbolicator-service/src/services/symbolication/sourcemap.rs
kamilogorek: For now it's perfectly fine, especially that if we ever change it, it'd invalidate all our visual snapshot tests.

Swatinem (approved):

PR #1050, opened 2023-02-20
ref: Unify source context application and handle lookups from source files

kamilogorek:
This changes what files we store as a part of `SourcemapService`. We have a separate cache for artifacts and a separate cache for sourcemap caches.

`SourcemapService` now delegates to a new `SourceMapLookup` which is created for each event and is responsible for handling the state of all caches needed for a single processing task.

It also introduces new methods that allow fetching artifacts on demand, as this is a requirement for cases, where sourcemaps contain a reference to the source file, but do not contain `sourceContents` itself, thus forcing us to try and fetch the file from the server.

Naming and API/usage still lefts a lot to be desired, but it's a good start and it makes all current integration tests pass.

#skip-changelog


crates/symbolicator-service/src/services/sourcemap.rs
```diff
@@ -1,338 +1,42 @@
...
-    }
-
-    pub async fn fetch_artifact(
-        &self,
-        source: Arc<SentrySourceConfig>,
-        file_id: SentryFileId,
-    ) -> CacheEntry<ByteView<'static>> {
-        let mut temp_file = NamedTempFile::new().unwrap();
-        fetch_file(
+    pub fn create_sourcemap_lookup(&self, source: Arc<SentrySourceConfig>) -> SourceMapLookup {
```
Swatinem, line 34: NOTE: I wanted to create a per-request (with internal `Scope` / `sources`) for native requests as well.

crates/symbolicator-service/src/services/symbolication/sourcemap.rs
```diff
@@ -85,44 +123,96 @@ fn js_processing_symbolicate_frame(
...
+    result.raw.abs_path = token.file_name().unwrap_or("<unknown>").to_string();
+    result.raw.lineno = Some(token.line().saturating_add(1));
+    result.raw.colno = Some(token.column().saturating_add(1));
 
     if let Some(file) = token.file() {
-        result.raw.filename = file.name().map(|name| name.to_string());
+        result.raw.filename = file.name().map(|f| f.to_string());
+
+        if file.source().is_some() {
+            apply_source_context_from_sourcemap_cache(&mut result, file).await;
```
Swatinem, line 135: as a followup, we should maybe batch these together as a separate step.

crates/symbolicator-service/src/services/symbolication/sourcemap.rs
```diff
@@ -85,44 +123,96 @@ fn js_processing_symbolicate_frame(
...
+    let post_line = current_line.saturating_add(5);
+
+    let mut lines: Vec<_> = source.lines().map(|l| l.unwrap_or_default()).collect();
+
+    // `BufRead::lines` doesn't include trailing new-lines, but we do want one if it was there.
+    if source.ends_with("\n".as_bytes()) {
+        lines.push("\n".to_string());
+    }
+
+    dbg!(current_line);
```
Swatinem, line 202:
```suggestion
```

Swatinem (commented):
looks like you did a `cargo update` as part of this PR? we should maybe do that separately.

crates/symbolicator-service/src/services/symbolication/sourcemap.rs
kamilogorek: That would be possible as well I guess

kamilogorek:
Removed `dbg!`, updated cache name and rebased to fix `Cargo.lock`.

Swatinem (approved):

PR #1052, opened 2023-02-21
ref: More docs and Cach cleanups

Swatinem:
- Increases the in-memory defaults a bit
- Cleans up the buggy `expiration_strategy` handling
- Remove a footgun from cache configuration
- Add more docs to the caching infrastructure

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/caching/config.rs
+++ b/crates/symbolicator-service/src/caching/config.rs
@@ -26,7 +26,7 @@ impl AsRef<str> for CacheName {
             Self::Cficaches => "cficaches",
             Self::PpdbCaches => "ppdb_caches",
             Self::ArtifactCaches => "artifact_caches",
-            Self::SourceMapCaches => "sourcesmap_caches",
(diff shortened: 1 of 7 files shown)


crates/symbolicator-service/src/caching/config.rs
```diff
@@ -26,7 +26,7 @@ impl AsRef<str> for CacheName {
             Self::Cficaches => "cficaches",
             Self::PpdbCaches => "ppdb_caches",
             Self::ArtifactCaches => "artifact_caches",
-            Self::SourceMapCaches => "sourcesmap_caches",
+            Self::SourceMapCaches => "sourcemap_caches",
```
Swatinem, line 29: kinda glad we didn’t actually start using the functionality with this typo in it :-D

loewenheim (approved):

crates/symbolicator-service/src/caching/config.rs
kamilogorek: Noone would ever notice, which means that it was never there 🤫

PR #1061, opened 2023-02-24
Un-batch SourceMap processing

Swatinem:
Hook up the new infra to sourcemap processing, moving it away from a heavily batched / parallelizes approach to a more straight-line code approach.

Files are fetched as-needed, which was already the case previously as needed for applying source context from individual files.

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/services/sourcemap_lookup.rs
+++ b/crates/symbolicator-service/src/services/sourcemap_lookup.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::fs::File;
 use std::io::{self, BufWriter, Write};
 use std::sync::Arc;
@@ -19,6 +19,7 @@ use url::Position;
 use crate::caching::{CacheEntry, CacheError, CacheItemRequest, CacheVersions, Cacher};
 use crate::services::download::sentry::SearchArtifactResult;
 use crate::services::download::DownloadService;
+use crate::types::JsStacktrace;
 
 use super::caches::versions::{ARTIFACT_CACHE_VERSIONS, SOURCEMAP_CACHE_VERSIONS};
 use super::fetch_file;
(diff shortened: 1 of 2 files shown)


crates/symbolicator-service/src/services/sourcemap_lookup.rs
```diff
@@ -369,6 +244,54 @@ impl SourceMapLookup {
...
+        if let Some(file) = self.try_get_file_from_bundles(&key) {
+            return self.files_by_key.entry(key).or_insert(Ok(file)).clone();
+        }
+
+        // Otherwise, try to get the file from an individual artifact.
+        // This is mutually exclusive with having a `DebugId`, and we only care about `abs_path` here.
+        // If we have a `DebugId`, we are guaranteed to use artifact bundles, and to have found the
+        // file in the check up above already.
+        if let Some(file) = self.try_fetch_file_from_artifacts(&key).await {
+            return self.files_by_key.entry(key).or_insert(Ok(file)).clone();
```
kamilogorek, line 268: I'm surprised there's no method to insert and return inserted value without using `Entry` api.

crates/symbolicator-service/src/services/sourcemap_lookup.rs
```diff
@@ -369,6 +244,54 @@ impl SourceMapLookup {
...
+        }
+
+        // Otherwise: Do a (cached) API lookup for the `abs_path` + `DebugId`
```
kamilogorek, line 271: Should we move this to after `try_get_file_from_bundles`? Or will it be prepopulated by `list_artifacts` as well? (asking as it's not right now)

crates/symbolicator-service/src/services/sourcemap_lookup.rs
```diff
@@ -369,6 +244,54 @@ impl SourceMapLookup {
...
+        if self.remote_artifacts.is_empty() {
+            // TODO: the endpoint should really support filtering files,
+            // and ideally only giving us a single file that matches, but right now it just gives
+            // us the whole list of artifacts
+            self.remote_artifacts = self.list_artifacts().await;
```
kamilogorek, line 276: It does support filtering by name, so maybe we just give it our 4-variants of `abs_path` and this should suffice?

crates/symbolicator-service/src/services/sourcemap_lookup.rs
Swatinem: I believe it will be pre-populated.

crates/symbolicator-service/src/services/sourcemap_lookup.rs
kamilogorek: You're right. Just checked and it's doing a partial matching, so should be enough.

```py
condition = Q(name__icontains=query[0])
    for name in query[1:]:
        condition |= Q(name__icontains=name)
```

PR #1063, opened 2023-02-27
Fix `missing_sourcemap` Status assertions

Swatinem:
Looks like merging both changes to the sourcemap processing, and switching to snapshot tests was causing conflicts here.

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/tests/integration/snapshots/integration__sourcemap__source_expansion.snap
+++ b/crates/symbolicator-service/tests/integration/snapshots/integration__sourcemap__source_expansion.snap
@@ -4,7 +4,7 @@ expression: response.unwrap()
 ---
 stacktraces:
   - frames:
-      - status: symbolicated
+      - status: missing_sourcemap
         filename: foo.js
         abs_path: "http://example.com/foo.js"
         lineno: 1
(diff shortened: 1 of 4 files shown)


loewenheim (approved):

crates/symbolicator-service/tests/integration/snapshots/integration__sourcemap__source_expansion.snap
```diff
@@ -4,7 +4,7 @@ expression: response.unwrap()
 ---
 stacktraces:
   - frames:
-      - status: symbolicated
+      - status: missing_sourcemap
```
kamilogorek, line 7: I'm not sure about this error tbh. In original tests, we simply applied source context to the frames, and that's it. No errors, no `raw_stacktrace` (as no source maps were applied, thus we only kept around original frames).

crates/symbolicator-service/tests/integration/snapshots/integration__sourcemap__sourcemap_embedded_source_expansion.snap
```diff
@@ -4,7 +4,7 @@ expression: response.unwrap()
 ---
 stacktraces:
   - frames:
-      - status: symbolicated
+      - status: missing_sourcemap
```
kamilogorek, line 7: This should say `missing_source` if anything, but we also dont have such status now.

PR #1064, opened 2023-02-27
test: Add sourcemap test for malformed abs_path

kamilogorek:
#skip-changelog


tests/fixtures/sourcemaps/08_malformed_abs_path/files.json
```diff
@@ -0,0 +1 @@
+[]
```
Swatinem, line 1: do we even need an empty file here? maybe in a different test we should add a test for failing to fetch the manifest

Swatinem (approved):

tests/fixtures/sourcemaps/08_malformed_abs_path/files.json
kamilogorek: We need it for now so that `sourcemap_router` wont break😅

PR #1075, opened 2023-03-01
cd: use deployment asset bucket instead of docker

joshuarli:
This utilizes deployment asset buckets instead of docker for passing DIFs around.
Symbolicator releases should work after this!

successful cloudbuild (i manually triggered): https://console.cloud.google.com/cloud-build/builds;region=global/31800b98-b220-48f0-b233-965e00f6252a?project=sentryio

#skip-changelog


Dockerfile
```diff
@@ -68,7 +68,6 @@ RUN mkdir -p /etc/symbolicator /data && \
 EXPOSE 3021
 
 COPY --from=symbolicator-build /usr/local/bin/symbolicator /bin
-COPY --from=symbolicator-build /opt/symbolicator-debug.zip /opt/symbolicator.src.zip /opt/
```
ashwoods, line 71: I'd consider leaving the debug symbols in the container, even if we do not use it directly. Could eventually be used downstream (in self hosted instances?) and it doesn't really hurt.

ashwoods (approved):

Dockerfile
ashwoods: We can also remove it and see if anyone complains. :)

joshuarli: Well, they're available separately in github releases and the size isn't insignificant. I'd say this is a good opportunity to remove it and see if it'll be missed. If so, no problem, can add back.

Swatinem: Well, the docker containers do not correspond to the github releases at all, so the debug files are not compatible with one another. I don’t really know how self-hosted is dealing with these things.
But that being said, we only need these things in extremely rare cases, as symbolicator almost never panics, and our normal events do not have stack traces that need symbolication/debug-files.

PR #1076, opened 2023-03-02
Use SourceFileCache to implement web-scraping

Swatinem:
- Removes the `ArtifactCache` and uses the`SourceFileCache` instead.
- Implements web-scraping for sourcemap and source file downloads.
- Converts `SourceMapLookup` internals to `ByteViewString`.
- ^ The above will allow us to simplify/unify source context application with native symbolication.

#skip-changelog


loewenheim (approved):

crates/symbolicator-service/src/services/caches/sourcefiles.rs
```diff
@@ -42,22 +55,47 @@ impl SourceFilesCache {
...
-            cache: Arc::new(Cacher::new(cache, shared_cache)),
+            cache: Cacher::new(cache, shared_cache),
             download_svc,
         }
     }
 
     /// Retrieves the given [`RemoteFile`] from cache, or fetches it and persists it according
     /// to the provided [`Scope`].
     pub async fn fetch_file(&self, scope: &Scope, file: RemoteFile) -> CacheEntry<ByteViewString> {
+        // FIXME: We should probably make it possible to somehow predictably re-fetch
```
kamilogorek, line 66: > respect the caching headers from external servers

That'll be quite tricky tbh, as we always want the exact file at the time exception happened.

There can be a case that we'll let's say fetch and cache a `app.js` file that has 24h cache expiration header, but it has been already updated on the server. For all new unique users that trigger an exception, this would possibly be somehow out-of-sync and they frames would point to incorrect lines in the source code, but at the same time, I'm not sure how possible this scenario is.

I think for now it's fine as is and we may actually try to overengineer ourselves here 😅

crates/symbolicator-service/src/services/sourcemap_lookup.rs
```diff
@@ -526,8 +505,19 @@ impl ArtifactFetcher {
             return Ok(file);
         }
 
-        // TODO:
         // Otherwise, fall back to scraping from the Web.
+        if let Some(url) = key.abs_path() {
```
kamilogorek, line 509: This should be guarded with the `allow_scraping` option that we pass as the part of request payload from the monolith.

kamilogorek:
```rust
pub struct JsSymbolicationRequestBody {
    #[serde(default)]
    pub source: Option<SentrySourceConfig>,
    #[serde(default)]
    pub stacktraces: Vec<JsStacktrace>,
    #[serde(default)]
    pub modules: Vec<RawObjectInfo>,
    #[serde(default)]
    pub dist: Option<String>,
}
```

This should be extended with:

```rs
#[serde(default = true)]
pub allow_scraping: bool
```

(default in monolith is `True`)

crates/symbolicator-service/src/services/sourcemap_lookup.rs
```diff
@@ -526,8 +505,19 @@ impl ArtifactFetcher {
...
+            // TODO: Do we want to scope this to the project/org?
```
kamilogorek, line 510: Definitely, as `allow_scraping` is per-project setting.

kamilogorek (changes requested):

crates/symbolicator-service/src/services/caches/sourcefiles.rs
Swatinem: Even then, you might have users with an older version cached. Or you have some send delay for whatever reason, and you are not able to fetch the *older* file at all. So yeah, we can revisit this later when it turns out to be a real problem, otherwise lets not overengineer :-D

kamilogorek (approved):

PR #1081, opened 2023-03-07
ref: Update dependencies, bump PortablePDB Cache version

Swatinem:
- Does a `cargo update`
- Bumps the PortablePDB cache version to refresh caches with skipped hidden SequencePoints
- Smaller changes related to updated `tower-http` and `wasmbin`.

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/Cargo.toml
+++ b/crates/symbolicator-service/Cargo.toml
@@ -33,13 +33,13 @@ moka = { version = "0.10", features = ["future"] }
 parking_lot = "0.12.0"
 regex = "1.5.5"
 reqwest = { version = "0.11.0", features = ["gzip", "json", "stream", "trust-dns"] }
-sentry = { version = "0.29.2", features = ["tracing"] }
+sentry = { version = "0.30.0", features = ["tracing"] }
 serde = { version = "1.0.137", features = ["derive", "rc"] }
 serde_json = "1.0.81"
 serde_yaml = "0.9.14"
 sha2 = "0.10.6"
 sourcemap = "6.2.1"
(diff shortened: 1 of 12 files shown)


crates/symbolicator-service/Cargo.toml
```diff
@@ -33,13 +33,13 @@ moka = { version = "0.10", features = ["future"] }
 parking_lot = "0.12.0"
 regex = "1.5.5"
 reqwest = { version = "0.11.0", features = ["gzip", "json", "stream", "trust-dns"] }
-sentry = { version = "0.29.2", features = ["tracing"] }
+sentry = { version = "0.30.0", features = ["tracing"] }
 serde = { version = "1.0.137", features = ["derive", "rc"] }
 serde_json = "1.0.81"
 serde_yaml = "0.9.14"
 sha2 = "0.10.6"
 sourcemap = "6.2.1"
```
kamilogorek, line 41: Can we bump `sourcemap` as well while we're at it?

crates/symbolicator-test/src/lib.rs
```diff
@@ -210,8 +210,7 @@ impl Server {
 
     /// Creates a new [`Router`] with the configuration as described in the main [`Server`] docs.
     pub fn test_router() -> Router {
-        let serve_dir = get_service(ServeDir::new(fixture("symbols")))
```
kamilogorek, line 213: Oh nice

crates/symbolicator-service/Cargo.toml
Swatinem: That happened as part of `cargo update` implicitly as its semver compatible. For some dependencies, I tend to bump directly, for others I’m kinda 🤷🏻‍♂️

kamilogorek: Ah I see in cargo.lock now.

kamilogorek (approved):

PR #1082, opened 2023-03-07
ref: Unify Source Context application

Swatinem:
This now uses the same code for native and JS source context, which now works with `&str`, and makes it generic over line trimming and number of context lines.

It also avoids outputting a trailing empty line when there is a trailing lineline in JS files.

Fun fact: The source-context application for native was broken since forever basically. Due to an off-by-one, it would only ever give us 4 lines of `pre_context`.

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/services/symbolication/source_context.rs
+++ b/crates/symbolicator-service/src/services/symbolication/source_context.rs
@@ -1,0 +1,98 @@
+pub const DEFAULT_CONTEXT_LINES: usize = 5;
+
+/// Resolves the source context around `lineno` from the `source` file.
+///
+/// N (`context_lines`) lines of context will be resolved in addition to `lineno`.
+/// This defaults to [`DEFAULT_CONTEXT_LINES`].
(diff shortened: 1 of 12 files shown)


loewenheim (approved):

crates/symbolicator-service/src/services/symbolication/source_context.rs
```diff
@@ -0,0 +1,98 @@
...
+/// This defaults to [`DEFAULT_CONTEXT_LINES`].
+/// If `trim_to_column` is provided, every output line will be truncated to ~150 characters
+/// centered around the provided column number, and a `{snip}` marker is added at the trimmed edges.
+///
+/// If no source line for `lineno` is found in `source`, it will return [`None`] instead.
+pub fn get_context_lines(
+    source: &str,
+    lineno: usize,
+    trim_to_column: Option<usize>,
+    context_lines: Option<usize>,
```
kamilogorek, line 15: I wonder if we really need that. Do we ever configure it anywhere or changed it in the whole history of Sentry? 🤔

PR #1084, opened 2023-03-08
fix: Skip `releases set-commits`

Swatinem:
We currently don’t have a valid GitHub integration in our S4S instance, hence the commits feature does not work and would otherwise error out.


loewenheim (approved):

kamilogorek:
image

PR #1094, opened 2023-03-13
ref: Enforce assumptions around DebugId

Swatinem:
This adds and documents a couple of shortcuts we do when we have a `DebugId`:

- We will not collect/query for file stems on Sentry.
- We assume the file will be found in one of the prefetched `ArtifactBundle`s
- We will not fall back to looking up by filename in `ArtifactBundle`s.
- We will not fall back to scraping from the Web.
- We assume the `SourceMap` has embedded sources.

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/services/symbolication/js.rs
+++ b/crates/symbolicator-service/src/services/symbolication/js.rs
@@ -4,7 +4,7 @@ use symbolic::sourcemapcache::{ScopeLookupResult, SourcePosition};
 use symbolicator_sources::SentrySourceConfig;
 
 use crate::caching::{CacheEntry, CacheError};
-use crate::services::sourcemap_lookup::{CachedFile, OwnedSourceMapCache, SourceMapLookup};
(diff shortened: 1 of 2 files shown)


crates/symbolicator-service/src/services/symbolication/js.rs
```diff
@@ -126,29 +106,42 @@ fn symbolicate_js_frame(
...
-    let mut did_apply_source = false;
     if let Some(file) = token.file() {
-        result.raw.filename = file.name().map(|f| f.to_string());
+        frame.raw.filename = file.name().map(|f| f.to_string());
         if let Some(file_source) = file.source() {
-            apply_source_context(&mut result.raw, file_source);
-            did_apply_source = true;
+            apply_source_context(&mut frame.raw, file_source);
+        } else if module.has_debug_id() {
+            tracing::error!("expected `SourceMap` with `DebugId` to have embedded sources");
```
kamilogorek, line 128: noice

PR #1096, opened 2023-03-14
ref: Use API query param to prefilter release artifact files

kamilogorek:
This should make it work for projects with thousands and thousands of release files, without making tens of hundreds of requests to prefetch release files metadata.

#skip-changelog


crates/symbolicator-service/src/services/sourcemap_lookup.rs
```diff
@@ -692,15 +693,25 @@ impl ArtifactFetcher {
 
 /// Extracts a "file stem" from a [`Url`].
 /// This is the `"file"` in `"./path/to/file.min.js?foo=bar"`.
```
Swatinem, line 695:
```suggestion
/// This is the `"/path/to/file"` in `"./path/to/file.min.js?foo=bar"`.
```

Swatinem (approved):

PR #1103, opened 2023-03-16
feat: source context links support

vaind:
Part of getsentry/symbolic#735


crates/symbolicator-service/src/services/module_lookup.rs
```diff
@@ -107,9 +107,15 @@ struct ModuleEntry {
 pub struct ModuleLookup {
     modules: Vec<ModuleEntry>,
     scope: Scope,
+    // TODO rename variable and type so it wouldn't get confused with source code. Maybe "Provider"?
```
Swatinem, line 110: I believe `symbol_sources` would also work as a name

crates/symbolicator-service/src/services/symbolication/mod.rs
```diff
@@ -1,14 +1,20 @@
+use std::collections::HashMap;
 use std::sync::Arc;
 
+use futures::future;
+use lazy_static::__Deref;
```
Swatinem, line 5: why exactly is this here? I don’t see any usage of `lazy_static` here

crates/symbolicator-service/src/services/symbolication/mod.rs
```diff
@@ -138,21 +147,45 @@ impl SymbolicationActor {
...
+            let futures = remote_sources.into_iter().map(|(url, frames)| async move {
+                if let Ok(source) = cache
+                    .fetch_file(scope_clone, HttpRemoteFile::from_url(url).into())
+                    .await
                 {
-                    frame.raw.pre_context = pre_context;
-                    frame.raw.context_line = Some(context_line);
-                    frame.raw.post_context = post_context;
+                    for frame in frames {
+                        ModuleLookup::set_context_lines_from_source(source.deref(), frame);
```
Swatinem, line 183:
```suggestion
                        ModuleLookup::set_context_lines_from_source(&source, frame);
```

crates/symbolicator-service/src/services/module_lookup.rs
```diff
@@ -383,22 +389,60 @@ impl ModuleLookup {
...
-    pub fn get_context_lines(
+    /// Update the frame with source context, if available. In case the source code references an
+    /// URL that's not yet loaded, returns `SetContextLinesResult::SourceMissing(url)`.
+    pub(crate) fn set_source_context(
         &self,
         debug_sessions: &HashMap<usize, Option<ObjectDebugSession<'_>>>,
-        frame: &RawFrame,
-    ) -> Option<(Vec<String>, String, Vec<String>)> {
+        frame: &mut RawFrame,
+    ) -> Option<SetSourceContextResult> {
```
Swatinem, line 398: IMO it might be simpler to return an `Option<Url>` here.
You only ever match the `Some(SourceMissing(...))` part.

crates/symbolicator-service/src/services/module_lookup.rs
```diff
@@ -383,22 +383,54 @@ impl ModuleLookup {
             .collect()
     }
 
-    /// This looks up the source of the given line, plus `n` lines above/below.
-    pub fn get_context_lines(
+    /// Update the frame with source context, if available. In case the source code references an
+    /// URL that's not yet loaded, returns `SetContextLinesResult::SourceMissing(url)`.
```
Swatinem, line 387:
```suggestion
    /// Update the frame with source context, if available.
    /// Returns an `Url` in case the source code has to be fetched.
```

Swatinem (approved):

crates/symbolicator-service/src/services/symbolication/mod.rs
```diff
@@ -138,21 +144,44 @@ impl SymbolicationActor {
...
+        {
+            let debug_sessions = module_lookup.prepare_debug_sessions();
+
+            for trace in &mut stacktraces {
+                for frame in &mut trace.frames {
+                    if let Some(url) =
+                        module_lookup.try_set_source_context(&debug_sessions, &mut frame.raw)
+                    {
+                        // Only resolve source context from URLs if the frame is "in-app".
+                        if frame.raw.in_app.unwrap_or(false) {
```
Swatinem, line 158: This also needs changes in the sentry processor to enable this, as it current does not send any `in_app` attributes.

The other problems with this that come to mind:
- `in_app` is being modified/set by grouping, which runs *after* symbolication.
- minidumps never set `in_app` at all. neither do some SDKs like Native. However those debug files do not (yet?) have support for source links.

I’m okay with merging this as is, but this might need to be revisited.

vaind: in .NET, in_app is being sent from the SDK. What would need to change for that to get propagated here?

PR #1113, opened 2023-03-23
fix: Use correct urls order in get_release_file_candidate_urls

kamilogorek:
ref: https://github.com/getsentry/sentry/blob/02bf7f0df65bd662a4caca0c4e53aa281aea1488/src/sentry/models/releasefile.py#L122-L144

#skip-changelog


Swatinem (approved):
A comment above the fn why this order matters would be very helpful. I originally wanted to optimize for the most likely pattern to match first. Now we have to always try everything because the most likely thing to match is actually the last one.

PR #1114, opened 2023-03-23
ref: Port get_function_for_token from monolith

kamilogorek:
ref: https://github.com/getsentry/sentry/blob/02bf7f0df65bd662a4caca0c4e53aa281aea1488/src/sentry/lang/javascript/processor.py#L580-L608

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/services/symbolication/js.rs
+++ b/crates/symbolicator-service/src/services/symbolication/js.rs
@@ -73,19 +73,19 @@ impl SymbolicationActor {
         for raw_stacktrace in &mut raw_stacktraces {
             let num_frames = raw_stacktrace.frames.len();
             let mut symbolicated_frames = Vec::with_capacity(num_frames);
-            let mut prev_frame_token_name = None;
+            let mut callsite_fn_name = None;
 
             for raw_frame in &mut raw_stacktrace.frames {
                 match symbolicate_js_frame(
                     &mut lookup,
                     raw_frame,
-                    prev_frame_token_name.clone(),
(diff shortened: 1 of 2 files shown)


crates/symbolicator-service/src/services/symbolication/js.rs
```diff
@@ -73,12 +73,19 @@ impl SymbolicationActor {
...
             let mut symbolicated_frames = Vec::with_capacity(num_frames);
+            let mut prev_frame_token_name = None;
 
             for raw_frame in &mut raw_stacktrace.frames {
-                match symbolicate_js_frame(&mut lookup, raw_frame, &mut missing_sourcescontent)
-                    .await
+                match symbolicate_js_frame(
+                    &mut lookup,
+                    raw_frame,
+                    prev_frame_token_name.clone(),
```
Swatinem, line 82:
```suggestion
                    std::mem::take(&mut prev_frame_token_name),
```
Because:
- in the `Ok` branch, you overwrite it anyway
- in the `Err` branch, you don’t want it to span over more than one (unable to symbolicate) frame anyway

Swatinem (commented):
The naming is quite confusing tbh, though I don’t really have a better idea either.
The "token name" basically is the mapped "name" at the caller location.
If we minified `fnThatThrows();` to `a()`, and the stack frame points to `a`, the "name" would be `fnThatThrows`. Maybe putting an example like this in a comment somewhere might make this clearer.

Swatinem (approved):

PR #1119, opened 2023-03-28
Propagate `CacheError`s and `RemoteFileUri`s

Swatinem:
This does two things in `SourceMapLookup` and `ArtifactFetcher`:
- It keeps `CacheError`/`CacheEntry` around for files that we expect to exist.
- And it also keeps track of the `CachedFileUri`, which will show us where the file was being fetched from.

So far these new errors and this new info is being discarded in symbolication, but this opens up the road to eventually feed this data back to Sentry and show it to the user.

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/services/sourcemap_lookup.rs
+++ b/crates/symbolicator-service/src/services/sourcemap_lookup.rs
@@ -383,13 +383,13 @@ impl FileKey {
 /// The source of an individual file.
 #[derive(Clone, Debug)]
 pub enum CachedFileUri {
-    /// This file was fetched using its own URI.
+    /// The file was fetched using its own URI.
     IndividualFile(RemoteFileUri),
     /// The file was found using [`FileKey`] in the bundle identified by the URI.
     Bundled(RemoteFileUri, FileKey),
-    /// The SourceMap was embedded in a minified file.
-    /// This will only ever happen for SourceMaps, and the SourceMap is always used in combination
-    /// with a minified File that has a [`CachedFileUri`] itself.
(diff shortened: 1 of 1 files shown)


crates/symbolicator-service/src/services/sourcemap_lookup.rs
```diff
@@ -378,14 +380,41 @@ impl FileKey {
...
+#[derive(Clone, Debug)]
+pub enum CachedFileUri {
+    /// This file was fetched using its own URI.
+    IndividualFile(RemoteFileUri),
+    /// The file was found using [`FileKey`] in the bundle identified by the URI.
+    Bundled(RemoteFileUri, FileKey),
+    /// The SourceMap was embedded in a minified file.
+    /// This will only ever happen for SourceMaps, and the SourceMap is always used in combination
+    /// with a minified File that has a [`CachedFileUri`] itself.
+    Embedded,
```
kamilogorek, line 393:
```suggestion
    /// The file was fetched using its own URI.
    IndividualFile(RemoteFileUri),
    /// The file was found using [`FileKey`] in the bundle identified by the URI.
    Bundled(RemoteFileUri, FileKey),
    /// The file was embedded in another file. This will only ever happen
    /// for Base64-encoded SourceMaps, and the SourceMap is always used
    /// in combination with a minified File that has a [`CachedFileUri`] itself.
    Embedded,
```

kamilogorek: Consistent wording

kamilogorek (approved):
Just a nit

PR #1120, opened 2023-03-31
Clean up anyhow usage and Symbolication Error reporting

Swatinem:
This makes sure that when capturing a Symbolication Error, we do get a proper stack trace and chained errors for it. Also cleans up some general anyhow usage across the codebase.

#skip-changelog

### DIFF
--- a/crates/symbolicator/src/service.rs
+++ b/crates/symbolicator/src/service.rs
@@ -25,7 +25,6 @@ use sentry::protocol::SessionStatus;
 use sentry::SentryFutureExt;
 use serde::{Deserialize, Deserializer, Serialize};
 use tempfile::TempPath;
-use thiserror::Error;
 use uuid::Uuid;
 
 use symbolicator_service::caching::CacheEntry;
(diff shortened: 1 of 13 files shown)


crates/symbolicator/src/service.rs
```diff
@@ -442,18 +414,27 @@ impl RequestService {
...
+                Ok(Err(err)) => {
+                    // NOTE: This will double-report this error to Sentry.
+                    // We want the `tracing::error` here for local command line output, but we also
+                    // want the `capture_anyhow` which will give a full error with a symbolicated backtrace.
+                    // So until we can actually get a `Backtrace` out of an `Error`, and `anyhow`
+                    // implements that functionality through its `Error` impl, we are stuck with this for now.
+                    // See <https://github.com/rust-lang/rust/issues/99301>
+                    let error: &dyn std::error::Error = err.as_ref();
+                    tracing::error!(error, "Symbolication failed");
+                    sentry_anyhow::capture_anyhow(&err);
```
Swatinem, line 426: this is unfortunate indeed. Ideally I would want to only keep the `tracing::error`, but we are not there yet.

crates/symbolicator/src/service.rs
```diff
@@ -383,7 +361,7 @@ impl RequestService {
         f: F,
     ) -> Result<RequestId, MaxRequestsError>
     where
-        F: Future<Output = Result<CompletedResponse, anyhow::Error>> + Send + 'static,
+        F: Future<Output = Result<CompletedResponse>> + Send + 'static,
```
kamilogorek, line 364:
```suggestion
        F: Future<Output = anyhow::Result<CompletedResponse>> + Send + 'static,
```

Do we want to use ^ for consistency? We can also do `use anyhowe::Result` in other places instead, I'm fine with both, as long as its consistent.

Swatinem: Good point. I believe `anyhow::Result` is a drop-in replacement for `Result`. It is just a type alias with a default Error filled in, so we should be able to just `use anyhow::Result` if we agree on that.

kamilogorek: Sure, lets do that

Swatinem: I replaced some of the cases with a `use anyhow::Result`. For other places where there is only a single use in a file, I rather kept it there.

kamilogorek (approved):

PR #1126, opened 2023-04-06
ref(sourcemaps): Make SourceMapCache optional for Module and bail early on processing

kamilogorek:
This change prevents errors about missing sources for events that should not report one.
When we have a frame that points to a valid JS file but has no `sourceMappingUrl` nor a Debug ID, we shouldn't treat it as an error.

We should also bail from further processing when the initial context lines application fails, as it means that minified source (which is required for creating a `SourceMapCache` in the first place) is not present or is not usable.

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/services/sourcemap_lookup.rs
+++ b/crates/symbolicator-service/src/services/sourcemap_lookup.rs
@@ -83,7 +83,7 @@ pub struct SourceMapModule {
     pub minified_source: CachedFileEntry,
     /// The converted SourceMap.
     // TODO(sourcemap): maybe this should not be public?
-    pub smcache: CachedFileEntry<OwnedSourceMapCache>,
+    pub smcache: Option<CachedFileEntry<OwnedSourceMapCache>>,
 }
 
 impl SourceMapModule {
@@ -99,7 +99,7 @@ impl SourceMapModule {
             was_fetched: false,
             source_file_base: None,
             minified_source: CachedFileEntry::empty(),
-            smcache: CachedFileEntry::empty(),
+            smcache: None,
         }
     }
 
@@ -525,7 +525,10 @@ impl ArtifactFetcher {
         &mut self,
         abs_path: Url,
         debug_id: Option<DebugId>,
-    ) -> (CachedFileEntry, CachedFileEntry<OwnedSourceMapCache>) {
+    ) -> (
+        CachedFileEntry,
+        Option<CachedFileEntry<OwnedSourceMapCache>>,
+    ) {
         // First, check if we have already cached / created the `SourceMapCache`.
         let key = FileKey::MinifiedSource {
             abs_path: abs_path.clone(),
@@ -540,6 +543,12 @@ impl ArtifactFetcher {
             Ok(minified_source) => minified_source.sourcemap_url.as_deref(),
             Err(_) => None,
         };
+
+        // If we don't have sourcemap reference, nor a `DebugId`, we skip creating `SourceMapCache`.
+        if sourcemap_url.is_none() && debug_id.is_none() {
+            return (minified_source, None);
+        }
+
         // We have three cases here:
         let sourcemap = match sourcemap_url {
             // We have an embedded SourceMap via data URL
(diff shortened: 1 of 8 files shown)


crates/symbolicator-service/src/services/symbolication/js.rs
```diff
@@ -213,23 +218,31 @@ async fn symbolicate_js_frame(
...
-fn apply_source_context_from_artifact(frame: &mut JsFrame, file: &CacheEntry<CachedFile>) {
+fn apply_source_context_from_artifact(
+    frame: &mut JsFrame,
+    file: &CacheEntry<CachedFile>,
+) -> Result<(), JsModuleErrorKind> {
     if let Ok(file) = file {
         apply_source_context(frame, &file.contents)
     } else {
-        // TODO(sourcemap): report missing source?
+        Err(JsModuleErrorKind::MissingSourcemap)
```
Swatinem, line 235:
```suggestion
        Err(JsModuleErrorKind::MissingSource)
```

We are either missing the minified or the original source file here, as we are not calling this function with a *sourcemap*.

crates/symbolicator-service/src/services/symbolication/js.rs
```diff
@@ -197,7 +202,7 @@ async fn symbolicate_js_frame(
     if let Some(file) = token.file() {
         frame.filename = file.name().map(|f| f.to_string());
         if let Some(file_source) = file.source() {
-            apply_source_context(&mut frame, file_source);
+            apply_source_context(&mut frame, file_source).ok();
```
Swatinem, line 205:
```suggestion
            apply_source_context(&mut frame, file_source)?;
```

crates/symbolicator-service/src/services/symbolication/js.rs
```diff
@@ -213,23 +218,31 @@ async fn symbolicate_js_frame(
                 None => &Err(CacheError::NotFound),
             };
 
-            apply_source_context_from_artifact(&mut frame, source_file);
+            apply_source_context_from_artifact(&mut frame, source_file).ok();
```
Swatinem, line 221:
```suggestion
            apply_source_context_from_artifact(&mut frame, source_file)?;
```

Here and above, I believe we can just directly propagate these.
I don’t think it matters much as you are setting `frame.lineno` above, so this should be unreachable.

crates/symbolicator-service/src/services/sourcemap_lookup.rs
```diff
@@ -989,7 +989,11 @@ fn write_sourcemap_cache(file: &mut File, source: &str, sourcemap: &str) -> Cach
     // TODO(sourcemap): maybe log *what* we are converting?
     tracing::debug!("Converting SourceMap cache");
 
-    let smcache_writer = SourceMapCacheWriter::new(source, sourcemap).unwrap();
+    // There's currently no way to access inner error of `SourceMapCacheWriterError`,
+    // if we change that. We might want to match on `SourceMap` and get actual underlying
+    // `sourcemap::Error` instead to see more details.
+    let smcache_writer = SourceMapCacheWriter::new(source, sourcemap)
+        .map_err(|_| CacheError::Malformed("Unable to produce a SourceMapCache".to_string()))?;
```
Swatinem, line 996: I believe you can just `to_string` / `Display` the inner error in this case? I believe we do the same thing for malformed files in native.

crates/symbolicator-service/src/services/symbolication/js.rs
```diff
@@ -202,9 +204,28 @@ async fn symbolicate_js_frame(
...
+            // the SourceMapCache tokens filename, and we want the original one.
+            errors.insert(JsModuleError {
+                abs_path: raw_frame.abs_path.clone(),
+                // TODO: This error should include a `sourcemap_label`, which is in form of:
+                // - `raw_frame.abs_path` if sourcemap was base64 encoded
+                // - resolved `sourceMappingUrl` if we used `abs_path` to find it
+                // - `debug-id://{debug_id}/~/{filename}` if we used `DebugId` to find it
+                // Or, because we control all the errors in the monolith, we can just drop this detail.
+                kind: JsModuleErrorKind::MissingSourceContent {
+                    sourcemap: String::from(""),
```
Swatinem, line 226: Can we use the `module.smcache.uri` for this?

Swatinem (approved):

PR #1128, opened 2023-04-06
Skip Ubuntu symbol server test

Swatinem:
According to the FAQ at https://ubuntu.com/server/docs/service-debuginfod-faq:

> Once a release goes unsupported, we [...] eventually stop serving debug symbols for its packages.

So I guess with those "stability" guarantees we either have to constantly update the testcase with up-to-date debug-ids, or we just skip it entirely :-(

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/tests/integration/public_sources.rs
+++ b/crates/symbolicator-service/tests/integration/public_sources.rs
@@ -48,6 +48,7 @@ async fn test_nuget_source() {
     assert_snapshot!(response.unwrap());
 }
 
+#[ignore]
 #[tokio::test]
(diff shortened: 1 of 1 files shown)


kamilogorek (approved):
😢

ashwoods:
For consistency, I think we should self-host whatever we want to test against.

Swatinem:
I agree, as this evidently is a point of failure. However these tests explicitly hit these very specific external symbol servers. We also have #1004 as we hit the microsoft symbol server more than we absolutely need to. In that case we should also keep at least one test around that actually does hit it though.

PR #1130, opened 2023-04-06
Migrate to monthly CalVer releases

ashwoods:


.github/workflows/release.yml
```diff
@@ -3,24 +3,31 @@ on:
...
-        description: Version to release
-        required: true
+        description: Version to release (optional)
+        required: false
       force:
         description: Force a release even when there are release-blockers (optional)
         required: false
+  schedule:
+    # We want the release to be at 9-10am Pacific Time
+    # We also want it to be 1 hour before the on-prem release
```
Swatinem, line 13: I remember our `release` workflows are quite slow, is 1 hour lead time enough here?

Swatinem (approved):

.github/workflows/release.yml
ashwoods: Same as relay, lets roll with this for now.

PR #1134, opened 2023-04-07
ref(sourcemaps): Allow for fallback to legacy bundles/individial artifacts for unavailable debugids

kamilogorek:
This is unfortunately a desired behavior that is in-line with what we already do in the monolith.


crates/symbolicator-service/src/services/sourcemap_lookup.rs
```diff
@@ -518,9 +513,8 @@ impl ArtifactFetcher {
         for module in modules {
             if let Some(debug_id) = module.debug_id {
                 debug_ids.insert(debug_id);
-                // If we have a `DebugId`, we assume that the `DebugId`-based lookup will succeed.
-                // In that case, we do not want to look up the file by its file stem.
-            } else if let Ok(url) = module.abs_path.as_ref() {
+            }
+            if let Ok(url) = module.abs_path.as_ref() {
```
Swatinem, line 517: I would not do this in `prefetch`, as `prefetch` should give us some kind of bundle to work with. we would then expect files to be available in that bundle.

Swatinem (approved):

crates/symbolicator-service/src/services/sourcemap_lookup.rs
kamilogorek: If we wont do this in the prefetch, then `abs_path` key is locked in a cache as `NotFound` for the rest of processing, as `artifact-lookup` endpoint was not able to query for it due to `url=` query param not being passed here.

crates/symbolicator-service/src/services/symbolication/js.rs
```diff
@@ -223,34 +223,30 @@ async fn symbolicate_js_frame(
...
 
-            if let Err(err) = apply_source_context_from_artifact(&mut frame, source_file) {
+            if apply_source_context_from_artifact(&mut frame, source_file).is_err() {
+                // It's arguable whether we should collect it, but this is what monolith does now,
+                // and it might be useful to indicate incorrect sentry-cli rewrite behavior.
                 errors.insert(JsModuleError {
                     abs_path: raw_frame.abs_path.clone(),
-                    kind: err,
+                    kind: JsModuleErrorKind::MissingSourceContent {
+                        source: file_key
```
kamilogorek, line 245: Not sure about pulling the details like this from here but it works ¯\_(ツ)_/¯

kamilogorek:
With the current state of this PR, all tests at https://github.com/getsentry/sentry/pull/47008 are green.

kamilogorek:
Once this PR is merged, we need to deploy and rerun all tests from https://github.com/getsentry/sentry/pull/47008 to make sure it's all consistent.

PR #1137, opened 2023-04-13
Make the JS `abs_path` actually an absolute path/url

Swatinem:
Previously, we just returned the `filename` here. But this is supposed to be a full url, relative to the SourceMap url (or the minified file url as fallback).

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/services/sourcemap_lookup.rs
+++ b/crates/symbolicator-service/src/services/sourcemap_lookup.rs
@@ -277,7 +277,7 @@ impl SourceMapLookup {
 
 /// Joins a `url` to the `base` [`Url`], taking care of our special `~/` prefix that is treated just
 /// like an absolute url.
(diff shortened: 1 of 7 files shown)


crates/symbolicator-service/src/services/sourcemap_lookup.rs
```diff
@@ -277,7 +277,7 @@ impl SourceMapLookup {
 
 /// Joins a `url` to the `base` [`Url`], taking care of our special `~/` prefix that is treated just
 /// like an absolute url.
-fn join_url(base: &Url, url: &str) -> Option<Url> {
+pub fn join_url(base: &Url, url: &str) -> Option<Url> {
```
Swatinem, line 280: doesn’t need to be `pub` after all (though doesn’t hurt either)

loewenheim (approved):

PR #1139, opened 2023-04-13
Implement ZipArchive compressed symbols

tdesveaux:
Windows symstore.exe added support to compress symbols with /compress ZIP arg.
Rather than simply compressing the file with deflate, it adds it in a ZipArchive.

We use this as CAB compression has a limit on file size (2GB: https://learn.microsoft.com/en-us/archive/blogs/windows_installer_team/what-are-the-upper-limits-of-the-cab-file-format).

From symstore.exe help:
```
/compress [type]    When storing files, store compressed files on the server.
                        Has two optional arguments "CAB" and "ZIP" to specify the
                        compression type to use. Defaults to CAB compression if none is
                        specified. Ignored when storing pointers.
```

Not fluent in rust so don't hesitate with feedback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Add support for in-memory caching. ([#1028](https://github.com/getsentry/symbolicator/pull/1028))
 - Add --log-level argument to `symbolicli`. ([#1074](https://github.com/getsentry/symbolicator/pull/1074))
 - Resolve source context from embedded source links (mainly in Portable PDBs) ([#1103](https://github.com/getsentry/symbolicator/pull/1103), [#1108](https://github.com/getsentry/symbolicator/pull/1108))
+- Add single-file ZipArchive support to `maybe_decompress_file` ([#1139](https://github.com/getsentry/symbolicator/pull/1139))
 
 ### Fixes
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4450,6 +4450,7 @@ dependencies = [
  "tracing",
  "url",
  "uuid",
+ "zip",
  "zstd",
 ]
 
--- a/crates/symbolicator-service/Cargo.toml
+++ b/crates/symbolicator-service/Cargo.toml
@@ -51,6 +51,7 @@ uuid = { version = "1.0.0", features = ["v4", "serde"] }
 zstd = "0.12.1"
 data-encoding = "2.3.3"
 humantime = "2.1.0"
+zip = { version = "0.6.4", default-features = false, features = ["deflate"] }
 
 [dev-dependencies]
 insta = { version = "1.18.0", features = ["redactions", "yaml"] }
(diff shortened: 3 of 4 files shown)


Swatinem (approved):
lgtm. A test would be nice, though I don’t think we have any tests for all the other compression methods either, so I guess its fine.

Swatinem:
looks like our changelog bot is not working properly for outside PRs, so it does not present you with a suggestion. But I think adding a changelog entry will satisfy the CI check.

tdesveaux:
> lgtm. A test would be nice, though I don’t think we have any tests for all the other compression methods either, so I guess its fine.

I made one locally to test the change. But as there was none for other paths and I don't know how to add raw data for the test, I did not include it.
Here it is, if you can give pointers on how to correctly add it (it might also not be the greatest test, again I welcome feedback):
```rust
#[cfg(test)]
mod tests {
    use std::{
        fs::File,
        io::{self, Read, Seek},
        path::Path,
    };

    use tempfile::NamedTempFile;

    use super::maybe_decompress_file;

    #[test]
    fn test_zip_archive_debugfile() {
        let mut original_debugfile =
            File::open("/home/tdesveaux/getsentry/symbolicator/XAudio2_8.dll").unwrap();
        // Create from `symstore.exe add /f C:\Windows\System32\XAudio2_8.dll /s {path_to_store} /t {random_tag} /compress ZIP`
        let compressed_debugfile_path =
            Path::new("/home/tdesveaux/getsentry/symbolicator/XAudio2_8.dl_");
        let mut compressed_debugfile = File::open(compressed_debugfile_path).unwrap();

        let mut compressed_tempfile =
            NamedTempFile::new_in(compressed_debugfile_path.parent().unwrap()).unwrap();

        io::copy(&mut compressed_debugfile, &mut compressed_tempfile).unwrap();

        maybe_decompress_file(&mut compressed_tempfile).unwrap();

        compressed_tempfile.as_file().sync_all().unwrap();
        assert_eq!(
            original_debugfile.metadata().unwrap().len(),
            compressed_tempfile.as_file().metadata().unwrap().len()
        );

        {
            let mut original_debugfile_buffer = Vec::new();
            original_debugfile
                .read_to_end(&mut original_debugfile_buffer)
                .unwrap();

            compressed_tempfile.as_file().rewind().unwrap();
            let mut uncompressed_debugfile_buffer = Vec::new();
            compressed_tempfile
                .read_to_end(&mut uncompressed_debugfile_buffer)
                .unwrap();

            assert_eq!(original_debugfile_buffer, uncompressed_debugfile_buffer);
        }
    }
}
```



> looks like our changelog bot is not working properly for outside PRs, so it does not present you with a suggestion. But I think adding a changelog entry will satisfy the CI check.

How would I go about to do this?

I don't mind at all if you want to close this PR and include this change yourselves so it's better in-line with your processes.

Swatinem:
Just add a line to the https://github.com/getsentry/symbolicator/blob/master/CHANGELOG.md referencing this PR.

And the code looks very good even without any tests directly. Usually I would advocate for an end-to-end test, adding a fixture and testing that a stack trace referencing it produces the correct results.

Swatinem (approved):

PR #1162, opened 2023-04-26
Fix Hermes processing and improve path joining

Swatinem:
This fixes React Native Hermes bytecode processing, which would have otherwise been rejected because it uses 0-byte files.

Also fixes two more cases of path joining, in particular leading `./` paths, and `webpack:/` with just a single `/`.

#skip-changelog


crates/symbolicator-service/src/services/sourcemap_lookup.rs
```diff
@@ -1234,6 +1245,24 @@ mod tests {
...
+        // path with a leading dot
+        assert_eq!(
+            join_paths(
+                "app:///_next/static/chunks/pages/_app-569c402ef19f6d7b.js.map",
+                "./node_modules/@sentry/browser/esm/integrations/trycatch.js"
+            ),
+            "app:///_next/static/chunks/pages/node_modules/@sentry/browser/esm/integrations/trycatch.js"
+        );
+
+        // webpack with only a single slash
```
kamilogorek, line 1258: Is this really possible? `webpack://` is a custom protocol, so it should always be 2 slashes :|

kamilogorek (approved):

crates/symbolicator-service/src/services/sourcemap_lookup.rs
Swatinem: Well, there were counter examples in the A/B tests :-(

PR #1169, opened 2023-04-28
Hook up per-item expiration to moka

Swatinem:
Same as #1136 but now as a `canary` branch.

See https://github.com/moka-rs/moka/pull/248#issuecomment-1501038013

I would like to run this on canary for a bit to validate its effects.

#skip-changelog

### DIFF
--- a/crates/symbolicator-service/src/caching/memory.rs
+++ b/crates/symbolicator-service/src/caching/memory.rs
@@ -67,13 +67,42 @@ impl<T: CacheItemRequest> Clone for Cacher<T> {
     }
 }
 
+/// A struct implementing [`moka::Expiry`] that uses the [`InMemoryItem`] [`Instant`] as the explicit
+/// expiration time.
+struct CacheExpiration;
+fn saturating_duration_since(now: Instant, target: Instant) -> Option<Duration> {
+    // NOTE: the argument to `checked_duration_since` is the *earlier* instant.
+    // `target` is in the future, and `now` is the earlier one.
+    Some(target.checked_duration_since(now).unwrap_or_default())
+}
+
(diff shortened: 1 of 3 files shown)


crates/symbolicator-service/src/caching/memory.rs
```diff
@@ -67,13 +67,42 @@ impl<T: CacheItemRequest> Clone for Cacher<T> {
...
 }
 
+/// A struct implementing [`moka::Expiry`] that uses the [`InMemoryItem`] [`Instant`] as the explicit
+/// expiration time.
+struct CacheExpiration;
+fn saturating_duration_since(now: Instant, target: Instant) -> Option<Duration> {
+    // NOTE: the argument to `checked_duration_since` is the *earlier* instant.
+    // `target` is in the future, and `now` is the earlier one.
+    Some(target.checked_duration_since(now).unwrap_or_default())
+}
```
kamilogorek, line 77:
```suggestion
fn saturating_duration_since(current_time: Instant, target_time: Instant) -> Option<Duration> {
    // NOTE: `target_time` is of type `Instant`, as it was created and stored at an earlier time. This function will return a duration of `0` in case no expiry time has been reached yet.
    Some(target_time.checked_duration_since(current_time).unwrap_or_default())
}
```

kamilogorek (changes requested):
Slightly more understandable this way imo

crates/symbolicator-service/src/caching/memory.rs
Swatinem: 👍🏻 I reworded that a bit and put it as a doc comment.

kamilogorek (approved):

PR #1184, opened 2023-05-10
Make `Sourcemap` header matching case-insensitive

Swatinem:
Well, looks like the artifact lookup API can sometimes return an all-lowercase `sourcemap` header.

#skip-changelog


iambriccardo (approved):
LGTM

PR #1212, opened 2023-06-12
ci: Make docs publishing independent from the releases

kamilogorek:
I'm about 76% positive that it'll work on the first try.

#skip-changelog

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -2,7 +2,6 @@ minVersion: 0.23.1
 changelogPolicy: auto
 targets:
   - name: github
-  - name: gh-pages
   - name: registry
     apps:
       app:symbolicator:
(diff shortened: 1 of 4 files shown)


.craft.yml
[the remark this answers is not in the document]
kamilogorek: As we will publish docs using separate GitHub Action now, this would override our deploy with each release triggered. By removing target we skip the publishing step, and requiredNames is a list of files that need to be there in order for release to proceede, but after this change, the file is gone, so it'd effectively break releasing pipeline.

Makefile
```diff
@@ -44,12 +44,6 @@ docserver: .venv/bin/python
 	.venv/bin/mkdocs serve
 .PHONY: doc
 
-travis-upload-docs: docs
```
kamilogorek, line 47: This is some old, not used code that I cleaned up.

.github/workflows/build.yml
```diff
@@ -124,30 +124,3 @@ jobs:
         with:
           name: ${{ github.sha }}
           path: target/release/*-Windows-x86_64.exe
-
```
kamilogorek, line 127: We only want to run "test docs build" during `CI`, not `build`, as we dont care about uploading those artifacts anywhere anymore.

loewenheim (approved):

PR #1226, opened 2023-06-27
Introduce pipedream to symbolicator

mattgauntseo-sentry:
This introduces a new set of GoCD pipelines generated from jsonnet.

The pipelines produced are:

- deploy-symbolicator
- deploy-symbolicator-us
- deploy-symbolicator-monitor

deploy-symbolicator is a simple pipeline that triggers a deployment of symbolicator in all regions .
deploy-symbolicator-us is the equivalent and deploy-symbolicator that currently exist. deploy-symbolicator-monitor is a new pipeline that deploys to a single-tenant style environment.

You can learn more about this change here: https://www.notion.so/sentry/2023-Q2-S4S-Pipedream-c3752317f5334eb9bb0cd29df27beecc

**Question**: The current deployment pipeline has `service=symbolicator,deploy_if_production=true` for the deploy label selector, however s4s and ST just have `service=symbolicator`, is this needed for US deployments or can we use `service=symbolicator` for prod deployments?

#skip-changelog


Swatinem (approved):
lgtm. Not quite sure about the label you mentioned. Maybe the intention was to not double-deploy the canary when moving from the canary to production stage of deployment.

PR #1246, opened 2023-07-06
ref: Add special case for extracting name of rewritten async function in Dart lang

kamilogorek:
Unfortunately `dart_web` is setting the platform to `javascript` for this variant of the SDK, so we cannot use it to make any distinction between the incoming events.

It appears that there is no way to make a distinction between regular JS code and the scenario we're solving here :(
The only way that I found is indeed actually using `scope.startsWith('$async$')` - https://github.com/dart-lang/sdk/blob/main/pkg/compiler/lib/src/js_backend/namer.dart#L1845-L1866

@marandaneto could you please provide us with the most barebone repo case for this? Basically, a single async function that throws?
I don't want to commit 62k lines of fixtures for such a simple test case.

Fixes https://github.com/getsentry/symbolic/issues/791


marandaneto:
> @kamilogorek thank you. Will do it on Monday unless @blaugold beats me :D

https://manoeltest.sentry.io/issues/4303672281/?project=4505499492941824&query=is%3Aunresolved&statsPeriod=24h&stream_index=0
Download source maps: https://manoeltest.sentry.io/settings/projects/flutter/source-maps/artifact-bundles/751f0a63-3c32-5bbc-b78b-51b6208a6b23/
The file is still massive, but I'm afraid I can't generate any shorter unless you clean it up for all nonrelated properties of the file.

Screenshot 2023-07-09 at 15 02 21

This is a bare-bones flutter web app, I tried a simpler dart web app, but [release builds don't generate source maps, or they are deleted automatically afterward](https://github.com/dart-lang/build/issues/916).

blaugold:
I've compiled the program below with `dart compile js bin/main.dart -O4`:

```dart
void main() async {
  await Future(() {});
  throw StateError('');
}
```

This yields [out.js and out.map.js](https://github.com/getsentry/symbolicator/files/11995747/out.zip).

Hope that helps :)

marandaneto:
> 

Oh yeah, good call, I was using `webdev build`, totally forgot about calling `dart compile` by hand, thanks @blaugold

kamilogorek:
Lovely, thanks! Verified it works with the new fixtures.

PR #1248, opened 2023-07-10
Add new `ResolvedWith::Index`

Swatinem:
This was added in https://github.com/getsentry/sentry/pull/52080 for the files that were looked up with the new artifact bundle index.

#skip-changelog


loewenheim (approved):

Swatinem:
> Wait, so what happens right now when Symbolicator encounters `resolved_with: "index"`?

It probably fails parsing the request, though I didn’t see this happening yet.

PR #1255, opened 2023-07-17
Switch pipelines to pipedream

mattgauntseo-sentry:
This switches the symbolicator pipelines over to pipedream which will deploy to monitor (s4s) and then saas.

At the moment the pipeline has 2 manual steps, one for deploying canary and one for deploying primary, this is done for both s4s AND saas deployments, **should I remove these manual steps or leave them as is for now?**

NOTE: The old pipelines will still be available but will be moved to a new group "symbolicator-old" in case they are needed during the transition.

#skip-changelog


Swatinem (approved):
I believe keeping canary around sounds like a good idea

PR #1256, opened 2023-07-17
ci: add GoCD pipeline validation

joshuarli:
This replaces the lint action with actual validation against GoCD, which is miles better.

#skip-changelog


mattgauntseo-sentry (changes requested):
I'd still like to keep the `Render GoCD Pipelines with Jsonnet` step in the line-pipeline.yml file

joshuarli:
> I'd still like to keep the `Render GoCD Pipelines with Jsonnet` step in the line-pipeline.yml file

Ah, wasn't branched off from latest master, I see it now.

mattgauntseo-sentry (approved):

PR #1267, opened 2023-07-21
Improve file-stem trimming

Swatinem:
The original idea of file-stem trimming was to find both `foo.min.js` and `foo.map` with just one request, as the stem of those two is `foo`.

However the simplistic "split at first `.`" fails for situations when we have `main.$HASH.min.js`, where the `$HASH` part was being trimmed off. However the API would have way too many matches for just `main` in that case, and might thus not return the match that we are interested in.

Therefore, we are only trimming known suffixes, and keep other `.`-separated parts intact. This might increase the number of API requests we do, but should yield better results.

#skip-changelog


loewenheim (approved):
Seems like a good idea :)

PR #1274, opened 2023-07-28
Remove canary from non-us regions and remove canary manual step

mattgauntseo-sentry:
Removing the canary stage entirely for non-saas deployments makes sense since only the US deployment has any canary pods.

_*Question*_: I think it's safe to skip the "create sentry release" step in canary and just run the prod create sentry release, but could you take a look and confirm.

This PR also removed the manual step from the canary deployment in the US.

The result of these changes is that deployments will have one manual step per deployment (the deploy-prod step.

**In the future it might be worth considering changing this remaining manual step to only be needed for US regions**, even better would be to add automatic checks to the canary deployment in the US using the approaches listed here: https://www.notion.so/sentry/Block-deployments-on-canary-cd96781d331242c097f6959117d70ef3

#skip-changelog

### DIFF
--- a/gocd/generated-pipelines/rollback-symbolicator.yaml
+++ b/gocd/generated-pipelines/rollback-symbolicator.yaml
@@ -3,9 +3,9 @@ pipelines:
   rollback-symbolicator:
     display_order: 1
     environment_variables:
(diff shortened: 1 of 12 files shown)


Swatinem (approved):
lgtm!

might as well remove the manual approval for deploy to prod on s4s, as that instance has way too little traffic and is not actively monitored.

Swatinem:
> I think it's safe to skip the "create sentry release" step in canary and just run the prod create sentry release

yes that should work fine

PR #1290, opened 2023-08-18
Remove generated pipelines and update gocd-jsonnet

mattgauntseo-sentry:


mattgauntseo-sentry:
I'll merge this after hackweek

Issue #1334, opened 2023-11-29
Evaluate using LTO, Profile-Guided Optimization (PGO) and Post Link Optimization (PLO)


zamazan4ik:
Hi!

Recently I checked optimizations like Link-Time Optimization (LTO), Profile-Guided Optimization (PGO) and Post-Link Optimizations (PLO) improvements on multiple projects. The results are available [here](https://github.com/zamazan4ik/awesome-pgo/). According to the tests, all these optimizations can help with achieving better performance in many cases for many applications. I think trying to enable them for Symbolicator can be a good idea.

I already did some benchmarks and want to share my results here. Hopefully, they will be helpful.

## Test environment

* Fedora 39
* Linux kernel 6.5.12
* AMD Ryzen 9 5900x
* 48 Gib RAM
* SSD Samsung 980 Pro 2 Tib
* Compiler - Rustc 1.74
* Symbolicator version: the latest for now from the `master` branch on commit `ac127975e4649dc2f40b177cb98556e307aca26e`
* Disabled Turbo boost (for better results consistency across runs)

## Benchmark

For the benchmark purposes, I used [this](https://github.com/getsentry/symbolicator/blob/master/tests/wrk/README.md) WRK-based scenario for Minidump. As a minidump, I use [this](https://github.com/getsentry/symbolicator/blob/master/tests/fixtures/linux.dmp) Linux dump. WRK command is the same for all benchmarks and PGO/PLO training phases: `WRK_MINIDUMP="../tests/fixtures/linux.dmp" ./wrk --threads 10 --connections 50 --duration 30s --script ../tests/wrk/minidump.lua http://127.0.0.1:3021/minidump`. Before each WRK benchmark, I once run `cargo run -p process-event -- ../tests/fixtures/linux.dmp` as it's recommended.

All PGO and PLO optimizations are done with [cargo-pgo](https://github.com/Kobzol/cargo-pgo) (I highly recommend using this tool). For PLO phase I use [LLVM BOLT](https://github.com/llvm/llvm-project/blob/main/bolt/README.md) tool.

LTO is enabled by the following changes to the `Release` profile in the root `Cargo.toml` file:
```
[profile.release]
# For release builds, we do want line-only debug information to be able to symbolicate panic stack traces.
debug = 1
codegen-units = 1
lto = true
```
For all benchmarks, binaries are stripped with a `strip` tool.

[...]

PR #1337, opened 2023-11-30
Enable LTO for release builds

Swatinem:
Related to https://github.com/getsentry/symbolicator/issues/1334 CC @zamazan4ik 

This is so far just doing the first step, using 1 CGU and LTO to optimize the compiler output, without relying on instrumentation and gathering profiling.

Using cargo `--timings`, I get local bound times of ~260s of Rust build + ~640s of linking time (LLVM does code generation at link time with LTO if I understand things correctly) for a total of ~15 minutes.

The first part we can mostly solve with dependency caching, but not the last linking step. So our docker builds will most likely take in the order of 10 minutes.

I would actually like to land this and maybe let it run over the weekend in a canary, so we can compare that to the baseline.

#skip-changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,8 @@ default-members = ["crates/symbolicator", "crates/symbolicli", "crates/symsorter
 [profile.release]
 # For release builds, we do want line-only debug information to be able to symbolicate panic stack traces.
 debug = 1
+codegen-units = 1
+lto = true
 
 [profile.local]
 # For running a local symbolicator, we want the best of both worlds: a fast executable, with quick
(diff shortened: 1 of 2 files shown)


zamazan4ik:
> ~640s of linking time

Regarding linking time. I can suggest you switch to LLD or even the Mold linker. [Here](https://github.com/rui314/mold#readme) are some benchmarks between Gold vs LLD vs Mold.

If you decide to use LLD and want to optimize the CI pipeline as much as possible, I can suggest recompiling LLD with PGO - the results can be found [here](https://github.com/llvm/llvm-project/issues/63486#issuecomment-1607953028).

zamazan4ik:
@Swatinem also I can suggest enabling LTO for other Sentry projects like https://github.com/getsentry/relay

loewenheim (approved):

PR #1344, opened 2023-12-20
Update Axum & deps, Split LTO into separate profile

Swatinem:
This updates `axum/server` and the Sentry SDK to the latest versions along with doing a `cargo update` and applying nightly clippy fixes.

Also splits LTO builds into a separate profile, as doing local `release` builds is way too slow when using 1 CGU + LTO.
So instead, create a new profile for that and use it in docker builds.

#skip-changelog

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -116,9 +116,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.75"
+version = "1.0.77"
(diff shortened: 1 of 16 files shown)


Swatinem:
I opened https://github.com/tokio-rs/axum/issues/2459 as the initial port using `axum::serve` was deadlocking the tests. Using `axum_server` instead works just fine.

loewenheim (approved):

PR #1354, opened 2024-01-31
Add another Cache for accessing files in `ArtifactBundle`s

Swatinem:
When profiling (via `symbolicator-stress`) processing of javascript events, the majority of the time is spent extracting minified / sourcemap files from artifact bundles (which are essentially zip files).

We introduce yet another cache which can avoid this costly unzip process by keeping these unzipped files in memory.

For the above mentioned stresstest, this has lead to a throughput increase of 1000x. Though not surprisingly, the stresstest processes the exact same event over and over and has thus a 100% cache hit rate.

#skip-changelog

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,12 @@
 [workspace]
 resolver = "2"
 members = ["crates/*"]
-default-members = ["crates/symbolicator", "crates/symbolicli", "crates/symsorter", "crates/wasm-split"]
+default-members = [
+    "crates/symbolicator",
+    "crates/symbolicli",
+    "crates/symsorter",
+    "crates/wasm-split",
+]
 
 [profile.dev]
 # For debug builds, we do want to optimize for both debug-ability, and fast iteration times.
@@ -47,4 +52,4 @@ cpp_demangle = { git = "https://github.com/getsentry/cpp_demangle", branch = "se
 # symbolic = { path = "../../../symbolic/symbolic", features = ["common-serde", "debuginfo", "demangle", "minidump-serde", "symcache"] }
 # See also https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html
 # [patch.crates-io]
-# symbolic = { path = "../symbolic/symbolic" }
+# symbolic-debuginfo = { path = "../symbolic/symbolic-debuginfo" }
--- a/crates/symbolicator-js/src/lookup.rs
+++ b/crates/symbolicator-js/src/lookup.rs
@@ -28,7 +28,7 @@ use std::fmt::{self, Write};
 use std::fs::File;
 use std::io::{self, BufWriter};
 use std::sync::Arc;
-use std::time::{Duration, SystemTime};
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -1,7 +1,12 @@
 [workspace]
 resolver = "2"
 members = ["crates/*"]
-default-members = ["crates/symbolicator", "crates/symbolicli", "crates/symsorter", "crates/wasm-split"]
+default-members = [
+    "crates/symbolicator",
+    "crates/symbolicli",
+    "crates/symsorter",
+    "crates/wasm-split",
```
anonrig, line 8: Why don't we have other crates in this list?

crates/symbolicator-js/src/lookup.rs
```diff
@@ -555,10 +557,82 @@ struct IndividualArtifact {
     resolved_with: ResolvedWith,
 }
 
+type FileInBundleCacheInner = moka::sync::Cache<(RemoteFileUri, FileKey), CachedFileEntry>;
+
+/// A cache that memoizes looking up files in artifact bundles.
+#[derive(Clone)]
+pub struct FileInBundleCache {
```
anonrig, line 564: Can we move this to an another file? It's hard to keep track already.

crates/symbolicator-js/src/lookup.rs
```diff
@@ -555,10 +557,82 @@ struct IndividualArtifact {
...
+    /// Creates a new `FileInBundleCache` with a maximum size of 2GiB and
+    /// idle time of 1h.
+    pub fn new() -> Self {
+        const GIGS: u64 = 1 << 30;
+        let cache = FileInBundleCacheInner::builder()
+            .max_capacity(2 * GIGS)
+            .time_to_idle(Duration::from_secs(60 * 60))
+            .name("file-in-bundle")
+            .weigher(|_k, v| {
+                let content_size = v
```
anonrig, line 584: Can you add a comment about the behavior of `weigher`?

crates/symbolicator-js/src/service.rs
```diff
@@ -44,8 +45,11 @@ impl SourceMapService {
             in_memory,
         ));
 
+        let files_in_bundles = FileInBundleCache::new();
+
         Self {
             objects,
+            files_in_bundles,
```
anonrig, line 52: Since we do the same thing on line 54:

```suggestion
            files_in_bundles: FileInBundleCache::new(),
```

Cargo.toml
Swatinem: these listed here are the ones that we `cargo build` and ship. all others are just internal.
We run CI and `cargo test` with `--workspace` so everything is tested.

anonrig:
@Swatinem Can you update the description to include any context about people who doesn't know anything about our talk/pair coding session? If you could include benchmark results, what are we doing to improve etc. it would be good for future documentation.

anonrig (approved):
🚀

PR #1374, opened 2024-02-16
chore(ci): Revert upload-artifact version

azaslavsky:


.github/workflows/build.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           mv target/release/symsorter build/symsorter-Linux-x86_64
           mv target/release/symbolicli build/symbolicli-Linux-x86_64
 
-      - uses: actions/upload-artifact@v4.3.1
+      - uses: actions/upload-artifact@v3
```
hubertdeng123, line 44: Looks like this was v3.1.1 [here](https://github.com/getsentry/symbolicator/pull/1366/files), let's keep it the same version as before?

hubertdeng123:
To give context, this change is needed since it's breaking the self-hosted release. There is a breaking change in upload-artifact v4 here with uploading artifacts with the same name
https://github.com/actions/toolkit/tree/main/packages/artifact

hubertdeng123 (approved):
lgtm pending comment on maintaing version 3.1.1

.github/workflows/build.yml
azaslavsky: Done.

PR #1375, opened 2024-02-16
Make stresstest observability configurable

Swatinem:


crates/symbolicator-stress/src/main.rs
```diff
@@ -50,15 +62,15 @@ fn main() -> Result<()> {
     let service_config = SymbolicatorConfig::get(config_path.as_deref())?;
 
     let mut logging_guard = logging::init(logging::Config {
-        // TODO: actually make this configurable
         backtraces: true,
-        tracing: true,
-        sentry: true,
-        metrics: true,
+        tracing: cli.tracing,
```
anonrig, line 66: Does this change the default value?

loewenheim (approved):

crates/symbolicator-stress/src/main.rs
Swatinem: it does, yes. but this is just a stresstest for running locally, it does not matter what is the default or not :-)

PR #1376, opened 2024-02-16
Increase percentage of sampled transactions

Swatinem:
You can take a look at those here: https://sentry.my.sentry.io/organizations/sentry/performance/?landingDisplay=all&project=5&query=&statsPeriod=14d


loewenheim (approved):

Swatinem:
> What is the motivation?

just because, no reason in particular.

I didn’t remember that we had such a low sample rate, I thought we were sampling a ton more transactions. I’m trying to figure out the overhead of all this instrumentation.

PR #1382, opened 2024-02-20
Create a `tracing` layer doing JSON output and Sentry at once

Swatinem:
Using composition to chain the JSON output and the Sentry layer means that both layers are capturing span attributes, duplicating that work.

This pretty much copy-pasted the code of both these layers and combines them, thus avoiding doing all the `SpanAttribute` capturing twice.

---

TBH, I don’t think this might be fully worth it, as copy-pasting all this code over looks really horrible :-(


loewenheim (commented):
Oof. I'm honestly not a fan.

PR #1397, opened 2024-03-01
RIP `BundleIndex`

Swatinem:
This feature was never fully rolled out, and we are now removing all the code from the Sentry codebase, so lets also remove it here.


Swatinem:
See https://github.com/getsentry/sentry/pull/66125 for the removal on the Sentry side.

loewenheim (approved):

PR #1409, opened 2024-03-15
Use upload-artifact v3

hubertdeng123:
reverts https://github.com/getsentry/symbolicator/pull/1380

upload-artifact v4 introduces a blocking change where artifact files cannot be uploaded with the same name. In order to get us on the newer version, we need to ensure the artifact file names that are uploaded have different names.


chadwhitacre (approved):

chadwhitacre:
Force-merging due to timezones to unblock the self-hosted release.

PR #1421, opened 2024-03-28
feat(proguard): add signature support and remove parameters from JvmFrame

viglia:
The parameters list will be computed from the signature passed. 
The signature itself will be deobfuscated.

This, at the moment, assumes that the `signature` passed is always a [JDI Type Signatures](https://docs.oracle.com/javase/8/docs/jdk/api/jpda/jdi/com/sun/jdi/doc-files/signature.html), which is always the case for the profiling use case.
If we ever predict to use this (param mapping deobfuscation) in scenarios where we might receive a signature in normal form (with Java types), we'll have to add some further support to detect the kind of signature and apply the proper signature parsing.

### DIFF
--- a/crates/symbolicator-proguard/src/java.rs
+++ b/crates/symbolicator-proguard/src/java.rs
@@ -1,0 +1,229 @@
+use std::collections::HashMap;
+use std::sync::OnceLock;
+
+fn java_base_types() -> &'static HashMap<String, String> {
+    static JAVA_BASE_TYPES: OnceLock<HashMap<String, String>> = OnceLock::new();
+    JAVA_BASE_TYPES.get_or_init(|| {
+        // https://docs.oracle.com/javase/8/docs/jdk/api/jpda/jdi/com/sun/jdi/doc-files/signature.html
+        HashMap::from([
+            ("Z".to_string(), "boolean".to_string()),
+            ("B".to_string(), "byte".to_string()),
+            ("C".to_string(), "char".to_string()),
+            ("S".to_string(), "short".to_string()),
+            ("I".to_string(), "int".to_string()),
+            ("J".to_string(), "long".to_string()),
+            ("F".to_string(), "float".to_string()),
+            ("D".to_string(), "double".to_string()),
+            ("V".to_string(), "void".to_string()),
+        ])
+    })
+}
+
+fn byte_code_type_to_java_type(
+    byte_code_type: String,
+    mappers: &[&proguard::ProguardMapper],
+) -> String {
+    if byte_code_type.is_empty() {
+        return "".to_string();
+    }
+    let mut chrs = byte_code_type.chars();
+    let token = chrs.next().unwrap();
+    if let Some(ty) = java_base_types().get(&token.to_string()) {
+        return ty.clone();
+    } else if token == 'L' {
+        // invalid signature
+        let l = chrs.clone().last();
+        if l.is_none() || l.unwrap() != ';' {
+            return "".to_string();
+        }
+        chrs.next_back(); // remove final `;`
+        let obfuscated = chrs.collect::<String>().replace('/', ".");
+        // if mappers.len() > 0 {}
+        for mapper in mappers {
+            if let Some(mapped) = mapper.remap_class(&obfuscated) {
+                return mapped.to_string();
+            }
+        }
+        return obfuscated;
+    } else if token == '[' {
+        return format!(
+            "{}[]",
+            byte_code_type_to_java_type(chrs.collect::<String>(), mappers)
+        );
+    }
+    byte_code_type
+}
+
+// parse_obfuscated_signature will parse an obfuscated signatures into parameter
+// and return types that can be then deobfuscated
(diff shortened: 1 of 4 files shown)


crates/symbolicator-proguard/src/java.rs
```diff
@@ -0,0 +1,229 @@
...
+    let token = chrs.next();
+    if token.unwrap_or_default() != '(' {
+        return None;
+    }
+
+    let sig = chrs.collect::<String>();
+    let split_sign = sig.rsplitn(2, ')').collect::<Vec<&str>>();
+    if split_sign.len() != 2 {
+        return None;
+    }
```
Swatinem, line 71: Isn’t there also an `rsplit` that returns a `Option<(&str,&str)>`?

Other than that, you might want to use `strip_prefix` to remove that leading `(`.

crates/symbolicator-proguard/src/java.rs
```diff
@@ -0,0 +1,229 @@
...
+    if return_type.is_empty() {
+        return None;
+    }
+
+    let mut types: Vec<String> = Vec::new();
+    let mut tmp_buf: Vec<char> = Vec::new();
+
+    let mut param_chrs = parameter_types.chars();
+    while let Some(token) = param_chrs.next() {
+        if java_base_types().contains_key(&token.to_string()) {
```
Swatinem, line 84: Same thing, you can tweak the order of the `if` to check for trivial cases earlier.

In general `&X.to_string()` is a big red flag, and not needed with my suggestion up above.

crates/symbolicator-proguard/src/java.rs
```diff
@@ -0,0 +1,229 @@
...
+            types.push(tmp_buf.iter().collect());
+            tmp_buf.clear();
+        } else if token == '[' {
+            tmp_buf.push('[');
+        } else {
+            tmp_buf.clear();
+        }
+    }
+
+    Some((types, return_type.to_string()))
```
Swatinem, line 112: `return_type` could just be a `&str`.

crates/symbolicator-proguard/src/java.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// returns a tuple where the first element is the list of the function
+// parameters and the second one is the return type
+pub fn deobfuscate_bytecode_signature(
+    signature: &str,
+    mappers: &[&proguard::ProguardMapper],
+) -> Option<(Vec<String>, String)> {
+    if let Some((parameter_types, return_type)) = parse_obfuscated_bytecode_signature(signature) {
```
Swatinem, line 121:
```suggestion
    let (parameter_types, return_type) = parse_obfuscated_bytecode_signature(signature)?;
```

You can use the `?` operator to propagate `None`.

crates/symbolicator-proguard/src/java.rs
```diff
@@ -0,0 +1,229 @@
...
+        let mut parameter_java_types: Vec<String> = Vec::with_capacity(parameter_types.len());
+        for parameter_type in parameter_types {
+            let new_class = byte_code_type_to_java_type(parameter_type, mappers);
+            parameter_java_types.push(new_class);
+        }
```
Swatinem, line 126:
```suggestion
        let parameter_java_types: Vec<_> = parameter_types.into_iter().map(|parameter_type|
            byte_code_type_to_java_type(parameter_type, mappers)
        ).collect();
```

crates/symbolicator-proguard/src/java.rs
```diff
@@ -0,0 +1,229 @@
...
+        let return_java_type = byte_code_type_to_java_type(return_type, mappers);
+
+        return Some((parameter_java_types, return_java_type));
+    }
+
+    None
+}
+
+// format_signature formats the types into a human-readable signature
+pub fn format_signature(types: &Option<(Vec<String>, String)>) -> String {
```
Swatinem, line 137:
```suggestion
pub fn format_signature(parameter_types: &[String], return_type: &str) -> String {
```

Returning an empty string on `None` is kind of an anti-pattern as well. Might as well do that fallback in the caller, or just propagate a proper `Option` instead of using an empty string as sentinel value.

Swatinem (commented):
there is quite some comments here, and things to improve.

In general I feel like this whole parsing of signatures might be better suited for the `proguard` crate itself, rather than doing it in Symbolicator here.

crates/symbolicator-proguard/src/java.rs
viglia: @Swatinem `rsplitn` is restricted to returning at most **n** items. `rsplit` does not ensure that.

Also, here we want to check specifically if the first token is not a `(` as it would be an invalid signature. And with `token.unwrap_or_default() != '('` we're both checking and consuming so no need for removing that later.

viglia:
@Swatinem thanks for the feedback!

I'm closing this PR as I'm moving this over to `rust-proguard` [here](https://github.com/getsentry/rust-proguard/pull/35)

viglia:
@Swatinem thanks for the feedback!

I'm closing this PR as I'm moving this over to `rust-proguard` [here](https://github.com/getsentry/rust-proguard/pull/35)

PR #1425, opened 2024-04-04
feat(proguard): add support for frame signature

viglia:
The signature will be properly converted from bytecode types to java types, plus the method parameters will be used to map the frame in absence of line numbers.

### DIFF
--- a/crates/symbolicator-proguard/src/symbolication.rs
+++ b/crates/symbolicator-proguard/src/symbolication.rs
@@ -193,7 +193,7 @@ impl ProguardService {
                 }
                 None
             });
-        let mut params = deobfuscated_signature
+        let params = deobfuscated_signature
             .as_ref()
             .map(|sig| sig.parameters_types().collect::<Vec<_>>().join(","));
         let stack_frame = frame
@@ -202,7 +202,7 @@ impl ProguardService {
                 proguard::StackFrame::new(&frame.module, &frame.function, lineno as usize)
             })
             .or_else(|| {
-                params.as_mut().map(|p| {
+                params.as_ref().map(|p| {
                     proguard::StackFrame::with_parameters(
                         &frame.module,
                         &frame.function,
--- a/crates/symbolicator-proguard/src/symbolication.rs
+++ b/crates/symbolicator-proguard/src/symbolication.rs
@@ -193,9 +193,9 @@ impl ProguardService {
                 }
                 None
             });
-        let mut params = deobfuscated_signature.as_ref().map(|sig| {
-            String::from_iter(sig.parameters_types().map(|param| format!("{},", param)))
-        });
+        let mut params = deobfuscated_signature
+            .as_ref()
+            .map(|sig| sig.parameters_types().collect::<Vec<_>>().join(","));
         let stack_frame = frame
             .lineno
             .map(|lineno| {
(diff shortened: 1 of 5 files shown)


crates/symbolicator-proguard/src/symbolication.rs
```diff
@@ -241,6 +257,12 @@ impl ProguardService {
                                 mapped_frame.in_app = Some(true);
                             }
                         }
+                        // if there is a signature that has been deobfuscated,
+                        // add it to the mapped frame
+                        if deobfuscated_signature.is_some() {
+                            mapped_frame.signature =
+                                Some(deobfuscated_signature.as_ref().unwrap().format_signature());
+                        }
```
Swatinem, line 265:
```suggestion
                        if let Some(signature) = &deobfuscated_signature {
                            mapped_frame.signature = Some(signature.format_signature());
                        }
```

crates/symbolicator-proguard/src/symbolication.rs
```diff
@@ -184,19 +184,35 @@ impl ProguardService {
...
+                        return Some(deobfuscated_signature);
+                    }
+                }
+                None
+            });
+        let params = deobfuscated_signature
+            .as_ref()
+            .map_or(String::new(), |sig| {
+                String::from_iter(sig.parameters_types().map(|param| format!("{},", param)))
+            });
```
Swatinem, line 200: Might be worth figuring out how to avoid allocating intermediate `String`s here.
I would also rather propagate the `None` here, instead of having an empty `String`.

Down below you can then rather match on `params`, as that is the thing you are using.

You are using the `deobfuscated_signature` here. I thought that the parameters in the proguard file are obfuscated as well? Or are those already deobfuscated?

This could really use a small integration test with a minimized proguard file.

crates/symbolicator-proguard/src/symbolication.rs
```diff
@@ -184,19 +184,35 @@ impl ProguardService {
...
         let stack_frame = frame
             .lineno
             .map(|lineno| {
                 proguard::StackFrame::new(&frame.module, &frame.function, lineno as usize)
             })
             .or_else(|| {
-                frame.parameters.as_ref().map(|params| {
-                    proguard::StackFrame::with_parameters(
+                if deobfuscated_signature.as_ref().is_some() {
```
Swatinem, line 207:
```suggestion
                if let Some(params) = &params {
```

crates/symbolicator-proguard/src/symbolication.rs
viglia: The parameters stored in the returned struct `deobfuscated_signature` are already converted to proper java types and deobfuscated (when necessary). Here we really just need to concatenate them.

crates/symbolicator-proguard/src/symbolication.rs
```diff
@@ -184,17 +184,30 @@ impl ProguardService {
...
+        });
         let stack_frame = frame
             .lineno
             .map(|lineno| {
                 proguard::StackFrame::new(&frame.module, &frame.function, lineno as usize)
             })
             .or_else(|| {
-                frame.parameters.as_ref().map(|params| {
+                params.as_mut().map(|p| {
+                    p.pop();
```
Swatinem, line 206: It took me quite some time to figure out that you are popping off the trailing `,` here. Please do that in the initial statement that creates `params`. It doesn’t need to be `mut` so you can mutate it 10 lines after creating it, where its very hard to follow whats actually happening.

viglia: @Swatinem I've actually removed that already but I haven't pushed yet.

I'm cleaning the code. I'll ask for a new review once I'm done.

Swatinem (approved):

PR #1430, opened 2024-04-08
fix(proguard): add signature when both the frame remap and the class remap fail

viglia:


crates/symbolicator-proguard/src/symbolication.rs
```diff
@@ -292,6 +292,15 @@ impl ProguardService {
...
+        // but add the signature if we received one and we were
+        // able to translate/deobfuscate it
+        if let Some(signature) = &deobfuscated_signature {
+            //frame.signature = Some(signature.format_signature());
+            return vec![JvmFrame {
+                signature: Some(signature.format_signature()),
+                ..frame.clone()
+            }];
+        }
         vec![frame.clone()]
```
Swatinem, line 304:
```suggestion
        // but add the signature if we received one and we were
        // able to translate/deobfuscate it
        let mut frame = frame.clone();
        if let Some(signature) = &deobfuscated_signature {
            frame.signature = Some(signature.format_signature());
        }
        vec![frame]
```

loewenheim (approved):

PR #1459, opened 2024-05-09
ref: use buildkit to build docker image

asottile-sentry:
faster caching, pull only occurs when needed


asottile-sentry:
this also simplifies the cloudbuild a bunch -- which I'm going to follow up with moving to artifact registry from deprecated gcr

anonrig (approved):

PR #1470, opened 2024-05-22
chore: remove unused dependencies

anonrig:
I've given @boshen 's cargo-shear a try: https://github.com/Boshen/cargo-shear

#skip-changelog


anonrig:
> The `insta` removals are false positives. I think you're running into this: https://github.com/Boshen/cargo-shear?tab=readme-ov-file#ignore-false-positives

I just removed false positives for now

loewenheim (approved):

PR #1503, opened 2024-07-30
all-repos: update actions/upload-artifact to v4

joshuarli:
Committed via https://github.com/asottile/all-repos


loewenheim (approved):

joshuarli (commented):
hmm will need to take a closer look at how these artifacts are used, i have a hunch this will break craft and release assets

https://github.com/getsentry/symbolicator/actions/runs/10064818143 currently v3 just puts everything under one zip

PR #1519, opened 2024-09-16
Changed test to no longer use public microsoft symbol server.

tobias-wilfert:
Addressing [#1004](https://github.com/getsentry/symbolicator/issues/1004) by changing test to make use of local fixture instead of public microsoft symbol server.

fix: #1004


loewenheim (approved):

loewenheim (approved):
Looks good to me. It's still testing the thing it's supposed to be but without Microsoft dependency.

PR #1523, opened 2024-09-19
WIP:  Share negative cache entries across Symbolicators

tobias-wilfert:
Still some work that needs to be done:

- [ ] Write some tests to ensure the logic works
- [ ] Discuss if there is a more elegant way to write the code
- [ ] Make sure that the negative files are evicted from the shared file after 1 hour
- [ ] See if the look up logic is still correct, might not work anymore, worried about this clashing with new logic:
https://github.com/getsentry/symbolicator/blob/cc2f915309ecb437b3558792edc85c3b13bd645c/crates/symbolicator-service/src/caching/shared_cache/mod.rs#L676C1-L676C79

Fix: [#422](https://github.com/getsentry/team-ingest/issues/422)


tobias-wilfert:
Talked with Sebastian about this and might halt the work on this for now, as getting the proper eviction mechanic inside the shared cache turned out to be a bigger problem than expected.

PR #1614, opened 2025-02-05
instr(pdb): Upgrade warning to error

jjbayer:
Temporarily update warning about .NET with Windows PDB to an error so we can aggregate by `request.scope` on the sentry error to see how many projects are affected.

ref: https://github.com/getsentry/symbolic/issues/871

#skip-changelog

### DIFF
--- a/crates/symbolicator-native/src/caches/ppdb_caches.rs
+++ b/crates/symbolicator-native/src/caches/ppdb_caches.rs
@@ -125,7 +125,7 @@ fn write_ppdb_cache(file: &mut File, object_handle: &ObjectHandle) -> CacheEntry
     let ppdb_obj = match object_handle.object() {
         Object::PortablePdb(ppdb_obj) => ppdb_obj,
         _ => {
-            tracing::warn!("Trying to symbolicate a .NET event with a non-PPDB object file");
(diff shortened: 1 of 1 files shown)


tobias-wilfert (approved):

Dav1dde (commented):
Should we instead have a metric for this?

jjbayer:
> Should we instead have a metric for this?

@Dav1dde I want to enumerate the projects that are affected. Currently we're seeing this warning in our logs ~1000 / day, so volume should be manageable.

Dav1dde:
@jjbayer 👍 I assume the project id gets added through the context/scope from the caller?

jjbayer:
> I assume the project id gets added through the context/scope from the caller?

Yes, `configure_scope` attaches tags to the sentry scope https://github.com/getsentry/symbolicator/blob/1f3a74e693546667634d63b2a9bdd4a2f6691e14/crates/symbolicator-native/src/caches/ppdb_caches.rs#L123-L128

and the "scope" tag contains the project ID IIUC:

https://github.com/getsentry/symbolicator/blob/1f3a74e693546667634d63b2a9bdd4a2f6691e14/crates/symbolicator-service/src/objects/data_cache.rs#L83-L86

PR #1621, opened 2025-02-14
fix: rectify symsorter `--ignore-errors` handling for ZIP archives.

supervacuus:
I recently observed a behavior in `symsorter` when running with `--ignore-errors` that is probably unintentional:

A ZIP archive it detected in the input path failed to read, leading to immediate termination, printing an error message in the top-level error logger, and unhelpful information that suggested I could use `--ignore-errors`.

Of course, what I really would expect in such a situation is the following:

* `symsorter` prints the error with the context of the ZIP archive path for which the error happens
* `symsorter` continues to process all other files
* `symsorter` does not suggest to use a flag which was already enabled
* `symsorter` returns a 0

This PR

* moves CLI parsing to `main()` so that the top-level error logger doesn't suggest using `--ignore-errors` when it is already in effect (because, in this case, it must have an error that cannot be ignored for some reason, because it shouldn't propagate to the top otherwise),
* extracts processing of ZIP archives into a separate function so it hides `--ignore-errors`-handling for ZIP archiving as an implementation detail,
* introduces a catch-all result scope for the ZIP archive processing (which has three ZIP-related function invocations that could fail besides `process_file()`) and 
* adds `--ignore-errors`-handling in that function analog to how the `maybe_ignore_error` macro handles it in `process_file()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Various fixes & improvements
+
+fix: rectify symsorter `--ignore-errors` handling for ZIP archives. ([#11621](https://github.com/getsentry/symbolicator/pull/1621)
+
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


loewenheim (approved):
Very nice, thank you for this!

PR #1651, opened 2025-03-17
fix(cfi): Reduce hallucinated frames during stack scanning

Dav1dde:
By validating a potential instruction against the loaded CFI module we can filter out bogus/hallucinated frames.

Same change for Symbolic's `minidump-stackwalk`: https://github.com/getsentry/symbolic/pull/904
Relates to: https://github.com/getsentry/sentry/issues/85336


loewenheim (approved):
Very nice catch!

PR #1682, opened 2025-04-09
fix(download): Flush the destination file, after download

Dav1dde:
Just in case, we should flush the data, this should happen automatically due to the file being closed, but there was instance that could be observed in a test where this was seemingly not the case.

In general it's the callers responsibility to flush the contents, except in the multi stream case, we need to manually flush, because we open a separate file handle.


iambriccardo (approved):
LGTM, are we okay to not `fsync`?

loewenheim (approved):

Dav1dde:
> LGTM, are we okay to not fsync?

yeah that's fine, we don't need to get it forced on disk, we just need to get it flushed.

PR #1691, opened 2025-04-15
fix(ci): Bump ubuntu to 22.04

hubertdeng123:
See https://github.com/getsentry/symbolicator/actions/runs/14475232477/job/40600459400

``` 
--
This is a scheduled Ubuntu 20.04 retirement. Ubuntu 20.04 LTS runner will be removed on 2025-04-15. For more details, see https://github.com/actions/runner-images/issues/11101
 

<br class="Apple-interchange-newline">
```

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   linux:
     name: Build Binary on Linux
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     steps:
       - name: Checkout sources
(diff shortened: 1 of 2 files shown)


Dav1dde (commented):
We did forget to upgrade these, but need to be careful, the built binary must be compatible with the used docker image's glibc version.

We know ubuntu 22 works, so I'd settle on that for now.

hubertdeng123:
> We know ubuntu 22 works, so I'd settle on that for now.
Gotcha, will change that then

Dav1dde (approved):

PR #1716, opened 2025-06-12
ci: push nightly images to dockerhub

aldy505:
Nightly images should be pushed to DockerHub. This is the only repository that don't do that.

#skip-changelog

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -242,7 +242,7 @@ jobs:
         run: |
           IMAGE_URL="ghcr.io/getsentry/symbolicator:${{ github.sha }}"
           docker pull "$IMAGE_URL"
-          docker login --username=sentrybuilder --password ${{ secrets.DOCKER_HUB_RW_TOKEN }}
+          docker login --username=sentrybuilder --password-stdin <<< "${{ secrets.DOCKER_HUB_RW_TOKEN }}"
 
           # We push 3 tags to Dockerhub:
           # first, the full sha of the commit
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -241,7 +241,6 @@ jobs:
         shell: bash
         run: |
           IMAGE_URL="ghcr.io/getsentry/symbolicator:${{ github.sha }}"
-          docker pull "$IMAGE_URL"
           docker login --username=sentrybuilder --password-stdin <<< "${{ secrets.DOCKER_HUB_RW_TOKEN }}"
 
           # We push 3 tags to Dockerhub:
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -232,7 +232,7 @@ jobs:
   publish-to-dockerhub:
     name: Publish Symbolicator to DockerHub
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/image.yml
```diff
@@ -228,3 +228,29 @@ jobs:
...
+    needs:
+      - assemble
+    steps:
+      - uses: actions/checkout@v4
+      - name: Push built docker image
+        shell: bash
+        run: |
+          IMAGE_URL="ghcr.io/getsentry/symbolicator:${{ github.sha }}"
+          docker pull "$IMAGE_URL"
+          docker login --username=sentrybuilder --password ${{ secrets.DOCKER_HUB_RW_TOKEN }}
```
BYK: Use the safer pattern for logging in please:
```
docker login --username '${{ github.actor }}' --password-stdin ghcr.io <<< "$GHCR_TOKEN"
```

.github/workflows/image.yml
```diff
@@ -228,3 +228,29 @@ jobs:
           docker buildx imagetools create -t "${TARGET_IMAGE}:latest" \
             "${GHCR_IMAGE}:arm64-${{ github.sha }}" \
             "${GHCR_IMAGE}:amd64-${{ github.sha }}"
+
+  publish-to-dockerhub:
+    name: Publish Symbolicator to DockerHub
+    runs-on: ubuntu-latest
+    if: ${{ (github.ref_name == 'main') }}
```
Dav1dde: There is no `main` branch here.

aldy505: You look too fast lol

.github/workflows/image.yml
```diff
@@ -228,3 +228,29 @@ jobs:
...
+    if: ${{ (github.ref_name == 'master') }}
+    needs:
+      - assemble
+    steps:
+      - uses: actions/checkout@v4
+      - name: Push built docker image
+        shell: bash
+        run: |
+          IMAGE_URL="ghcr.io/getsentry/symbolicator:${{ github.sha }}"
+          docker pull "$IMAGE_URL"
```
tobias-wilfert: I don't think we need to do this:

> ` docker pull "$IMAGE_URL"`

 for `docker buildx imagetools` also doing this might cause issues with the multi-arch image since we would pull only one architecture.

aldy505: ah okay

tobias-wilfert: We had issues with that in Sentry-Cli in the past: https://github.com/getsentry/sentry-cli/pull/2418

.github/workflows/image.yml
```diff
@@ -228,3 +228,28 @@ jobs:
           docker buildx imagetools create -t "${TARGET_IMAGE}:latest" \
             "${GHCR_IMAGE}:arm64-${{ github.sha }}" \
             "${GHCR_IMAGE}:amd64-${{ github.sha }}"
+
+  publish-to-dockerhub:
+    name: Publish Symbolicator to DockerHub
+    runs-on: ubuntu-latest
+    if: ${{ (github.ref_name == 'master') }}
```
BYK: And event is push?

PR #1719, opened 2025-06-24
feat(logging): Send tracing events at or above INFO to Sentry as logs instead of breadcrumbs

lcian:
- Bumps the SDK to 0.41.0
- Sets up all the Sentry `tracing` layers to send logs when the event level is at or above INFO.
- We're using e.g. `start_session` throughout the repo, therefore we need the `release-health` feature of `sentry` which brings in some unnecessary dependencies. See https://github.com/getsentry/sentry-rust/issues/805#issuecomment-2879047123 and https://github.com/rust-lang/cargo/issues/10801 for more information.
- Note that this change will create new issues in our Symbolicator project in Sentry. After deploying this we should archive the old ones with `<unknown>` as the title.

### DIFF
--- a/crates/symbolicator-stress/Cargo.toml
+++ b/crates/symbolicator-stress/Cargo.toml
@@ -10,11 +10,18 @@ axum = "0.7.2"
 clap = { version = "4.3.2", features = ["derive"] }
 futures = "0.3.12"
 humantime = "2.0.1"
-sentry = { version = "0.41.0", features = [
+sentry = { version = "0.41.0", default-features = false, features = [
+    # default features, except `release-health` is disabled
+    "backtrace",
+    "contexts",
+    "debug-images",
+    "panic",
+    "transport",
+    # additional features
     "anyhow",
     "tracing",
     "logs",
-] }
+]}
 serde = { version = "1.0.137", features = ["derive"] }
 serde_json = "1.0.81"
 serde_yaml = "0.9.14"
--- a/crates/symbolicator-js/Cargo.toml
+++ b/crates/symbolicator-js/Cargo.toml
@@ -22,7 +22,16 @@ reqwest = { workspace = true, features = [
     "stream",
     "hickory-dns",
 ] }
-sentry = { version = "0.41.0", features = ["tracing"] }
+sentry = { version = "0.41.0", default-features = false, features = [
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -38,7 +38,8 @@ codegen-units = 256
 
 [patch.crates-io]
 # This patch adds an `ip_filter` able to filter out internal IP addresses in DNS resolution
-reqwest = { git = "https://github.com/getsentry/reqwest", branch = "restricted-connector" }
+#reqwest = { git = "https://github.com/getsentry/reqwest", branch = "restricted-connector" }
+reqwest = { git = "https://github.com/lcian/reqwest", branch = "lcian/deps/bump-rustls-pki-types" }
```
lcian: Temporarily using my branch to run CI with it, then I'll remove this line

crates/symbolicator-stress/Cargo.toml
```diff
@@ -10,10 +10,10 @@ axum = "0.7.2"
 clap = { version = "4.3.2", features = ["derive"] }
 futures = "0.3.12"
 humantime = "2.0.1"
-sentry = { version = "0.34.0", features = [
+sentry = { version = "0.41.0", features = [
     "anyhow",
-    "debug-images",
```
lcian, line 15: `debug-images` is a default feature

Cargo.lock
```diff
@@ -4050,30 +4287,45 @@ checksum = "388a1df253eca08550bef6c72392cfe7c30914bf41df5269b68cbd6ff8f570a3"
...
-version = "0.34.0"
+version = "0.41.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5484316556650182f03b43d4c746ce0e3e48074a21e2f51244b648b6542e1066"
+checksum = "507ac2be9bf2da56c831da57faf1dadd81f434bd282935cdb06193d0c94e8811"
 dependencies = [
  "httpdate",
  "native-tls",
  "reqwest",
+ "sentry-actix",
```
Litarnus, line 4297: Do we need `actix` here? Seems like it pulls in a lot of additional dependencies

loewenheim (approved):

Dav1dde (approved):

PR #1729, opened 2025-07-09
ref(gocd): use console script entry points

mchen-sentry:
Updating to use console script entry points.

Removed unused google cloud build check script.

Please see: https://github.com/getsentry/devinfra-deployment-service/pull/698

### DIFF
--- a/gocd/templates/bash/deploy.sh
+++ b/gocd/templates/bash/deploy.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-eval $(regions-project_env-vars --region="${SENTRY_REGION}")
+eval $(regions-project-env-vars --region="${SENTRY_REGION}")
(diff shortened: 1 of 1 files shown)


gocd/templates/bash/cloudbuild-checks.sh
```diff
@@ -1,6 +0,0 @@
-#!/bin/bash
-
-/devinfra/scripts/checks/googlecloud/checkcloudbuild.py \
-  ${GO_REVISION_SYMBOLICATOR_REPO} \
-  sentryio \
-  "us-central1-docker.pkg.dev/sentryio/symbolicator/image"
```
Dav1dde, line 6: Why is this deleted?

Dav1dde (approved):

PR #1730, opened 2025-07-15
fix(ci): Update from windows 2019 to windows 2022

hubertdeng123:
Seems like symbolicator release is failing due to deprecated windows version: https://github.com/getsentry/symbolicator/actions/runs/16299649338/job/46030635724

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,7 +112,7 @@ jobs:
 
   windows:
     name: Build Tools on Windows
-    runs-on: windows-2025
+    runs-on: windows-2022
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -112,7 +112,7 @@ jobs:
 
   windows:
     name: Build Tools on Windows
-    runs-on: windows-2019
+    runs-on: windows-2025
```
Dav1dde:
```suggestion
    runs-on: windows-2022
```

We should aim to support the oldest version

PR #1737, opened 2025-07-18
ref(symbolicator): Raise fetch limit/page size when querying dysms from Sentry

Dav1dde:
Should fix SYMBOLICATOR-Y2, this is now happening because the API searches for debug and code id.


tobias-wilfert (approved):

Dav1dde:
Most likely the wrong fix, trying to fix it in the endpoint via: https://github.com/getsentry/sentry/pull/95861

PR #1740, opened 2025-07-21
ci: run self-hosted e2e test

aldy505:
I am trying to add this e2e test across repositories that self-hosted depends on. Hopefully it'll make tracking broken commits easier in the future.


loewenheim (changes requested):
Hi, thank you for the contribution! It seems to me that the test should run as part of the `ci` workflow, not `image`.

aldy505:
@loewenheim it should be on image, since it waits for the current SHA-commit image to be built.

loewenheim (approved):
> @loewenheim it should be on image, since it waits for the current SHA-commit image to be built.

Hmm, right. That is perhaps something to be figured out later.

PR #1770, opened 2025-08-21
feat: devenv, uv, pre-commit

joshuarli:
This introduces devenv, uv and pre-commit in accordance with these repo standards:
https://www.notion.so/sentry/Standard-Spec-devenv-2238b10e4b5d8072a992d8b67d94be22
https://www.notion.so/sentry/Standard-Spec-python-uv-2248b10e4b5d8045b8fff30f8b8b67ca
https://www.notion.so/sentry/Standard-Spec-pre-commit-2238b10e4b5d8019aa1fe1158e39c38e

The ability to run `devenv sync` and get a dev environment in any repo is the DX i'm aiming for,
and these standards provide a foundation from which later we can easily push updates to all repos.

You should be able to checkout this branch, run `direnv allow` and `devenv sync` (might be prompted to run `~/.local/share/sentry-devenv/bin/devenv update` first), which will set up everything needed for development (other than cargo).

Notes:
- I was also able to add `rustfmt` as a local pre-commit hook (need to keep the rust edition in sync with cargo.toml it seems) but it seems like clippy doesn't work on individual files (perhaps clippy-driver?) - anyway, decided not to touch these parts for now to keep things simple.


pyproject.toml
```diff
@@ -0,0 +1,22 @@
...
+    "mkdocs>=1.6.1",
+    "mkdocs-material>=9.6.17",
+    "pre-commit>=4.3.0",
+    "pygments>=2.19.2",
+]
+
+[tool.uv]
+environments = ["sys_platform == 'darwin' or sys_platform == 'linux'"]
+
+# TODO pypi deps
```
joshuarli: waiting for https://github.com/getsentry/pypi/pull/1610

joshuarli: and https://github.com/getsentry/symbolicator/pull/1771

joshuarli:
> I honestly question the utility of `clippy` being a pre-commit hook. I don't think it's helpful to require all individual commits to be correctly linted, and you can't merge without it anyway.

Yeah, I just realized clippy/clippy-driver is a wrapper around rustc - rust linting inherently requires compilation, making it too slow for pre-commit anyways.

PR #1780, opened 2025-09-09
Add support for fetching minidumps directly from objectstore

Swatinem:
Instead of receiving attachments via a multipart upload, this will instead fetch the minidump via the `objectstore-client`. For that, it currently still transfers the necessary metadata via yet another multipart field.

Eventually, the minidump endpoint should transition over to a JSON body instead of multipart, in the distant future once this is fully rolled out.

---

This is currently blocked because of some dependency issue. I ran a `cargo update` to resolve this, but then ran into a problem that our `patch` for `reqwest` is too old and is thus not being used.

Judging by this, it would appear that noone has run a `cargo update`, or cared about maintaining dependencies ever since I switched teams?


crates/symbolicator/src/service.rs
```diff
@@ -220,8 +222,13 @@ impl RequestService {
             });
         }
 
-        let max_concurrent_requests = config.max_concurrent_requests;
+        let attachments_store = objectstore_client::ClientBuilder::new(
+            config.storage.service_url,
+            config.storage.jwt_secret,
```
loewenheim: The `storage` config doesn't seem to exist yet.

Cargo.toml
```diff
@@ -96,6 +96,7 @@ minidump = "0.26.0"
 minidump-processor = "0.26.0"
 minidump-unwind = "0.26.0"
 moka = { version = "0.12.8", features = ["future", "sync"] }
+objectstore-client = { git = "https://github.com/getsentry/objectstore.git" }
```
Dav1dde, line 99: No git dependencies without a revision please, I know it's commited in the lockfile, but git doesn't have semantic versioning.

Swatinem:
I was thinking about a "pre-signed GET" myself, as it would be simpler than having to pass down different parameters.
While it might be possible to create such an URL, we do not strictly follow http semantics in relation to compression, so there might be some problems there.

Though we would have to fully test this first.

Swatinem:
We probably can’t use that for debug files, as we have to *query* them based on the debug-id. But in general, using a url directly is probably a good idea.

Dav1dde:
> We probably can’t use that for debug files, as we have to query them based on the debug-id. But in general, using a url directly is probably a good idea.

I think that should still work, we query them from Sentry and Sentry gives us pre-signed URLs back we can use to download them in Symbolicator or am I missing something?

PR #1796, opened 2025-09-29
Remove storing failed minidumps in the `diagnostics_cache`

Swatinem:
Minidumps that failed processing used to be stored in the `diagnostics_cache` so that we could download and inspect them for debugging purposes. That however is a tedious process, and most likely hasn’t been used by anyone in years.

We can clean up this functionality, as doing so simplifies the code and unblocks further improvements to fetch minidump attachments directly from the future `objectstore`.


jjbayer (approved):
Would be good to add a motivation to the PR description so in a few years we still know why we removed this.

PR #1799, opened 2025-10-01
feat(symbolicator): Healthcheck command for self-hosted

aldy505:
Similar work to https://github.com/getsentry/relay/pull/5044 due to distroless migration

This works on relay and this is making CI's red right now.

Changes made:
- Adds a new `healthcheck` sub-command
- Defaults to the values from the config
- Changes logger to use `stderr` to still allow for logging and parsing of the healthcheck output (Relay uses the same mechanism here)

### DIFF
--- a/crates/symbolicator/src/cli.rs
+++ b/crates/symbolicator/src/cli.rs
@@ -55,8 +55,8 @@ enum Command {
     Healthcheck {
         /// Timeout for the healthcheck request.
         /// Defaults to 30 seconds.
-        #[arg(long)]
-        timeout: Option<Duration>,
+        #[arg(long, default_value_t = 30)]
+        timeout: u64,
 
         /// Host to check. Defaults to `localhost`.
         #[arg(long, default_value_t = String::from("localhost"))]
@@ -214,7 +214,7 @@ pub fn execute() -> Result<()> {
             port,
         } => {
             let client = reqwest::blocking::Client::builder()
-                .timeout(timeout.into())
+                .timeout(std::time::Duration::from_secs(timeout))
                 .build()
                 .unwrap_or_default();
             let url = format!("http://{host}:{port}/healthcheck");
--- a/crates/symbolicator/src/cli.rs
+++ b/crates/symbolicator/src/cli.rs
@@ -1,5 +1,5 @@
 //! Exposes the command line application.
-use std::net::ToSocketAddrs;
+use std::net::{SocketAddr, ToSocketAddrs};
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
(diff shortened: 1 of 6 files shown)


crates/symbolicator/src/cli.rs
```diff
@@ -191,6 +202,37 @@ pub fn execute() -> Result<()> {
...
+        Command::Healthcheck {
+            timeout,
+            host,
+            port,
+        } => {
+            let client = reqwest::blocking::Client::builder()
+                .timeout(Duration::from_secs(timeout))
+                .build()
+                .unwrap_or_default();
+            let url = format!("http://{host}:{port}/healthcheck", host, port);
```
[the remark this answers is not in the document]
aldy505:
```suggestion
            let url = format!("http://{host}:{port}/healthcheck");
```

Dav1dde (approved):

crates/symbolicator/src/logging.rs
```diff
@@ -62,7 +62,8 @@ pub unsafe fn init_logging(config: &Config) {
     let fmt_layer = {
         let layer = tracing_subscriber::fmt::layer()
             .with_timer(UtcTime::rfc_3339())
-            .with_target(true);
+            .with_target(true)
+            .with_writer(std::io::stderr);
```
jjbayer, line 66: Will this cause problems anywhere? Are we collecting logs somewhere where we only consider stdout?

crates/symbolicator/src/cli.rs
```diff
@@ -191,6 +209,9 @@ pub fn execute() -> Result<()> {
             repeat.map(|inner| inner.map(|time| time.into())),
         )
         .context("failed to clean up caches")?,
+        Command::Healthcheck { addr, timeout } => {
+            healthcheck::healthcheck(config, addr, timeout).context("healthcheck failed")?
+        }
```
jjbayer, line 214: Is there no other way to get a healthcheck command into a distroless image? E.g. ship the image with a second binary that we can use in multiple projects?

crates/symbolicator/src/logging.rs
Dav1dde: Going to add a changelog entry for this! I think it's just a better default though and matches what we use in Relay.

crates/symbolicator/src/cli.rs
Dav1dde: It would be nice to have, a very simple curl, like binary.

But without setting up a proper pipeline to (re-)build base images, this is not great. The upside of doing it within `symbolicator` and `relay` is, the command can read the config and extract host/port from the config, meaning you can use the same command line args.

Since we already depend on reqwest anyways it's imo an acceptable trade-off.

aldy505: We could ship a simple `curl`-like program. I think it's better this way, it'll be lighter.

PR #1802, opened 2025-10-06
feat(build): use action-build-push-images

joshuarli:


.github/workflows/image.yml
```diff
@@ -107,118 +112,38 @@ jobs:
...
+      - name: Build and push production image
+        if: github.ref_name == github.event.repository.default_branch
+        uses: getsentry/action-build-and-push-images@a53f146fc1ea3cb404f2dcf7378f5b60dd98d3ca
         with:
-          name: symbolicator-image@${{ matrix.arch }}
-          path: /tmp/symbolicator-${{ matrix.arch }}.tar
-
-  assemble-ghcr:
-    needs: [build-setup, build-image]
-    if: "needs.build-setup.outputs.full_ci == 'true'"
```
joshuarli, line 136: this is now done on every commit since e2e now needs ghcr.io/getsentry/symbolicator:${{ github.sha }} because it can't download symbolicator-amd64.tar anymore

not a big deal imo, very little overhead since actually building the final docker image is really fast

.github/workflows/image.yml
```diff
@@ -107,118 +112,38 @@ jobs:
...
           platforms: linux/${{ matrix.arch }}
-          tags: symbolicator-${{ matrix.arch }}
-          outputs: type=docker,dest=/tmp/symbolicator-${{ matrix.arch }}.tar
-          push: false
-
-      - name: Upload Image
-        uses: actions/upload-artifact@v4
+          dockerfile_path: './Dockerfile'
+          build_context: './docker-ctx'
+          ghcr: true
```
joshuarli, line 129: hmm it's not pushing anything currently even though `ghcr: true`

Publishing to GHCR: false
Publishing to GAR: false

https://github.com/getsentry/symbolicator/actions/runs/18296387570/job/52095521270?pr=1802

.github/workflows/image.yml
```diff
@@ -260,18 +200,9 @@ jobs:
...
-
-      - name: Load Docker Image
-        run: docker load --input /tmp/symbolicator-amd64.tar
-
       - name: Run Sentry self-hosted e2e CI
         uses: getsentry/self-hosted@master
         with:
           project_name: symbolicator
-          image_url: symbolicator-amd64
+          image_url: ghcr.io/getsentry/symbolicator:${{ github.sha }}
```
joshuarli: hmm not sure why this is resolvig to `ghcr.io/getsentry/symbolicator:fbf420b5b599548d6598d552965a99f42760783b`, doesnt even exist

.github/workflows/image.yml
```diff
@@ -107,122 +112,57 @@ jobs:
...
+          tag_nightly: false
+          tag_latest: false
+          google_ar_image_name: us-central1-docker.pkg.dev/sentryio/symbolicator/image
+          google_workload_identity_provider: projects/868781662168/locations/global/workloadIdentityPools/prod-github/providers/github-oidc-pool
+          google_service_account: gha-gcr-push@sac-prod-sa.iam.gserviceaccount.com
 
   gocd-artifacts:
     needs: [build-setup, build-image]
-    if: "needs.build-setup.outputs.full_ci == 'true'"
+    if: github.ref_name == github.event.repository.default_branch
```
Dav1dde: That is not correct. The Full CI trigger should also upload gocd artifacts. It's a way to test workflows in PRs but also a way to do a full 'production ready' build.

Dav1dde (commented):
I don't think this workflow works the way you set it up. It never builds the multi arch image, which a majority of the workflow did you ended up replacing. Have you checked that the output/uploaded image is actually multi arch?

Also this workflow now no longer works on forks because it needs push permissions where previously it ran fine for external contributors.

.github/workflows/image.yml
```diff
@@ -107,122 +112,57 @@ jobs:
...
-
       - name: Prepare Docker Context
         run: |
           mkdir docker-ctx
           cp Dockerfile docker-ctx/
           mv symbolicator docker-ctx/
 
-      - name: Build Image
-        uses: docker/build-push-action@v6
+      - name: Build and push ghcr image
```
hubertdeng123, line 121: What we need to do here to get forks working is https://github.com/getsentry/relay/blob/62787c1b1db6063e49b51f7f696e58ef2fedb9a7/.github/workflows/ci.yml#L472-L495

That way things work for forks since workflows using the docker image are still ok even if no permissions are given to push to ghcr.

hubertdeng123: Since the `cargo build` step for the specific os is done in GHA, we can separate out the workflow for building the image into another job that is separate from the build process. This job passes in `platforms: linux/amd64,linux/arm64` and we can directly build the multiplatform manifest as done in relay without assembling the image.

.github/workflows/image.yml
joshuarli: ah i see, wasn't sure what full ci trigger was for - thanks!

.github/workflows/image.yml
```diff
@@ -107,118 +112,41 @@ jobs:
...
+        uses: getsentry/action-build-and-push-images@b172ab61a5f7eabd58bd42ce231b517e79947c01
         with:
-          context: docker-ctx/
+          image_name: 'symbolicator'
           platforms: linux/${{ matrix.arch }}
```
hubertdeng123: We can directly build multi-arch image here without running across arm and amd runners, since the meat of the work of `cargo build` is already done from a previous step and shouldn't result in significant increases in time
```suggestion
          platforms: linux/amd64,linux/arm64
```

.github/workflows/image.yml
```diff
@@ -119,6 +119,7 @@ jobs:
           mv symbolicator docker-ctx/
 
       - name: Build and push ghcr image
+        if: "!github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]'"
```
joshuarli, line 122: @hubertdeng123 thoughts on just disabling publishing if `github.event.pull_request.head.repo.fork`, even if publish_on_pr=true, in the action itself? seems like a good default to me

hubertdeng123: that makes sense to me, it should error out due to lack of permissions but we should make it an explicit failure mode

.github/workflows/image.yml
```diff
@@ -107,118 +112,42 @@ jobs:
...
+        if: "needs.build-setup.outputs.full_ci == 'true'"
+        uses: getsentry/action-build-and-push-images@b172ab61a5f7eabd58bd42ce231b517e79947c01
         with:
-          pattern: symbolicator-image@*
-          path: /tmp
-          merge-multiple: true
-
-      - *assemble
+          image_name: 'symbolicator'
+          platforms: linux/amd64,linux/arm64
```
hubertdeng123, line 140: Do we need multiplatform image for production? Perhaps @Dav1dde knows better here, but I guess there is no harm having it

.github/workflows/image.yml
```diff
@@ -107,118 +112,42 @@ jobs:
...
+          dockerfile_path: './Dockerfile'
+          build_context: './docker-ctx'
+          ghcr: false
+          google_ar: true
+          # note: nightly will only be tagged if on default branch
+          tag_nightly: true
```
hubertdeng123, line 146: We only tag nightly for self-hosted releases on ghcr, prod should be tagged with latest (weird I know) 

```suggestion
          tag_nightly: false
          tag_latest: true
```

.github/workflows/image.yml
```diff
@@ -260,18 +189,9 @@ jobs:
...
-
-      - name: Load Docker Image
-        run: docker load --input /tmp/symbolicator-amd64.tar
-
       - name: Run Sentry self-hosted e2e CI
         uses: getsentry/self-hosted@master
         with:
           project_name: symbolicator
-          image_url: symbolicator-amd64
+          image_url: ghcr.io/getsentry/symbolicator:${{ github.event.pull_request.head.sha || github.sha }}
```
hubertdeng123, line 196: I think this will still fail on forks, we may need to ensure we're building the image on forks, saving the artifact, and loading it here to ensure it works

hubertdeng123: Something similar to https://github.com/getsentry/relay/blob/master/.github/workflows/ci.yml#L521-L531

.github/workflows/image.yml
hubertdeng123: er yeah it shouldn't fail, but give out a warning

.github/workflows/image.yml
joshuarli: i thought we want to skip publishing (so that CI can still pass overall) if it's a run from a fork?
(then again, we lose self hosted e2e testing)

it's okay to have the permissions and push images to GHCR then? i feel like it's fine (as long as `nightly` isn't published)?

.github/workflows/image.yml
joshuarli: hmm actually never mind, i was confused

.github/workflows/image.yml
Dav1dde: We need multi arch for production **and** self hosted.

Dav1dde, line 140: Have you actually tested the images you're building? You're creating a amd64 image with a arm64 binary and vice versa.

I don't think this setup can even work.

hubertdeng123: Yeah, we need to make sure multi-arch is available for self-hosted on ghcr, but are we running symbolicator on arm64 and amd64 in prod? Since this is pushing to artifact registry

hubertdeng123: @joshuarli This is accurate, we need to ensure that the `cargo build` step is separate from building the docker image as right now we're publishing the multi arch image with the wrong binary twice. We need to ensure that the docker image build and publish happens once and copies the correct binary when building. This may require necessary changes to the dockerfile to make sure we're copying the right binary. 

Example in relay: https://github.com/getsentry/relay/blob/468fc6fca10a0d29bb78f39149b1b3fa358758f9/Dockerfile.release#L8

.github/workflows/image.yml
hubertdeng123: We should definitely skip publishing for forks, I think I linked the wrong thing above over there, but it's still possible to build the image and save the image as an artifact to be used for the self-hosted e2e tests here even on a fork. Essentially that would require keeping the existing logic for 

```
- name: Download Docker Image
   uses: actions/download-artifact@v5
   with:
     pattern: symbolicator-image@amd64
     path: /tmp
  
- name: Load Docker Image
   run: docker load --input /tmp/symbolicator-amd64.tar

```

and adding a job to save the docker image as an artifact to be reused later. Here's the right link: https://github.com/getsentry/relay/blob/master/.github/workflows/ci.yml#L485-L503

.github/workflows/image.yml
Dav1dde: > but are we running symbolicator on arm64 and amd64 in prod? Since this is pushing to artifact registry

We are.

Dav1dde: @hubertdeng123 do you think it would be feasible to adjust the build action to support assembly from other images, similar to what is done here by loading multiple different architectures? It's just unfortunate that docker still doesn't provide good tools to deal with multi arch images.

The setup here was inspired from [`objectstore`](github.com/getsentry/objectstore), so we'll have to do the same changes to that repo (eventually).

.github/workflows/image.yml
joshuarli: Yeah, looks like i'll have to keep that stuff around

.github/workflows/image.yml
joshuarli: yeah right now both images have the amd64 binary, i need to rework this pr 
```
docker run --platform=linux/arm64 ghcr.io/getsentry/symbolicator:55ae450aac76c48d1c12340d96282d8c0eabc09c                      
rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2
```

hubertdeng123: @Dav1dde Do you mean add logic to the composite action to take existing images and assemble them into the multiplatform manifest?

Currently there exists two processes for building multiplatform images. It's not possible for us to add assembly of images to the 2nd approach since we can't add a matrix strategy to a composite action, it can only be done from the job that calls the composite action. 

1. Artifact build on platform specific runners, image build on one runner with emulator (relay is doing this)
- `action-build-and-push-images` is called once and directly published multi platform image without performance degradation since the most time consuming part of building artifacts is done on the appropriate runners
- requires dockerfile to handle copying the binary with the correct platform

2. Image build happens on platform specific runners, assembly is done manually at the end
- `action-build-and-push-images` is called twice and builds and publishes images with suffix of either `arm64` or `amd64`
- assembly for multiplatform image can only happen after the two jobs are complete

Dav1dde: > Currently there exists two processes for building multiplatform images. It's not possible for us to add assembly of images to the 2nd approach since we can't add a matrix strategy to a composite action, it can only be done from the job that calls the composite action.

I am more thinking, let the matrix build the images (like Symbolicator/Objectstore do). And then the `action-build-and-push-images` can run outside of the matrix and be just passed paths to the raw images (as they are passed around via CI artifacts).

hubertdeng123: Created an item to address that, makes sense to have

PR #1805, opened 2025-10-14
feat(build): replace assembly with action-build-push-images

joshuarli:
redo of https://github.com/getsentry/symbolicator/pull/1802; we need to preserve the docker tar building step for the self-hosted e2e tests

discussion here: https://github.com/getsentry/symbolicator/pull/1802/files#r2428050962

tl;dr it would be best to have the option to pass a docker tar so we don't have to rebuild with action-build-push-images, but we're lucky here that the build is so lightweight that this is acceptable in my mind for now. Tracked here: https://linear.app/getsentry/issue/DI-1382/action-build-and-push-images-should-have-the-option-to-take-a-docker

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -240,5 +240,7 @@ jobs:
           project_name: symbolicator
           image_url: symbolicator-self-hosted
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-          env:
-            DOCKER_BUILDKIT: 0
+        env:
+          # XXX: this action doesn't work if buildx is available (it's installed by action-build-and-push-images)
+          #      because docker compose will build with type=docker and not type=image
+          DOCKER_BUILDKIT: 0


.github/workflows/image.yml
```diff
@@ -131,94 +137,52 @@ jobs:
...
-          IMAGES=()
-          for image in /tmp/symbolicator-*.tar; do
-            NAME="$(basename $image .tar)"
-            ARCH="${NAME#*-}"
-            TARGET="${IMAGE}:${{ github.sha }}-${ARCH}"
-
-            docker load --input "${image}"
-            docker tag "${NAME}" "${TARGET}"
-            docker push "${TARGET}"
+          cp Dockerfile /tmp/docker-ctx
```
hubertdeng123, line 139: What's the reason behind copying the dockerfile to this tmp directory? Can we just build from the project root directory?

hubertdeng123: Ah, I see we're copying the binary built for the specific platform to that directory and we want it available when building the image

.github/workflows/image.yml
```diff
@@ -131,94 +137,52 @@ jobs:
...
-
-      - name: Download Images
-        uses: actions/download-artifact@v5
-        with:
-          pattern: symbolicator-image@*
-          path: /tmp
-          merge-multiple: true
-
-      - *assemble
+          image_name: 'symbolicator'
```
hubertdeng123, line 152: Nice, we publish to ghcr and gar in one job

.github/workflows/image.yml
```diff
@@ -112,9 +118,9 @@ jobs:
 
       - name: Prepare Docker Context
         run: |
-          mkdir docker-ctx
+          mkdir -p docker-ctx/binaries/linux/${{ matrix.arch }}
           cp Dockerfile docker-ctx/
-          mv symbolicator docker-ctx/
+          mv symbolicator docker-ctx/binaries/linux/${{ matrix.arch }}
 
       - name: Build Image
```
hubertdeng123: We should also use `getsentry/action-build-and-push-images` here to ensure we have the right metadata attached, it should be possible to output to a .tar file

.github/workflows/image.yml
hubertdeng123: Is it possible to link to example publish jobs that have succeeded?

.github/workflows/image.yml
joshuarli: Not really much of a reason other than it emphasizes exactly what context's needed to build the image. In true distroless fashion.

.github/workflows/image.yml
```diff
@@ -88,6 +88,12 @@ jobs:
           cargo build --release --locked --target="${RUST_TARGET}" --features="${FEATURES}"
           mv "target/${RUST_TARGET}/release/symbolicator" ./symbolicator
 
+      - name: Upload Binary
+        uses: actions/upload-artifact@v4
+        with:
+          name: symbolicator-binary@${{ matrix.arch }}
+          path: ./binaries/linux/${{ matrix.arch }}/symbolicator
```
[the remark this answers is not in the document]
joshuarli: oops yeah i need to flesh this out still

.github/workflows/image.yml
joshuarli: yep, here's one for 2dc729a034610f4e21cc18b888cf8a8c39b76377 https://github.com/getsentry/symbolicator/pkgs/container/symbolicator/544600995?tag=2dc729a034610f4e21cc18b888cf8a8c39b76377

.github/workflows/image.yml
```diff
@@ -131,94 +138,56 @@ jobs:
...
-        with:
-          pattern: symbolicator-image@*
-          path: /tmp
-          merge-multiple: true
-
-      - *assemble
+          image_name: 'symbolicator'
+          platforms: linux/amd64,linux/arm64
+          build_context: '/tmp/docker-ctx'
+          publish_on_pr: true  # TEMPORARY FOR TESTING
```
joshuarli: self reminder to remove this before merging

.github/workflows/image.yml
joshuarli: did it - worked!

.github/workflows/image.yml
```diff
@@ -88,6 +88,12 @@ jobs:
           cargo build --release --locked --target="${RUST_TARGET}" --features="${FEATURES}"
           mv "target/${RUST_TARGET}/release/symbolicator" ./symbolicator
 
+      - name: Upload Binary
+        uses: actions/upload-artifact@v4
```
hubertdeng123: nit: use sha here instead of v4

.github/workflows/image.yml
```diff
@@ -112,113 +118,76 @@ jobs:
...
-          push: false
 
       - name: Upload Image
         uses: actions/upload-artifact@v4
         with:
           name: symbolicator-image@${{ matrix.arch }}
           path: /tmp/symbolicator-${{ matrix.arch }}.tar
 
-  assemble-ghcr:
+  assemble:
```
hubertdeng123, line 116: I wonder if there's an argument towards keeping ghcr publishing and gar publishing separate jobs given a scenario where ghcr goes down, the job fails entirely and deploys are blocked. I like the idea of combining all this into one workflow though. WDYT @Dav1dde?

.github/workflows/image.yml
```diff
@@ -112,113 +118,76 @@ jobs:
...
+      packages: write  # required for GHCR
+      contents: read
+      id-token: write  # required for GAR
 
     steps:
-      - name: Docker Login
-        run: docker login --username '${{ github.actor }}' --password-stdin ghcr.io <<< "$GHCR_TOKEN"
-        env:
-          GHCR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/checkout@v5
```
hubertdeng123: nit: pin by sha

joshuarli: will do right before merging

joshuarli:
this is ready for review at this point. multiarch was published to GAR here (will revert publish on prs before merging this PR): 
- https://console.cloud.google.com/artifacts/docker/sentryio/us-central1/symbolicator/image/sha256:f9e0db754c56db9e46295ade00f701aa36c9719665b304ca86607089dbc6361a;tab=manifest?project=sentryio 
- and multiarch on ghcr as well: https://github.com/getsentry/symbolicator/pkgs/container/symbolicator/547166575?tag=990e9c1dae89a9b8f66686d2ad558acd0e25fd51

i've pulled all 4 images and they work!

.github/workflows/image.yml
```diff
@@ -102,123 +108,88 @@ jobs:
...
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3
 
       - name: Prepare Docker Context
         run: |
-          mkdir docker-ctx
+          mkdir -p docker-ctx/binaries/linux/${{ matrix.arch }}
           cp Dockerfile docker-ctx/
-          mv symbolicator docker-ctx/
+          mv symbolicator docker-ctx/binaries/linux/${{ matrix.arch }}
```
Dav1dde: Why is this still happening if you upload the binary?

.github/workflows/image.yml
Dav1dde: I think it was just easier to deal with conditionals (publish to this or that) and docker credentials in separate jobs. Good idea to consolidate these.

.github/workflows/image.yml
Dav1dde: We can get rid of the context, this was me not realizing there is a `.dockerignore` that was making trouble and then not cleaning it up.

Dockerfile
```diff
@@ -6,7 +6,12 @@ COPY --from=gcr.io/distroless/cc-debian12:nonroot --chown=nonroot:nonroot /home/
...
 
-ARG BINARY=./symbolicator
-COPY ${BINARY} /bin/symbolicator
+ARG TARGETPLATFORM
+
+ARG BINARY=./binaries/$TARGETPLATFORM/symbolicator
+COPY --chmod=0755 ${BINARY} /bin/symbolicator
+
+# sanity check
+RUN ["/bin/symbolicator", "help"]
```
Dav1dde: Should definitely not merge this.

.github/workflows/image.yml
joshuarli: Not sure what you're asking here? Are you suggesting I should shape the `path: ./symbolicator` in upload to the new paths I need for the docker context so to not mv twice? I just didn't want to touch the paths in `split debug info` so left it as `./symbolicator`.

Dockerfile
joshuarli: Will revert before merging!

.github/workflows/image.yml
Dav1dde: You are in the matrix build here, to build the binary, then later assemble multiple binaries into a single multi arch image.

But right here you also seem to be still building a docker image. I am not sure why and I think this should not be happening nor necessary.

joshuarli: It's not building a finished image, just the platform-specific `.tar` for `self-hosted-end-to-end`. The assemble step depends on `build-image` (only for the binaries).

Dav1dde: Ah I see, I missed that you switched the target 👍

Wonder if we instead should build the image only demand in the self hosted step with the binary. Might clean this up a bit.

.github/workflows/image.yml
```diff
@@ -102,123 +108,88 @@ jobs:
...
-        run: gcloud auth configure-docker us-central1-docker.pkg.dev
-
-      - name: Download Images
-        uses: actions/download-artifact@v5
+      - name: Prepare Docker Context
+        run: |
+          cp Dockerfile /tmp/docker-ctx
+          mkdir -p /tmp/docker-ctx/binaries/linux/amd64 /tmp/docker-ctx/binaries/linux/arm64
+          mv /tmp/docker-ctx/symbolicator-binary-amd64/symbolicator /tmp/docker-ctx/binaries/linux/amd64
+          mv /tmp/docker-ctx/symbolicator-binary-arm64/symbolicator /tmp/docker-ctx/binaries/linux/arm64
```
Dav1dde, line 142: If we get rid of the `docker-ctx` namespace, this should be redundant?

.github/workflows/image.yml
joshuarli: Hmmm yeah that's a good idea, we can avoid uploading then downloading the tar then!

.github/workflows/image.yml
joshuarli: I kind of like having a dir for docker context - makes it easy to make sense of what's needed to build

Dav1dde:
Mh so it doesn't pull it from the local registry 🤔

PR #1806, opened 2025-10-17
fix(proguard): Keep filenames for inline frames

romtsn:
Since [this PR](https://github.com/getsentry/rust-proguard/pull/36) was introduced to rust-proguard, we can now successfully remap filenames for any kind of frames (also inline/foreign ones), so this logic is obsolete.


crates/symbolicator-proguard/src/symbolication.rs
```diff
@@ -1035,4 +1027,110 @@ io.wzieba.r8fullmoderenamessources.R -> a.d:
...
+                    .into_iter()
+            })
+            .map(|frame| (frame.filename.unwrap(), frame.abs_path.unwrap()))
+            .unzip();
+
+        assert_eq!(
+            remapped_filenames,
+            [
+                "Renderer.kt",
+                "Delegate.kt",
```
romtsn, line 1112: WIthout the change, this line and the one below would be empty

PR #1825, opened 2025-11-18
ci: prepare SH e2e test on a different job

aldy505:


aldy505:
Hubert took it over :D

PR #1837, opened 2025-11-27
feat(download): Remove streaming timeout

### LINKED ISSUE #1836: Remove content length based timeout
Symbolicator currently [has a download timeout calculated based on the file's content length.](https://github.com/getsentry/symbolicator/blob/master/crates/symbolicator-service/src/download/mod.rs#L914)

It has been discussed to remove this limit and use the maximum timeout of 5 minutes for all files.

loewenheim:
This (morally) reverts #489, reducing our download timeout logic to a simple constant timeout. It also increases the default value of `max_download_timeout` from 5min to 10min.

Fixes #1836. Fixes SYMBOLI-41.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Breaking Changes
+
+- Removed the `streaming_timeout` config option. There is now only a constant timeout, controlled
+  by `max_download_timeout`. The default of `max_download_timeout` is increased to 10 minutes. (#1837) by @loewenheim
+
 ### Various fixes & improvements
 
 - Added config options `crash_file_body_max_bytes` and `symbolicate_body_max_bytes` controlling  the maximum body size for requests to crash file and general symbolication endpoints,
@@ -11,8 +16,6 @@
   (#1827) by @loewenheim
 - CAB files are now extracted in-process, removing the dependency on the external
   `cabextract` binary (#1834) by @loewenheim
-- Removed the `streaming_timeout` config option. There is now only a constant timeout, controlled
-  by `max_download_timeout`. The default of `max_download_timeout` is increased to 10 minutes. (#1837) by @loewenheim
 
 ## 25.11.0
 


crates/symbolicator-service/src/download/mod.rs
```diff
@@ -752,27 +749,14 @@ impl SymResponse<'_> {
...
-            .map(|cl| content_length_timeout(cl, self.timeouts.streaming))
-            .unwrap_or(self.timeouts.max_download);
-
         let mut stream = self.response.bytes_stream().map_err(CacheError::from);
         // Transfer the contents, into the destination and track progress.
-        tokio::time::timeout(timeout, async move {
-            while let Some(chunk) = stream.next().await.transpose()? {
-                self.measure.add_bytes_transferred(chunk.len() as u64);
-                destination.write_buf(chunk).await?;
-            }
```
loewenheim, line 770: This completely removes both the streaming and max timeout from the download. I believe this is justified because
1. our `reqwest` clients contain the max timeout
2. our download futures are wrapped in the max timeout in `download`.

crates/symbolicator-service/src/download/s3.rs
```diff
@@ -182,24 +182,13 @@ impl S3Downloader {
...
 
         let mut destination = std::pin::pin!(destination.into_write());
 
-        tokio::time::timeout(
-            timeout,
-            super::download_stream(source_name, stream, destination.as_mut()),
-        )
-        .await
-        .map_err(|_| CacheError::Timeout(timeout))??;
+        super::download_stream(source_name, stream, destination.as_mut()).await?;
```
loewenheim, line 191: Similar to above, this completely removes both the streaming and max timeout from the download. I believe this is justified because the download future is wrapped in the mix timeout in `download`.

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
   (#1827) by @loewenheim
 - CAB files are now extracted in-process, removing the dependency on the external
   `cabextract` binary (#1834) by @loewenheim
+- Removed the `streaming_timeout` config option. There is now only a constant timeout, controlled
+  by `max_download_timeout`. The default of `max_download_timeout` is increased to 10 minutes. (#1837) by @loewenheim
```
jjbayer: Should we move this to `**Breaking Changes**`?

PR #1845, opened 2025-12-12
feat(proguard): Support rewriteFrame annotation

romtsn:
Attempts to reuse `remap_frame` from ProguardCache to delegate as much logic to it as possible and keep it lean on the caller side.

Additionally, had to extend the `JvmStacktrace` interface, because we now need the `JvmException` associated with the stacktrace to correctly apply rewrite rules. 

Depends on https://github.com/getsentry/rust-proguard/pull/67

### DIFF
--- a/crates/symbolicator-proguard/src/symbolication.rs
+++ b/crates/symbolicator-proguard/src/symbolication.rs
@@ -16,8 +16,10 @@ use symbolicator_service::types::FrameOrder;
 
 /// Result of attempting a full frame remap with rewrite rules.
 enum FullRemapResult {
-    /// Outline frame or rewrite rules cleared all frames - skip this frame entirely.
-    Skip,
+    /// Outline frame - skip this frame but preserve rewrite eligibility for next frame.
+    OutlineFrame,
+    /// Rewrite rules cleared all frames - skip this frame and disable rewrite for subsequent frames.
+    RewriteCleared,
     /// Successfully remapped to these frames.
     Frames(Vec<JvmFrame>),
     /// No mappings found - try fallback methods.
@@ -290,8 +292,12 @@ impl ProguardService {
                     &mut carried_outline_pos[mapper_idx],
                     &mut remap_buffer,
                 ) {
-                    FullRemapResult::Skip => {
-                        // Outline frame or rewrite rules cleared all frames
+                    FullRemapResult::OutlineFrame => {
+                        // Outline frame - skip but preserve rewrite eligibility for next frame
+                        continue 'frames;
+                    }
+                    FullRemapResult::RewriteCleared => {
+                        // Rewrite rules cleared all frames - skip and disable rewrite
                         next_frame_can_rewrite = false;
                         continue 'frames;
                     }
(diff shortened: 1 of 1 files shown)


crates/symbolicator-proguard/src/interface.rs
```diff
@@ -128,6 +128,8 @@ pub struct JvmException {
 /// A JVM stacktrace.
 #[derive(Debug, Default, Clone, Deserialize, Serialize, PartialEq, Eq)]
 pub struct JvmStacktrace {
+    /// The exception that caused the stacktrace.
+    pub exception: Option<JvmException>,
```
romtsn, line 132: unsure if this is a breaking change in the symbolicator-service API (but probably not because it's an Option)?

crates/symbolicator-proguard/src/interface.rs
```diff
@@ -128,6 +128,8 @@ pub struct JvmException {
...
     /// The stacktrace's frames.
     pub frames: Vec<JvmFrame>,
 }
```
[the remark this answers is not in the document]
romtsn: nah, it will be populated from the caller side (sentry)

crates/symbolicator-proguard/src/interface.rs
loewenheim: Yeah, that should be fine. Should still be pointed out in the API docs, though.

crates/symbolicator-proguard/src/symbolication.rs
```diff
@@ -14,6 +14,16 @@ use symbolicator_service::caching::CacheError;
...
 
+/// Result of attempting a full frame remap with rewrite rules.
+enum FullRemapResult {
+    /// Outline frame or rewrite rules cleared all frames - skip this frame entirely.
+    Skip,
+    /// Successfully remapped to these frames.
+    Frames(Vec<JvmFrame>),
+    /// No mappings found - try fallback methods.
+    NoFrames,
+}
```
loewenheim, line 27: Very nice use of an enum :+1:

crates/symbolicator-service/src/caches/versions.rs
```diff
@@ -294,7 +294,7 @@ pub const PROGUARD_CACHE_VERSIONS: CacheVersions = CacheVersions {
         CacheVersion::new(4, CachePathFormat::V2),
     ],
 };
-static_assert!(proguard::PRGCACHE_VERSION == 3);
+static_assert!(proguard::PRGCACHE_VERSION == 4);
```
loewenheim, line 301: You also need a cache version update.

romtsn: done in https://github.com/getsentry/symbolicator/pull/1845/commits/2cab1cab9fadd826d09e31add0aab90646fbaed6 and https://github.com/getsentry/symbolicator/pull/1845/commits/f59f60dafd3b75c4b2e3b25a1f98bf7279d6e240

crates/symbolicator-proguard/src/interface.rs
romtsn: done in https://github.com/getsentry/symbolicator/pull/1845/commits/bde0a0f70d28107457769c1018760b3dc2b3ccbd

romtsn:
@loewenheim this should be good to review now, added docs and updated CacheVersion, also added an integration test

PR #1880, opened 2026-02-18
fix: Do not double write `user-agent`

tobias-wilfert:
With the logic introduced here https://github.com/getsentry/sentry/pull/88892 certain symbol sources might be associated with certain user-agents. For these sources symbolicator currently double writes the user-agent which is undesirable. This changes the logic such that we either write symolicator as the user-agent or the custom one but not both.

### DIFF
--- a/crates/symbolicator-service/src/download/http.rs
+++ b/crates/symbolicator-service/src/download/http.rs
@@ -9,7 +9,7 @@ use crate::{
     config::DownloadTimeouts,
 };
 
-use super::{Destination, USER_AGENT};
+use super::Destination;
 
 /// Downloader implementation that supports the HTTP source.
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


crates/symbolicator-service/src/download/http.rs
```diff
@@ -46,7 +46,19 @@ impl HttpDownloader {
...
+        let has_custom_user_agent = file_source
+            .source
+            .headers
+            .0
+            .keys()
+            .any(|k| k.eq_ignore_ascii_case("user-agent"));
+
+        if !has_custom_user_agent {
+            builder = builder.header(header::USER_AGENT, USER_AGENT);
+        }
```
Dav1dde: Can we use `default_headers` instead? https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.default_headers

tobias-wilfert: TIL, I tried to change this now (hope that setting a user-agent in a code path where we previously did not is ok)?

loewenheim: That should not be an issue I believe. Also TIL, `default_headers` is nice.

PR #1882, opened 2026-02-19
ref(s3): Use pre-signed URLs with a parallelized download for S3

Dav1dde:
Uses a pre-signed URL to pass to the parallelized HTTP download impl instead of relying on the AWS SDK single stream download.

This should speed up large file downloads by ~4x.

~~The refactor has an unfortunate downside that the error handling is no longer out-of-the-box provided by the SDK, meaning we fallback to the generic handling which does not parse the XML/error response of the S3 API.~~
S3 Error handling is now implemented.

Potential future improvement: Specify etag headers returned from the first download/stream to ensure a consistent view on the file.

### DIFF
--- a/crates/symbolicator-service/src/download/s3.rs
+++ b/crates/symbolicator-service/src/download/s3.rs
@@ -172,11 +172,14 @@ impl ErrorHandler for S3ErrorHandler {
             return error;
         }
 
-        let details = match metadata.as_ref().map(|m| (m.code(), m.message())) {
-            Some((Some(code), None)) => code.to_owned(),
-            Some((None, Some(message))) => format!("{status}: {message}"),
-            Some((Some(code), Some(message))) => format!("{code}: {message}"),
-            None | Some((None, None)) => status.to_string(),
+        let details = match (
+            metadata.as_ref().and_then(|m| m.code()),
+            metadata.as_ref().and_then(|m| m.message()),
+        ) {
+            (Some(code), None) => code.to_owned(),
+            (None, Some(message)) => format!("{status}: {message}"),
+            (Some(code), Some(message)) => format!("{code}: {message}"),
+            (None, None) => status.to_string(),
         };
 
         tracing::debug!(


crates/symbolicator-service/src/download/s3.rs
```diff
@@ -117,95 +126,24 @@ impl S3Downloader {
...
-        let mut destination = std::pin::pin!(destination.into_write());
-
-        super::download_stream(source_name, stream, destination.as_mut()).await?;
-
-        destination.flush().await?;
-
-        Ok(())
+        super::download_reqwest(source_name, builder, &self.timeouts, destination).await
     }
 }
```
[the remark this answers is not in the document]
Dav1dde: Yeah no shit, I even changed the tests and mention it in the PR description.

mujacica:
Do we support only AWS S3 or S3-compatible buckets as well? Not sure if all of them support AWS pre-signed mechanisms (SigV4) - but I think that might be a very niche use-case. Also the GetObject approach should work with server-sided encryption via pre-signed out-of-box - so that should be fine too.

crates/symbolicator-service/src/download/s3.rs
```diff
@@ -117,96 +129,104 @@ impl S3Downloader {
...
 
-        let mut body = std::pin::pin!(response.body);
-        let stream = futures::stream::poll_fn(move |cx| body.as_mut().poll_next(cx))
-            .map_err(|err| CacheError::download_error(&err));
+        let details = match metadata.as_ref().map(|m| (m.code(), m.message())) {
+            Some((Some(code), None)) => code.to_owned(),
+            Some((None, Some(message))) => format!("{status}: {message}"),
+            Some((Some(code), Some(message))) => format!("{code}: {message}"),
+            None | Some((None, None)) => status.to_string(),
+        };
```
loewenheim: The nesting can be reduced by accessing `code` and `message` separately. I suppose it's up for debate which version is more elegant.
```suggestion
        let details = match (
            metadata.as_ref().and_then(|m| m.code()),
            metadata.as_ref().and_then(|m| m.message()),
        ) {
            (Some(code), None) => code.to_owned(),
            (None, Some(message)) => format!("{status}: {message}"),
            (Some(code), Some(message)) => format!("{code}: {message}"),
            (None, None) => status.to_string(),
        };
```

Dav1dde: Thanks, that's nicer!

PR #1918, opened 2026-04-09
ref(crash): Switch to Sentry's native crash backend

loewenheim:
Based on https://github.com/getsentry/relay/pull/5789.


loewenheim:
Closing this for now—it's more complicated to set up than expected, see https://github.com/getsentry/relay/pull/5800.

PR #1919, opened 2026-04-09
fix: Patch RUSTSEC-2026-0049 by removing legacy rustls-webpki 0.101.7

jjbayer:
* Update `rustls-webpki` from 1.103.7 to 1.103.10 (the patched version that resolves the linked vulnarability). 
* Remove the older `rustls-webpki` dependency and a bunch of other duplicates by listing `aws-sdk-s3` features explicitly.

Closes #1909

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -564,20 +564,15 @@ dependencies = [
  "aws-smithy-runtime-api",
  "aws-smithy-types",
  "bytes",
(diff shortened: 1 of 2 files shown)


loewenheim (approved):
Thanks for taking  care of this!

Dav1dde (approved):
Nice!

PR #1930, opened 2026-04-21
fix(proguard): Support Compose group-key mappings

### LINKED ISSUE #1065: Compose diagnostic stack traces not deobfuscated in Sentry
Jetpack Compose source-info stack traces (via `Composer.setDiagnosticStackTraceMode(ComposeStackTraceMode.Auto)`) are still obfuscated in Sentry for an app using Compose 1.10.1.

Customer reports they are on Sentry Gradle plugin 6.x, KGP 2.3.x, AGP 8.13.2, with `includeProguardMapping = true` and `autoUploadProguardMapping = true` (autoInstallation and tracingInstrumentation disabled).

Hypothesis: Compose mapping entries (e.g. `m$1349809332`) may be missing from the uploaded mapping because `merge{Variant}ComposeMapping` didn’t run, or `uploadSentryProguardMappings{Variant}` runs before `merge{Variant}ComposeMapping`.

Reference: [https://developer.android.com/develop/ui/compose/tooling/stacktraces](<https://developer.android.com/develop/ui/compose/tooling/stacktraces>)

romtsn:
Support Compose group-key mappings in symbolicator

Symbolicator caches ProGuard mappings through rust-proguard before JVM symbolication. `ComposeStackTrace` -> `$$compose` sections emitted by recent Android builds can use two-space member indentation, which caused older cached mappings to keep only the class-level remap and leave frames as `ComposeStackTrace.m$...` instead of resolving them to the original Compose methods.

This bumps proguard to 5.10.2, adds an integration test with a raw two-space `$$compose` mapping fixture, and bumps symbolicator's own ProGuard cache version so previously truncated caches are recomputed with the fixed parser.

Closes getsentry/sentry-android-gradle-plugin#1065

### DIFF
--- a/crates/symbolicator-service/src/caches/versions.rs
+++ b/crates/symbolicator-service/src/caches/versions.rs
@@ -283,6 +283,9 @@ pub const BUNDLE_INDEX_CACHE_VERSIONS: CacheVersions = CacheVersions {
 
 /// Proguard Cache, with the following versions:
 ///
+/// - `7`: Recompute caches after starting to support Compose
+///   mappings with two-space indentation.
+///
 /// - `6`: Information about whether a method has rewrite rules is now part
 ///   of the cache format.
 ///
@@ -298,7 +301,7 @@ pub const BUNDLE_INDEX_CACHE_VERSIONS: CacheVersions = CacheVersions {
 ///
 /// - `1`: Initial version.
 pub const PROGUARD_CACHE_VERSIONS: CacheVersions = CacheVersions {
-    current: CacheVersion::new(6, CachePathFormat::V2),
+    current: CacheVersion::new(7, CachePathFormat::V2),
     fallbacks: &[],
     previous: &[
         CacheVersion::new(1, CachePathFormat::V1),
@@ -306,6 +309,7 @@ pub const PROGUARD_CACHE_VERSIONS: CacheVersions = CacheVersions {
         CacheVersion::new(3, CachePathFormat::V2),
         CacheVersion::new(4, CachePathFormat::V2),
         CacheVersion::new(5, CachePathFormat::V2),
+        CacheVersion::new(6, CachePathFormat::V2),
     ],
 };
 static_assert!(proguard::PRGCACHE_VERSION == 4);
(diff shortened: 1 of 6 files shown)


crates/symbolicator-service/src/caches/versions.rs
```diff
@@ -283,6 +283,9 @@ pub const BUNDLE_INDEX_CACHE_VERSIONS: CacheVersions = CacheVersions {
 
 /// Proguard Cache, with the following versions:
 ///
+/// - `7`: Recompute caches after starting to support Compose
+///   mappings with two-space indentation.
```
loewenheim: I wouldn't put the "recompute" part in the docs.
```suggestion
/// - `7`: Caches now support Compose
///   mappings with two-space indentation.
```

crates/symbolicator-service/src/caches/versions.rs
```diff
@@ -298,14 +301,15 @@ pub const BUNDLE_INDEX_CACHE_VERSIONS: CacheVersions = CacheVersions {
 ///
 /// - `1`: Initial version.
 pub const PROGUARD_CACHE_VERSIONS: CacheVersions = CacheVersions {
-    current: CacheVersion::new(6, CachePathFormat::V2),
+    current: CacheVersion::new(7, CachePathFormat::V2),
     fallbacks: &[],
```
loewenheim: I strongly suggest adding version 6 as a fallback here. Not having it means that all current caches are instantly invalidated and new ones need to be computed immediately. This isn't really necessary in this case—the old caches are still parseable. If you put the fallback, old caches will be reused until the new ones have been computed.

loewenheim (changes requested):

crates/symbolicator-service/src/caches/versions.rs
romtsn: TIL, thank you! added it

loewenheim (approved):

loewenheim (changes requested):
Sorry, forgot: A changelog entry would be appreciated :)

loewenheim (approved):

PR #1933, opened 2026-04-22
fix(proguard): Remap root-package exception classes

romtsn:
## Summary

When R8 renames a Java class into the root package (no dots), the SDK sends the exception with an empty `module` and a bare `ty` such as `kj`. `map_exception` was building the lookup key as `format!("{}.{}", module, ty)`, producing `.kj`, which never matches the mapping's class key `kj` — so `exception.type` stayed obfuscated even though the same class, when seen as a stack frame's `module`, was remapped fine.

Concretely, Compose now ships `androidx.compose.runtime.tooling.DiagnosticComposeException`, which R8 aggressive mode can rename to `kj`. Frames on `kj` were symbolicating; `exception.type` wasn't.

This PR:
- Adds `JvmException::full_class_name()` returning `Cow<str>` — borrows `ty` when `module` is empty, formats `module.ty` otherwise.
- Uses it for both the `map_exception` lookup key and the rewrite-rule `exception_descriptor` in `map_stacktrace` (two identical copies before).
- Tolerates a remapped class without a dot, since a root-package obfuscated class can in principle map back to another root-package class. Replaces the `tracing::error!` branch (which was treating this as "can not actually occur") with a graceful empty-module fallback.

Companion fix on the Sentry side: https://github.com/getsentry/sentry/pull/113691 — sentry's current filter drops module-less exceptions before they reach symbolicator, so this change by itself has no user-visible effect until that ships too.

## Test plan

- [x] New unit test `remap_exception_root_package` covering `kj` → `androidx.compose.runtime.tooling.DiagnosticComposeException`
- [x] Existing `remap_exception_simple` still passes
- [x] Full `symbolicator-proguard` test suite: 13/13 pass

### DIFF
--- a/crates/symbolicator-proguard/src/interface.rs
+++ b/crates/symbolicator-proguard/src/interface.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::sync::Arc;
 use std::{collections::HashMap, fmt};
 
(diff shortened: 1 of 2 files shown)


Dav1dde (approved):
LGTM, maybe give others a change to take a look as well before merging

loewenheim (approved):

PR #1942, opened 2026-05-04
dep: Update reqwest fork

loewenheim:
This updates our `reqwest` fork to https://github.com/getsentry/reqwest/commit/55b7d4586ef50c705e567f4c04a23a090ece9416, which is on par with upstream/master.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -946,6 +946,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2fd1289c04a9ea8cb22300a459a72a385d7c73d3259e2ed7dcb2af674838cfa9"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
+
+[[package]]
 name = "chrono"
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -3987,7 +4184,7 @@ checksum = "989425268ab5c011e06400187eed6c298272f8ef913e49fcadc3fda788b45030"
 dependencies = [
  "httpdate",
  "native-tls",
- "reqwest",
+ "reqwest 0.12.28",
```
Dav1dde: We'll also want to bump sentry to avoid the double reqwest dependency.

Cargo.toml
```diff
@@ -35,7 +35,7 @@ codegen-units = 256
 
 [patch.crates-io]
 # This patch adds an `ip_filter` able to filter out internal IP addresses in DNS resolution
-reqwest = { git = "https://github.com/getsentry/reqwest", branch = "swatinem/update-restricted-connector" }
+reqwest = { git = "https://github.com/getsentry/reqwest", branch = "restricted-connector" }
```
Dav1dde, line 38: With `0.13` reqwest enables rustls by default, I think we should for now stick with openssl and make the switch to rustls separately (if we want to)

Dav1dde (changes requested):

loewenheim:
@Dav1dde good catches, thanks!

Dav1dde (approved):

PR #1944, opened 2026-05-05
fix(download): Force identity encoding on range requests

Dav1dde:
This is effectively already happening with the current version of reqwest, the update to 0.13 in #1942 changes the behaviour and we need to explicitly force it.

The better solution is to implement proper support for compressed ranges as described in #1891 

Test snapshots updated because of a `tower-http` bug serving files without extensions with the `identity` encoding: https://github.com/tower-rs/tower-http/issues/664

### DIFF
--- a/crates/symbolicator-native/tests/integration/snapshots/integration__symbolication__add_bucket-2.snap
+++ b/crates/symbolicator-native/tests/integration/snapshots/integration__symbolication__add_bucket-2.snap
@@ -41,12 +41,17 @@ modules:
             has_unwind_info: false
             has_symbols: true
             has_sources: false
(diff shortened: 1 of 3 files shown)


loewenheim (approved):

Dav1dde:
For the future readers, add `--compressed` or an `Accept-Encoding` header:

```
$ curl -H 'range: bytes=0-100' -H 'accept: */*' https://raw.githubusercontent.com/dotnet/runtime/d099f075e45d2aa6007a22b71b45a08758559f80/src/libraries/Common/src/System/ThrowHelper.cs -vvv
```

PR #1946, opened 2026-05-06
fix(objectstore): Retry attachment downloads

Dav1dde:
Ideally we'd be using the download service directly (as noted in the comment), there is an attempt in #1928 but requires a bit more thought and work.

Right now we do see some download failures which can be mitigated with retries. So while it's not great and makes the `GenericErrorHandler` public with a helper fn, it's what we got.

### DIFF
--- a/crates/symbolicator-native/src/symbolication/attachments.rs
+++ b/crates/symbolicator-native/src/symbolication/attachments.rs
@@ -2,7 +2,7 @@ use std::fs::File;
 
 use futures::TryStreamExt;
 use symbolicator_service::caching::CacheError;
-use symbolicator_service::download::DownloadService;
+use symbolicator_service::download::{self, DownloadService};
 use tokio::io::{AsyncSeekExt, AsyncWriteExt, BufWriter};
 
 use crate::interface::AttachmentFile;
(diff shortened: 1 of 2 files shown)


crates/symbolicator-native/src/symbolication/attachments.rs
```diff
@@ -20,31 +19,37 @@ pub async fn download_attachment(
...
-
-    file.rewind().await?;
-
-    Ok(file.into_std().await)
+    download::retry(|| async {
+        // TODO: maybe its worth using the actual `DownloadService` instead of straight going to the `trusted_client`.
+        // Doing so would in theory allow us to have retries and error report, as well as being able to
+        // download files in multiple chunks concurrently, but I don’t think our `objecstore` server currently
+        // supports range requests, and those would also mess with streaming decompression.
+        // Not to mention that using the `DownloadService` is not that straight forward.
```
jjbayer: nit: I would move the comment above the `retry` code because it applies to it as well (also smaller diff).

jjbayer (approved):

loewenheim (approved):

PR #1956, opened 2026-05-22
feat(symbolicli): Enable local JS symbolication

### LINKED ISSUE #1939: symbolicli: Enable purely local JS symbolication
JS symbolication doesn't work in offline mode right now, because a Sentry source is always required. The easiest (if perhaps not most principled) way to allow symbolication with local artifact bundles would be to simulate a Sentry source locally.

loewenheim:
This makes it possible to symbolicate JS events locally with `symbolicli` by running a pretend Sentry artifact lookup endpoint. Just put some artifact bundles in a directory and call
```
symbolicli <event.json> --offline --symbols <DIR>
```
and you should be good. For now it only supports bundles, not individual artifact files.

Closes #1939. Closes SYMBOLI-58.


crates/symbolicli/src/event.rs
loewenheim, line 1: This file is just the `event` module split out of `main.rs`, with one small fix noted below.

crates/symbolicli/src/event.rs
```diff
@@ -0,0 +1,331 @@
...
+#[derive(Debug, Deserialize, Default)]
+struct DebugMeta {
+    images: Vec<Module>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(untagged)]
+pub enum Module {
+    Sourcemap(JsModule),
+    Object(RawObjectInfo),
```
loewenheim, line 179: I switched the order here compared to master. Because of `serde(untagged)` and the fact that `RawObjectInfo` matches any `type`, we need to put other variants first.

crates/symbolicli/src/main.rs
```diff
@@ -449,337 +476,3 @@ mod remote {
         }
     }
 }
-
-mod event {
```
loewenheim, line 453: As mentioned before this module was moved out of this file.

crates/symbolicli/src/js_local_source.rs
```diff
@@ -0,0 +1,265 @@
...
+            }
+            let path_encoded = urlencoding::encode_binary(path.as_os_str().as_encoded_bytes());
+            out.push(LookupResult::Bundle {
+                id: path.to_string_lossy().to_string(),
+                url: index.url(&format!("bundles/{path_encoded}")),
+                resolved_with: ResolvedWith::Release,
+            });
+            found_bundles.insert(path);
+        }
+    }
```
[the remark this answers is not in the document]
loewenheim: True in principle, but in practice the lookup function is only ever called with one debug_id/url.

crates/symbolicli/src/js_local_source.rs
```diff
@@ -0,0 +1,265 @@
...
+            .flat_map(|s| s.iter().cloned())
+        {
+            if found_bundles.contains(&path) {
+                continue;
+            }
+            let path_encoded = urlencoding::encode_binary(path.as_os_str().as_encoded_bytes());
+            out.push(LookupResult::Bundle {
+                id: path.to_string_lossy().to_string(),
+                url: index.url(&format!("bundles/{path_encoded}")),
+                resolved_with: ResolvedWith::Release,
```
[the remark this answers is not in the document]
loewenheim: Seems to work fine in my tests (including `test_existing_bundle`).

crates/symbolicli/src/js_local_source.rs
```diff
@@ -0,0 +1,269 @@
...
+            }
+
+            let bundle_path: Arc<Path> = entry.path().into();
+            let relative_path: Arc<Path> = bundle_path.strip_prefix(base_path).unwrap().into();
+
+            let archive = std::fs::File::open(&bundle_path)?;
+            let mut archive = ZipArchive::new(archive)?;
+
+            let manifest = archive.by_name("manifest.json")?;
+            let manifest: SourceBundleManifest = serde_json::from_reader(manifest)?;
```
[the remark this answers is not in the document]
loewenheim: I believe this is acceptable.

crates/symbolicli/src/js_local_source.rs
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    if let Some(url) = key.url {
+        for path in index
+            .by_url
+            .get(&ReleaseDistUrl {
+                release: key.release.clone(),
+                dist: key.dist.clone(),
+                url: url.clone(),
+            })
```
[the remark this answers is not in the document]
loewenheim: Fixed.

crates/symbolicli/src/js_local_source.rs
```diff
@@ -0,0 +1,286 @@
...
+use walkdir::WalkDir;
+use zip::ZipArchive;
+
+/// Start a server that mimicks the Sentry artifact lookup endpoint, serving artifact
+/// bundles from the given directory.
+pub fn start_server(path: impl AsRef<Path> + Clone) -> anyhow::Result<SentrySourceConfig> {
+    let addr = SocketAddr::from(([127, 0, 0, 1], 0));
+    let listener = TcpListener::bind(addr).unwrap();
+    listener.set_nonblocking(true).unwrap();
+    let socket = listener.local_addr().unwrap();
```
Dav1dde: Instead of unwrapping, why not bubble `?` the errors?

crates/symbolicli/README.md
```diff
@@ -48,6 +48,6 @@ You can control the level of logging output by passing the desired log level to
 Available levels are `off`, `error`, `warn`, `info`, `debug`, `trace`. The default is `info`.
 
 # Local Symbols
-The `--symbols` option allows you to supply a local directory containing debug files to use
-in addition to the configured sources. The directory must be sorted according to the
-`unified` layout. The easiest way to accomplish that is using `symsorter`.
+The `--symbols` option allows you to supply local debug information.
+- For native events, pass a local directory containing debug files to use in addition to the configured sources. The directory must be sorted according to the `unified` layout. The easiest way to accomplish that is using `symsorter`.
+- For JS events, pass a directory containing artifact bundles (individual files are not supported right now). Note that this only works in offline mode; in online mode, only files from Sentry will be used.
```
Dav1dde:
```suggestion
- For JS events, pass a directory containing artifact bundles (individual files are not supported right now). Note that this only works in offline mode; in online mode, only files from Sentry are used.
```

I think that reads better, the sentence before uses `are` as well.

Dav1dde (approved):

crates/symbolicli/src/js_local_source.rs
loewenheim: Yeah, you're right. Copy-pasted this and didn't clean up.

PR #1965, opened 2026-06-12
feat(native): Allow limiting size of compressed debug file sections

loewenheim:
This bumps `symbolic` to 13.2.0, giving us the ability to limit the sizes of decompressed sections in ELF files. It adds a new config setting `object_file_max_decompressed_section_size` to configure this behavior. The default is 4GiB.

Closes INGEST-949.

### DIFF
--- a/crates/symbolicator-service/src/config.rs
+++ b/crates/symbolicator-service/src/config.rs
@@ -10,6 +10,7 @@ use std::time::Duration;
 use anyhow::{Context, Result};
 use sentry::types::Dsn;
 use serde::{Deserialize, Deserializer, de};
+use symbolic::debuginfo::ParseObjectOptions;
 use tracing::level_filters::LevelFilter;
 
 use symbolicator_sources::SourceConfig;
@@ -532,6 +533,11 @@ pub struct Config {
     /// This applies to the `/symbolicate`, `/symbolicate-any`, `/symbolicate-js`,
     /// and `/symbolicate-jvm` endpoints.
     pub symbolicate_body_max_bytes: usize,
+
+    /// Maximum decompressed size of compressed sections in debug files.
+    ///
+    /// Defaults to 4GiB
+    pub object_file_max_decompressed_section_size: Option<usize>,
 }
 
 impl Config {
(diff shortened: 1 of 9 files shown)


crates/symbolicator-service/src/objects/data_cache.rs
```diff
@@ -27,6 +27,13 @@ use crate::utils::sentry::ConfigureScope;
 
 use super::meta_cache::FetchFileMetaRequest;
 
+static PARSE_OBJECT_OPTIONS: LazyLock<ParseObjectOptions> = LazyLock::new(|| {
+    let mut options = ParseObjectOptions::default();
+    // Limit decompressed debug file section size to 4GiB.
+    options.max_decompressed_section_size = Some(4 << 30);
```
Dav1dde: Is it feasible to actually make this a configuration option for Symbolicator?

loewenheim: We can absolutely do that.

Dav1dde: I think that's what we should do, if you want to just get this out now and do that later, also fine by me.

Dav1dde (approved):

crates/symbolicator-service/src/config.rs
```diff
@@ -549,6 +555,12 @@ impl Config {
     pub fn default_sources(&self) -> Arc<[SourceConfig]> {
         self.sources.clone()
     }
+
+    pub fn parse_object_options(&self) -> ParseObjectOptions {
+        let mut opts = ParseObjectOptions::default();
+        opts.max_decompressed_section_size = self.object_file_max_decompressed_section_size;
+        opts
```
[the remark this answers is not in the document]
loewenheim: Incorrect. If the key is missing it does _not_ get deserialized as `None`. To disable the setting it has to be explicitly set to `null`.

crates/symbolicator-service/src/config.rs
```diff
@@ -620,10 +632,11 @@ impl Default for Config {
             propagate_traces: true,
             // We currently accept 200MiB minidumps in Sentry. This allows for that size
             // plus some extra for the rest of the request.
-            crash_file_body_max_bytes: 250 * 1024 * 1024,
+            crash_file_body_max_bytes: 250 << 20,
```
Dav1dde: I find this harder to read 🙈

Dav1dde (approved):

crates/symbolicator-service/src/config.rs
loewenheim: Fair enough, I'll revert.

PR #1969, opened 2026-06-24
ref(symbolicator): Replace CallOnDrop with defer

Dav1dde:
They do the same thing.


loewenheim (approved):

jjbayer (approved):
Interesting, why were there two implementations in the first place?

loewenheim:
> Interesting, why were there two implementations in the first place?

`CallOnDrop` is the older one; I didn't realize it existed when I added `defer`.

PR #1970, opened 2026-06-24
fix(api): Reduce inflight requests on panic

Dav1dde:
The inflight requests counter would fail to be decrement on panic. This fixes the issue, but also introduces a new deferred primitive to make accidentally forgetting impossible in the type system.

Fixes: INGEST-969

### DIFF
--- a/crates/symbolicator-service/src/utils/defer.rs
+++ b/crates/symbolicator-service/src/utils/defer.rs
@@ -1,3 +1,6 @@
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+
 /// Guard that runs a closure when dropped.
 ///
 /// The closure must not panic under any circumstance. Since it is called while dropping an item,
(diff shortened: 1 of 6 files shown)


crates/symbolicator-service/src/utils/defer.rs
```diff
@@ -17,3 +20,168 @@ impl<F: FnOnce()> Drop for DeferGuard<F> {
 pub fn defer<F: FnOnce()>(f: F) -> DeferGuard<F> {
     DeferGuard(Some(f))
 }
+
+/// A shared counter similar to a semaphore.
+///
+/// The counter can be temporarily incremented and is automatically decremented again.
+#[derive(Debug, Clone)]
+pub struct DeferCounter(Arc<AtomicUsize>);
```
jjbayer, line 28: Could this be implemented with an `Arc<()>`? Didn't think it through entirely but since `Arc` already has a built-in counter, you might not need the AtomicUsize within.

loewenheim (approved):
Very nice, thanks for doing this!

crates/symbolicator-service/src/utils/defer.rs
Dav1dde: That should work, but I am not sure it's worth the hassle.

Dav1dde: Merging for now, happy to try this/do it in a follow-up

PR #1972, opened 2026-06-25
feat(native): Allow limiting size of embedded sources

loewenheim:
This bumps `symbolic` to 13.6.0 and adds a config option `object_file_max_decompressed_source_size` that lets us limit the sizes of compressed source files embedded in debug files.

Closes INGEST-966.

### DIFF
--- a/crates/symbolicator-service/src/config.rs
+++ b/crates/symbolicator-service/src/config.rs
@@ -538,6 +538,11 @@ pub struct Config {
     ///
     /// Defaults to 4GiB
     pub object_file_max_decompressed_section_size: Option<usize>,
+
+    /// Maximum decompressed size of compressed source files embedded in debug files.
+    ///
+    /// Defaults to 100MiB.
+    pub object_file_max_decompressed_source_size: Option<usize>,
 }
 
 impl Config {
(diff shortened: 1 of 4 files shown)


crates/symbolicator-service/src/config.rs
```diff
@@ -559,6 +564,7 @@ impl Config {
     pub fn parse_object_options(&self) -> ParseObjectOptions {
         let mut opts = ParseObjectOptions::default();
         opts.max_decompressed_section_size = self.object_file_max_decompressed_section_size;
+        opts.max_decompressed_embedded_source_size = self.object_file_max_decompressed_source_size;
```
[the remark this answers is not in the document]
loewenheim: No, a missing key is not deserialized as `None`, it is deserialized as the default value because of `serde(default)` on the struct.

crates/symbolicator-service/src/config.rs
```diff
@@ -538,6 +538,11 @@ pub struct Config {
     ///
     /// Defaults to 4GiB
     pub object_file_max_decompressed_section_size: Option<usize>,
+
+    /// Maximum decompressed size of compressed source files embedded in debug files.
+    ///
+    /// Defaults to 100MiB.
+    pub object_file_max_decompressed_source_size: Option<usize>,
```
jjbayer, line 545: If this defaults to 100 MiB, does it even make sense to make it an `Option`? Users can always set a very large number if they want de-facto no limit.

jjbayer (approved):

crates/symbolicator-service/src/config.rs
loewenheim: Same applies to all of these size-limiting options—in principle they're all bounded by `usize::MAX`. For understanding I think it's cleaner if they're `Option`s.

PR #1977, opened 2026-07-03
chore: Bump symbolic version

tobias-wilfert:
Fix: INGEST-950

### DIFF
--- a/crates/symbolicator-native/src/caches/cficaches.rs
+++ b/crates/symbolicator-native/src/caches/cficaches.rs
@@ -8,7 +8,7 @@ use sentry::types::DebugId;
 use symbolicator_service::caches::CacheVersions;
 use tempfile::NamedTempFile;
 
-use symbolic::cfi::CfiCache;
+use symbolic::cfi::{CfiCache, FromObjectOptions};
 use symbolic::common::ByteView;
 use symbolic::debuginfo::breakpad::BreakpadModuleRecord;
 use symbolicator_service::caches::versions::CFICACHE_VERSIONS;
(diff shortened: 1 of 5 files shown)


crates/symbolicator-service/src/config.rs
```diff
@@ -567,6 +573,12 @@ impl Config {
         opts.max_decompressed_embedded_source_size = self.object_file_max_decompressed_source_size;
         opts
     }
```
tobias-wilfert, line 575: Tried to follow this approach.

crates/symbolicator-native/src/caches/cficaches.rs
```diff
@@ -69,13 +69,20 @@ fn parse_cfi_cache(bytes: ByteView<'static>) -> CacheContents<(u32, CfiItem)> {
...
+    ops: FromObjectOptions,
 }
 
 impl CfiCacheActor {
-    pub fn new(cache: Cache, shared_cache: SharedCacheRef, objects: ObjectsActor) -> Self {
+    pub fn new(
+        cache: Cache,
+        shared_cache: SharedCacheRef,
+        objects: ObjectsActor,
+        ops: FromObjectOptions,
```
Dav1dde, line 80: Wondering if this should just take the entire Symbolicator config instead

crates/symbolicator-service/src/config.rs
```diff
@@ -567,6 +573,12 @@ impl Config {
...
+
+    pub fn from_object_options(&self) -> FromObjectOptions {
+        let mut opts = FromObjectOptions::default();
+        opts.max_unwind_chain_len = self.max_unwind_chain_len;
+        opts
+    }
 }
 
 /// Checks if we are running in docker.
```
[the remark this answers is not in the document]
tobias-wilfert: @loewenheim Will we need a cache version bump for this?

crates/symbolicator-native/src/caches/cficaches.rs
tobias-wilfert: Happy to change it if we think that is better.

crates/symbolicator-service/src/config.rs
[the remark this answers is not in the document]
loewenheim: No. The new option doesn't improve the quality of caches or add anything new, so there's no point regenerating them.

loewenheim (approved):

crates/symbolicator-native/src/caches/cficaches.rs
Dav1dde: It's a bit awkward that the `CfiCacheActor` now depends on a config of `FromObject`, but not too strong of an opinion here as we can always change it.

@loewenheim wdyt?

loewenheim: I think it's reasonable

PR #1985, opened 2026-07-10
feat: add configurable unwind_strategy for minidump stackwalking

aholza:
## Problem

Symbolicator has always tried CFI (`.eh_frame`/`.debug_frame`) before the
frame-pointer chain when stackwalking minidumps. Debuggers like GDB do the
opposite by default. For binaries built with frame pointers preserved
(`-fno-omit-frame-pointer`) whose CFI happens to be present but inaccurate
(e.g. FPO'd system libraries with truncated/wrong CFI), CFI-first can walk
into garbage before falling back, while frame-pointer-first would have
produced a correct unwind immediately.

## Solution

Adds an `unwind_strategy` config option (`cfi_first`, the default and
unchanged historical behavior, or `frame_pointer_first`) to `Config`,
threaded through `SymbolicationActor` → `SymbolicatorSymbolProvider` →
`minidump-unwind`'s per-architecture `get_caller_frame()`, which now
branches on the requested strategy to reorder the CFI vs. frame-pointer
attempts. Stack scanning remains the last-resort fallback in both cases.

This requires a small, currently-pending upstream change to
`minidump-unwind` (see https://github.com/rust-minidump/rust-minidump —
PR link once opened there), adding an `UnwindStrategy` enum and a
`SymbolProvider::unwind_strategy()` default method so the ordering can be
controlled from the calling crate rather than hardcoded per architecture.
`Cargo.toml` currently pins `minidump`/`minidump-common`/`minidump-unwind`
to that fork branch pending review/release there.

Also exposes `--unwind-strategy` on `symbolicli` for local comparison
against tools like GDB.

## Testing

- Full `symbolicator-native` suite (34 tests + 1 doctest) plus
  `minidump-unwind`'s own suite (76 tests) on the forked crate.
- New integration test `test_minidump_frame_pointer_first_matches_cfi_first`
  confirming both strategies agree when CFI is trustworthy.
- New snapshot test `test_minidump_windows_frame_pointer_first` against a
  real Windows minidump with an FPO'd `rpcrt4.dll`, showing
  `frame_pointer_first` produces `trust: fp` frames where `cfi_first`
  produces `trust: scan` for the same addresses — the actual real-world
  divergence this option exists for.
- End-to-end: ran both strategies via `symbolicli` against a real crash
  minidump to confirm the flag works cleanly through the full pipeline.

## Dependencies

Blocked on review/release of the companion `rust-minidump` PR; the
`[patch.crates-io]` git-branch pin in `Cargo.toml` can be replaced with a
plain crates-io version bump once that lands.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -810,6 +810,22 @@ dependencies = [
 [[package]]
 name = "breakpad-symbols"
 version = "0.26.1"
+dependencies = [
+ "async-trait",
+ "cachemap2",
+ "circular",
+ "debugid",
+ "futures-util",
+ "minidump-common",
+ "nom",
+ "range-map",
+ "thiserror 2.0.17",
+ "tracing",
+]
+
+[[package]]
+name = "breakpad-symbols"
+version = "0.26.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b544545512b213899840bb26f72cd13f9fb4c3de3dd5f8dc7219688f54aebe9"
 dependencies = [
@@ -2829,8 +2845,6 @@ dependencies = [
 [[package]]
 name = "minidump"
 version = "0.26.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a902ca21d9772a66d3d1b050b3436dcadb192694be01409e0219902dcf4bc1e8"
 dependencies = [
  "debugid",
  "encoding_rs",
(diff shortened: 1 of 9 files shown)


Dav1dde:
Please explain your motivation for these changes. Why are _you_ interested in this?

aholza:
Motivation is to replicate what GDB does for stack unwinding. He does first frame pointer unwind and then falls back to CFI unwinding.

Dav1dde:
That's not your personal motivation, why do you want this?

Dav1dde:
At this time we have no plans to support this. If there is a concrete problem you want to solve, we can look into that instead.

PR #1994, opened 2026-07-29
ref(cache): Report cleanup stats as we go

loewenheim:
Rather than collecting the cleanup stats for an individual cache cleanup run in a struct and reporting them at the end, simply emit them immediately.

One wrinkle here is that I'm not entirely sure using a counter for the "retained" stats makes sense—aggregating over several cleanup runs would double count them. I'm not sure this is really possible to resolve without doing what we previously did.

### DIFF
--- a/crates/symbolicator-service/src/caching/cleanup.rs
+++ b/crates/symbolicator-service/src/caching/cleanup.rs
@@ -128,10 +128,15 @@ impl Cache {
     /// If `dry_run` is `true`, no files will actually be deleted.
     #[tracing::instrument(skip(self), fields(cache = %self.name))]
     pub fn cleanup(&self, dry_run: bool) -> Result<()> {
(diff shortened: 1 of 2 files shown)


Dav1dde:
> One wrinkle here is that I'm not entirely sure using a counter for the "retained" stats makes sense

A [gauge](https://docs.rs/metrics/latest/metrics/trait.GaugeFn.html#tymethod.increment) should work for this, it can be `set` and `incremented`.

Dav1dde (approved):
See my gauge comment.

loewenheim:
> A [gauge](https://docs.rs/metrics/latest/metrics/trait.GaugeFn.html#tymethod.increment) should work for this, it can be set and incremented.

Oh, I thought gauges could only be set. That's perfect then.

PR #1999, opened 2026-08-05
feat(download): Add `sentry` client with auto-decompression off

lcian:
We want to start storing (and serving) debug files compressed.
Without this PR, that won't work: Symbolicator uses range requests, but Sentry (Objectstore) would serve the range over the compressed bytes and return a response with `Content-Encoding: zstd`, so Symbolicator would try to decompress that stream individually and fail, as shown here: https://github.com/getsentry/sentry/pull/121182.

This creates an additional `sentry` client, which has the same settings as the `trusted` client, except that it disables auto-decompression. This way, we won't run into the problem described above: Symbolicator will fetch the whole compressed file via range requests, and then decompress it via `maybe_decompress_file`.

Ref FS-390

### DIFF
--- a/crates/symbolicator-service/src/download/mod.rs
+++ b/crates/symbolicator-service/src/download/mod.rs
@@ -141,8 +141,8 @@ impl DownloadService {
         };
         let trusted_settings = ClientSettings {
             connect_to_reserved_ips: true,
-            // Sentry debug files can be downloaded in ranges. Receiving compressed partial
-            // responses breaks the individual range streams.
+            // Sentry can return the raw byte stream on range requests, which might be part of a
+            // compressed stream, so transparent decompression would fail on those individual parts.
             compression: false,
             ..restricted_settings
         };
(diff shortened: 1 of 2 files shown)


crates/symbolicator-service/src/download/mod.rs
```diff
@@ -155,6 +162,7 @@ impl DownloadService {
         };
 
         let trusted_client = crate::utils::http::create_client(&trusted_settings);
+        let sentry_client = crate::utils::http::create_client(&sentry_settings);
```
lcian, line 165: We need a separate client, as disabling compression on the `trusted_client` breaks attachments, as for those we don't use `fetch_file`/`maybe_decompress_file`, but rather `trusted_client.get` directly.
I don't know if that code can be easily changed in a way that would allow us to avoid creating a separate client.

Dav1dde:
See also https://github.com/getsentry/symbolicator/issues/1891 why this is necessary. Once the range request handling is improved, we can undo this change.

Dav1dde (approved):
Since files aren't compressed yet, this may have a significant impact on amount of bytes transferred. Considering this is all internal traffic, I expect this to be okay though.

lcian:
The [test](https://github.com/getsentry/sentry/pull/121182) passes now.

PR #2000, opened 2026-08-05
ref: Add logging for 403s

elramen:
Add additional logging for 403 failures when trying to scrape source map with user configured auth token.

Ref: https://linear.app/getsentry/issue/INGEST-1111

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4834,6 +4834,7 @@ version = "26.6.0"
 dependencies = [
  "data-url",
  "futures",
+ "hex",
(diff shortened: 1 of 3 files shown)


Dav1dde:
Few thoughts:
- For native events, we put this information into the event, so it is visible also from Sentry and we can show the user what went wrong. Is that not the case for JS, if not, what does it take to add the info to the event instead?
- This is an unconditional log line, which can be triggered on a per event basis, what tells us that this isn't too much information being logged?

PR #2001, opened 2026-08-05
ref: Improve logging for missing sourcemap with debug id

elramen:
Improve logging for get-sourcemap failures by logging the specific error, e.g. `NotFound`, `Timeout` etc.

Ref: https://linear.app/getsentry/issue/INGEST-1112/

### DIFF
--- a/crates/symbolicator-js/src/lookup.rs
+++ b/crates/symbolicator-js/src/lookup.rs
@@ -679,10 +679,12 @@ impl ArtifactFetcher {
             if let Some(debug_id) = debug_id
                 && rand::random::<f64>() < 0.0001
             {
+                let cache_error = sourcemap.entry.as_ref().err();
                 tracing::error!(
                     source_url = %self.source.url,
                     abs_path,
                     %debug_id,
+                    ?cache_error,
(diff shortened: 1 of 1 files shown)


crates/symbolicator-js/src/lookup.rs
Dav1dde: Instead of `is_err` you can match on the error instead.

crates/symbolicator-js/src/lookup.rs
```diff
@@ -679,10 +679,12 @@ impl ArtifactFetcher {
             if let Some(debug_id) = debug_id
                 && rand::random::<f64>() < 0.0001
             {
+                let cache_error = sourcemap.entry.as_ref().err();
                 tracing::error!(
                     source_url = %self.source.url,
                     abs_path,
                     %debug_id,
+                    ?cache_error,
```
Dav1dde: if you have the error here, you can `error = &err as &dyn std::error::Error` which gets Sentry to pick up the error as error instead of just the debug impl.

Dav1dde (changes requested):

Dav1dde (approved):

PR #2009, opened 2026-08-11
fix: Increase decompressed source map limit to 190 MiB

elramen:
Increase decompressed source map limit as the 100 MiB limit is too small for some customers. The 190 MiB should be safe - there is headroom even at max concurrent requests based on the average RSS.

Ref https://linear.app/getsentry/issue/INGEST-1112

### DIFF
--- a/crates/symbolicator-service/src/config.rs
+++ b/crates/symbolicator-service/src/config.rs
@@ -542,7 +542,7 @@ pub struct Config {
 
     /// Maximum decompressed size of compressed source files embedded in debug files.
     ///
-    /// Defaults to 100MiB.
+    /// Defaults to 200MiB.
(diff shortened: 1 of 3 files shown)


crates/symbolicator-service/src/config.rs
```diff
@@ -542,7 +542,7 @@ pub struct Config {
 
     /// Maximum decompressed size of compressed source files embedded in debug files.
     ///
-    /// Defaults to 100MiB.
+    /// Defaults to 200MiB.
```
tobias-wilfert: Should this not be 190MIB or ~200MB? Also any specific reason to not go with just 200MiB instead of 190?

tobias-wilfert (approved):

PR #2012, opened 2026-08-11
fix(index): Bound the index last id

Dav1dde:
Warden finding in another PR, we want a bound which is smaller than `u32::MAX` and still reasonable, to not have to fetch an infinity (aka `u32::MAX`) of files.

### DIFF
--- a/crates/symbolicator-service/src/download/index.rs
+++ b/crates/symbolicator-service/src/download/index.rs
@@ -28,8 +28,13 @@ use super::DownloadService;
 /// log file.
 const LASTID_FILE: &str = "000Admin/lastid.txt";
 
+/// Maximum last id we consider valid.
+///
+/// Symbolicator fetches all index files up to this maximum.
+const LASTID_MAX: u32 = 10_000;
+
 /// Maximum length for a valid id stored in the [`LASTID_FILE`].
-const LASTID_MAX_LENGTH: usize = 200;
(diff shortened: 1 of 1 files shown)


crates/symbolicator-service/src/download/index.rs
```diff
@@ -28,8 +28,13 @@ use super::DownloadService;
...
 const LASTID_FILE: &str = "000Admin/lastid.txt";
 
+/// Maximum last id we consider valid.
+///
+/// Symbolicator fetches all index files up to this maximum.
+const LASTID_MAX: u32 = 10_000;
+
 /// Maximum length for a valid id stored in the [`LASTID_FILE`].
-const LASTID_MAX_LENGTH: usize = 200;
+const LASTID_MAX_LENGTH: usize = (u32::MAX.ilog10() as usize + 1) * 2; // Amount of digits + some headroom
```
elramen:
```suggestion
const LASTID_MAX: u32 = 10_000;

/// Maximum length for a valid id stored in the [`LASTID_FILE`].
///
/// Digits in [`LASTID_MAX`], doubled to tolerate whitespace,
/// newlines, or leading zeros.
const LASTID_MAX_LENGTH: usize = (LASTID_MAX.ilog10() as usize + 1) * 2;
```

elramen (approved):

PR #2018, opened 2026-08-13
ref(native): Make demangle cache as a wrapper type

Dav1dde:
Small refactor to make the cache easier to use and to have a central place for demangling. For example only 1 of 2 code paths actually used the cache for demangling.


tobias-wilfert (approved):
Other than the cursor comment looks good.

PR #2035, opened 2026-08-28
feat(variables): Begin reading minidump memory for variables

Dav1dde:
Threads the minidump down to the variable extraction and implements a basic frame pointer offset resolution.

Obviously there is still a lot left to do, interpreting the memory based on the type etc, this is just threading down the minidump.

Note: This uses `scroll` version `0.12` instead of `0.12` because of the `minidump` crate.

### DIFF
--- a/crates/symbolicator-native/src/memory.rs
+++ b/crates/symbolicator-native/src/memory.rs
@@ -32,7 +32,7 @@ impl MemoryAccess for Minidump<'static, ByteView<'static>> {
 
 /// Extension trait for [`MemoryAccess`].
 pub trait MemoryAccessExt: MemoryAccess {
-    /// Helper which access the memory of a dump and converts the memory to the specified type.
+    /// Helper which accesses the memory of a dump and converts the memory to the specified type.
     fn get_value_at_address<T>(&self, addr: u64) -> Option<T>
     where
         T: SizeWith<scroll::Endian>,
--- a/crates/symbolicator-native/src/memory.rs
+++ b/crates/symbolicator-native/src/memory.rs
@@ -1,0 +1,49 @@
+use minidump::Minidump;
+use scroll::ctx::{SizeWith, TryFromCtx};
+use symbolic::common::ByteView;
+
+/// Unified memory access for crash dumps.
+pub trait MemoryAccess: std::fmt::Debug + Send + Sync {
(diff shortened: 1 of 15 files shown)


crates/symbolicator-native/src/symbolication/process_minidump.rs
```diff
@@ -672,6 +674,7 @@ impl SymbolicationActor {
             rewrite_first_module,
             frame_order: FrameOrder::CalleeFirst,
             extract_variables: request.extract_variables,
+            memory: Some(Arc::new(minidump)),
```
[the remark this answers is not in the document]
Dav1dde: I think that's okay as this is supposed to be the default.

crates/symbolicator-native/src/symbolication/process_minidump.rs
```diff
@@ -631,6 +631,8 @@ impl SymbolicationActor {
         metric!(distribution("minidump.upload.size") = len as f64);
 
         let bv = ByteView::map_file(minidump_file)?;
+        // let minidump = SelfCell::try_new(bv, |bv| Minidump::read(unsafe { &*bv }))?;
+
```
loewenheim:
```suggestion
```

crates/symbolicator-native/src/memory.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    fn endian(&self) -> scroll::Endian {
+        self.endian
+    }
+}
+
+/// Extension trait for [`MemoryAccess`].
+pub trait MemoryAccessExt: MemoryAccess {
+    /// Helper which access the memory of a dump and converts the memory to the specified type.
```
loewenheim:
```suggestion
    /// Helper which accesses the memory of a dump and converts the memory to the specified type.
```

crates/symbolicator-native/src/symbolication/process_minidump.rs
loewenheim, line 675: Nothing wrong with it necessarily, but why is this an `Arc`? Do we clone it anywhere?

Dav1dde: As discussed offline, the struct itself is `Clone`.

loewenheim (approved):

PR #2045, opened 2026-09-10
Expose wasm-split as a library

d2anamaria:
`wasm-split` can only be used as a command line tool today. Its logic lives
inside `main()`, so any other crate that needs to add a build ID or split a WASM
file has to shell out to the binary.

This makes the crate a library with a thin binary on top. `sentry-cli` and other
callers can now call `wasm_split::split(...)` directly from a prepare command,
passing plain paths and flags. The library pulls in no argument parsing, so
callers do not inherit a `clap` dependency.

The algorithm is unchanged. Build IDs are still reused when the module already
carries one, the external DWARF URL still falls back to the debug file name, and
the main module is still only rewritten when something about it actually
changed. `split` reports the build ID and the resolved output paths back to the
caller instead of printing them.

The command line tool keeps the same flags, the same output and the same exit
codes. It now just maps its arguments onto the library and prints the hex build
ID.


loewenheim:
How would you use the `wasm-split` library from e.g. `sentry-cli`? Symbolicator isn't published to crates.io.

IMO it's an accident of history that `wasm-split` is included with Symbolicator (it doesn't even share any code with the main program) and the correct thing to do would be to move it into its own repo.

d2anamaria:
> How would you use the `wasm-split` library from e.g. `sentry-cli`? Symbolicator isn't published to crates.io.
> 
> IMO it's an accident of history that `wasm-split` is included with Symbolicator (it doesn't even share any code with the main program) and the correct thing to do would be to move it into its own repo.

You're right, today you'd need a git dependency on getsentry/symbolicator. So cargo would clone the whole repo, but it only builds wasm-split because it's a leaf crate. Exposing it as a library (my recent change) is what lets cli depend on it in-process. Before that, the only option was shelling out to the binary.

loewenheim:
>You're right, today you'd need a git dependency on getsentry/symbolicator. So cargo would clone the whole repo, but it only builds wasm-split because it's a leaf crate. Exposing it as a library (my recent change) is what lets cli depend on it in-process; before that, the only option was shelling out to the binary.

I don't think this is a good way to handle this; introducing a dependency on Symbolicator to get access to a library which doesn't actually have anything to do with Symbolicator to begin with is pretty convoluted. If you like we can look into splitting it out together.

loewenheim:
Superseded by https://github.com/getsentry/cli/pull/1589.

Issue #2044, opened 2026-09-10
Add instrumentation for download progress on timeouts


loewenheim:
When we time out trying to download a file, it would be useful to know how far along we got. Ideally this could be emitted as a log.

PR #2047, opened 2026-09-14
feat(download): Add warnings for download timeouts

loewenheim:
This adds `warn` level logs for when a download times out either when making the connection or when the download is already in progress.

The `warn` level is used so these logs become events. There is currently no way to propagate information from the Sentry scope to logs.

Contributes to SYMBOLI-75. Contributes to #2044.


elramen (approved):

crates/symbolicator-service/src/download/mod.rs
```diff
@@ -1183,6 +1187,12 @@ impl Drop for MeasureSourceDownloadGuard<'_> {
                 "compression" => compression,
             );
         }
+
+        if matches!(self.state, MeasureState::Pending) {
+            let completion = (raw_bytes_transferred as f64) / (size as f64);
```
olksdr, line 1192: `size` can be 0?

elramen (approved):

crates/symbolicator-service/src/download/mod.rs
loewenheim: In that case, `completion` is `inf` and the next line filters it away.

PR #2049, opened 2026-09-14
meta: Remove vscode settings

elramen:
Let each developer use their own vscode settings (e.g. `formatAfterDelay` instead of `formatOnSave`).


tobias-wilfert (approved):
I guess since most of the formating is enforced in the CI this should be fine.

